Issue #625, opened 2018-03-05
Create a Wagtail setting that allows admins to choose fields for the 'General' filter section


raq929:
Currently, the 'General' section of the filters is hardcoded on the front end. I think we should create a Wagtail setting that allows admins to pick which fields on the IncidentPage model should be displayed in this section (Similar to how Categories now let admins do this.) Then the data from the setting needs to be passed to React, and the 'General' section can use the 'NewFilter' module like all the categories do now. 
This makes the code less brittle and resilient to field name changes, etc.

PR #650, opened 2018-03-19
Ensure that incident filters are sorted alphabetically

### LINKED ISSUE #649: Non-deterministic field choices for `incident_filter`
No matter how many times I run `./manage.py makemigrations` it keeps making new migrations like [this one](https://github.com/freedomofpress/pressfreedom/pull/640#discussion_r174779385) which makes me suspect the ordering of choices on that field is non-deterministic.

harrislapiroff:
Previously they were being iterated straight out of a
`dict`, but `dict`s are not guaranteed an order as of 3.6
and earlier. Since these filters are used as choices on a model
field, this was causing an amusing problem where migrations would
consistently be created for that model every time `makemigrations`
was run, since the choice order was changing every time.

As of 3.7, dictionaries preserve insertion order so, if desired,
we could remove the use of OrderedDict when upgrading to 3.7 and
not mandate sorting alphabetically.

Resolves #649


melinath (changes requested):
I think that this should sort choices in the FilterChoicesIterator and sort by verbose_name instead of choice value. I've implemented it this way on my branch for #625, but I'm down with it being in this PR instead.

PR #822, opened 2019-08-07
Use _top_image template instead of _teaser_image

harrislapiroff:
A previous feature change erroneously caused the _teaser_image
template to be used on incident detail pages. This meant that
incident detail templates would have lower resolution images, not
display in color without hover, and would unnecessarily self-referentially
link. This change restores the original template.


chigby (approved):
I tested this out and it works as described! Sometimes, these things they happen. Anyway, seems to work now.

PR #826, opened 2019-09-09
Separate CSV headers into a `set` for testing

chigby:
This change is motivated by two factors:

1. If this test fails, the output will now show exactly which fields are expected or unexpected (which is more useful).
2. If this test has to be updated, the diff is now limited to exactly which lines/fields have changed,  instead of a single giant text blob that changes internally. This has the side effect of making #776, which makes changes to this test, much easier to rebase.

So with this change we get better test output and better diffs.

One thing we lose is the expected ordering of the fields, but I don't
think that actually matters.  IMO, we're not testing the order of the
columns in the CSV, here.  If the CSV export changes the order of the
existing columns, I don't think that warrants a test failure.

### DIFF
--- a/incident/tests/test_incidents_export.py
+++ b/incident/tests/test_incidents_export.py
@@ -50,8 +50,70 @@ class IncidentExportTestCase(TestCase):
             response['Access-Control-ALlow-Methods'], 'GET,OPTIONS,HEAD'
         )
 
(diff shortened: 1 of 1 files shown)


SaptakS (approved):

harrislapiroff (approved):
lgtm!

PR #930, opened 2020-07-01
Fixes autocomplete suggestion

### LINKED ISSUE #917: Autocomplete fields go blank when submitting filter form
Steps to reproduce:

1. On the [All Incidents](https://pressfreedomtracker.us/all-incidents/) page enter a tag:

   image

2. Press "Apply Filters"

3. Note that field goes blank

   image

The filtering appears to still work, but the blank token is unsettling.

SaptakS:
Fixes #917 

PS: I found out that fetching the details for a tag (title, etc.) are done by the Autocomplete component. So when the URL already has the params `?tags=1` for example, then the [constructor for the Autocomponent component calls the `fetchInitialValues()`](https://github.com/wagtail/wagtail-autocomplete/blob/master/client/src/components/AutocompleteInput/AutocompleteInput.js#L28) hence loading the details. Whereas when selected from the dropdown, the dropdown list fetched, already has the title attribute for each tag dictionary. However, after I click the "Apply Filter" it re-renders everything (now the tags value is {id: 1} because of re-rendering) and Autocomplete also re-renders but the fetchInitialValues isn't called. Hence there is no title attribute present, and hence missing text.

The fetchInitialValues should be called even during re-rendering to get the other attributes like title. Now I have included the directly in `get value()` function of Autocomplete component. Might make sense to put that in render() instead.

The other way would be, on re-rendering, do something similar to fetchInitialValues of Autocomplete in the IncidentFilter component. But since the fetchInitialValues is already being used in case the URL has the params, I though of using the same function in the same component. Not sure if it's the best idea.

@chigby @harrislapiroff  if it looks good to you, I will add some kind of jest test to check that the autocomplete values have the title attribute.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -8,6 +8,9 @@ module.exports = api => {
           '@babel/preset-env',
           {
             modules: false,
+            "targets": {
+              "node": "current"
+            },
           },
         ],
         '@babel/preset-react',
--- a/client/autocomplete/js/components/Autocomplete.js
+++ b/client/autocomplete/js/components/Autocomplete.js
@@ -33,6 +33,9 @@ class Autocomplete extends PureComponent {
 
 	get value() {
 		if (this.props.controlled) {
+			if (!!this.props.value && !this.props.value[0].title) {
+				this.fetchInitialValues(this.props.value)
+			}
 			return this.props.value
 		} else {
 			return this.state.value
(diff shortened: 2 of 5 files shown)


chigby (approved):
These tests look good to me, in that they test the specific problem that was fixed in the other half of this PR, and also they provide a nice basis for anyone wanting to build out further tests of this component. I'm approving of it.

Issue #957, opened 2020-08-31
Spec out architecture for user friendly statistics entry


harrislapiroff:
As a subtask of #791, please spec out a proposed implementation of this feature. I'm guessing this will likely take the form of a draftail plugin and potentially a widget or field for things like stat boxes (or maybe just converting those to limited feature rich text fields for minimal maintenance surface?)

harrislapiroff:
We implemented this feature, but there's a few things we might want to improve:

1. The feature doesn't work in dark mode (light text on light background)
2. The feature is only on incident pages and we probably most want it on blog pages
3. We want this same panel for the stat fields on category pages as mentioned in #791 

Should check with @KAMcCudden if this would be sufficiently useful to fix up

harrislapiroff:
@KAMcCudden reports that this is not a feature they're using now, but it _would_ be useful on Analysis (blog pages) if we fixed it up.

Jendunham122:
This was demo’d to the team yesterday, and we love it. One question: is it possible to have a hyperlink to the filter link that users can click on?

Example: **At least XX journalists** [stat bloc + hyperlink] were detained while covering the protest.

harrislapiroff:
@Jendunham122 Yeah, let's make that happen. I'll add it as completion criteria

harrislapiroff:
Upgraded impact to high because the team is excited about it

PR #962, opened 2020-09-23
Draftail plugin for entering stats by search URL

chigby:
This pull request adds a Draftail extension for [managing a new entity](https://docs.wagtail.io/en/v2.7.1/advanced_topics/customisation/extending_draftail.html#creating-new-entities) that wraps our custom statistics tags.

It works like this:

1. A "Stats" button is added to the Rich Text options
2. Pressing this button opens a dialog box where you can select a "dataset" (corresponding to the particular statistic tag you want) and paste a search URL constructed using the filter UI on the front-end of the PFT site.
3. After inputting this information and clicking "Save," the extension queries a new summary API to get the current value for the statistic. 
4. The resulting number is inserted into the rich text field as a clickable, editable entity. On the back end, this is represented by a statistics tag with the parameters that correspond to the dataset chosen and search parameters from the URL.

Some potential further work:

* Better design of the dialog box and/or the entity's representation in the rich text box
* Limiting which rich text fields this button appears in.
* Smoothing out the UX in other ways 

Refs #957


SaptakS (approved):
Code,  functionality and tests look good to me. @harrislapiroff if you are okay with the UI/UX, this should be ready to merge I think.

tracker/settings/dev.py
```diff
@@ -79,3 +83,6 @@ def get_default_gateway_linux():
             'CONN_MAX_AGE': os.environ.get('DJANGO_DB_MAX_AGE', 600)
         }
     }
+
+# Prevent endless waiting if problem loading webpack bundles.
+WEBPACK_LOADER['DEFAULT']['TIMEOUT'] = 15  # noqa: F405
```
harrislapiroff: This might need to be a little longer. I couldn't start up on my computer without getting timeout errors. Setting it to `60` worked for me.

harrislapiroff (changes requested):
This looks fantastic. One change specified inline and a few other visual changes requested, then let's get it up on staging!

- [x] Modal max with 600px
- [x] Modal background #FFF
- [x] Modal border-radius 4px
- [x] More space between radio buttons and their labels
- [x] If it’s possible to get better vertical alignment of the buttons and their labels (not shown in the screenshot) do it, but no need to spend more than 5 minutes on it
- [x] Help text for the URL: “Perform a search on the main site for incidents to include in this statistic and copy the URL here.”  (Maybe even link "main site")

image

chigby:
@harrislapiroff Should be ready to review again, now

harrislapiroff (changes requested):
Looks like we need to verify that places we allow stats tags also *parse* them (I don't suppose this can be integrated into the wagtail rich text parsing that handles embeds and images?):

image

I'm also noticing some very buggy behavior when trying to type after adding a stat that I don't know how to characterize, so here's a gif

Screen Recording 2021-01-26 at 5 45 21 PM

First the cursor went to the beginning of the field after entering the stats token, next I trued to move it manually to the end of the field but as soon as I typed something the entire field changed to just the stat number twice and the cursor moved back to the start. Very odd!

I also noticed that the modal doesn't seem to validate whether I've selected a radio or not? I'd suggest having it default to "Total" and maybe having it validate if none are selected.

And one last thing that I'll call a stretch goal (don't bother if it will take more than 30m)—there's a noticeable delay between hitting "save" and the modal being dismissed. Could we add a spinner or some other loading indicator during that?

- [x] template-side processing of stats in all necessary locations
- [x] weirdness around typing after entering a stat
- [x] radio button default and maybe validation
- [x] **stretch goal:** loading indicator

chigby:
Still kind of perplexed by the typing weirdness. I'm wondering if it's possible that I'm misusing `TooltipEntity` from wagtail? I'm also wondering if it's somehow related to exactly what text is inserted at the `Modifier.replaceText` call, but I've tried a lot of things there and it doesn't really affect the behavior of the textbox in a consistent way, or possibly even affect it at all.

chigby:
I actually think my latest commit fixed the editing behavior & typing oddness

chigby:
Added a (simple) textual loading indicator when the "Save" button is activated on the modal.

harrislapiroff (approved):
All the fixes work! Let's get it on staging! And also open a ticket for creating a stat widget for getting this behavior outside of rich text fields.

PR #973, opened 2020-11-18
HTTPS Analytics

chigby:
This change implements most of what we did in the SDO repository (https://github.com/freedomofpress/securedrop.org/pull/771). Essentially we require our analytics assets to be loaded over HTTPS, even when browsing in HTTP (as in Tor).

This change replaces the `django-analytical` package with our own JS file that contains all the necessary information and rendered according to a single boolean variable in the settings file.


SaptakS (approved):
Looks working to me

SaptakS (approved):
LGTM

PR #980, opened 2020-12-01
Restrict incident body images to width of container

### LINKED ISSUE #979: Image overflow in IncidentPage body field
When adding a wide image as an image block in the body field of an Incident Page, the image displays at full width and extends beyond the width of the text area.

Example: https://pressfreedomtracker.us/admin/pages/742/edit/

image

We have CSS to avoid this problem on stream pages and blog pages, seems like it's time to apply these styles to incident pages as well.

Reported by @KAMcCudden

chigby:
This pull request limits the width of images in the body field of incident pages to be at most the width of the body container.

Fixes #979


SaptakS (approved):
Looks good to me.

PR #984, opened 2020-12-03
Cleans up default settings files

conorsch:
Ensures that a SECRET_KEY value will be declared via env var in
production contexts. Clarifies same in the dev settings file.
Removes entirely the unused Elasticsearch config.

Refs https://github.com/freedomofpress/tracker/milestone/1


maeve-fpf (approved):

SaptakS (approved):
Codewise looks good to me. Approving and merging based on review by @maeve-fpf

Issue #647, opened 2018-03-13
Make all country and organization references customizable


eloquence:
As part of the open source release, we should ensure that it is straightforward to customize logo and text references to a specific country or organization.

# User stories

As a PFT administrator, I want to be able to customize the PFT for a specific country or region, and for a specific organization, so that it serves my specific use case.

As the Freedom of the Press Foundation, we want to avoid confusion about who runs a given PFT instance, so that we can avoid reputational issues or miscommunications.

harrislapiroff:
It's no longer on our roadmap to make the PFT codebase white-labeled, so, closing this.

PR #991, opened 2020-12-15
Add build/deploy notes

maeve-fpf:
Addresses https://github.com/freedomofpress/tracker/issues/646. This is a pretty high-level summary, but should give us something to work with that we can improve with real-life use.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -207,6 +207,60 @@ helpful to restore your snapshot from develop, so that the migrations
 for the new branch, which were presumably based off of develop, will
 have a clean starting point.
 
+Deployment
+=============
+
+Building
+-------------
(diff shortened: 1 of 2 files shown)


conorsch (changes requested):
Wonderful, looks great, @maeve-fpf ! We can clarify upon request, as you say in-line. The only suggestion I have for now is that we mention #647 in the build docs, so folks are aware that it's not really fully customizable for non-US environments yet. Might help guide folks to pinging us on that issue, if they're interested in spinning up an instance.

conorsch (approved):
Marvellous, thanks for accommodating, @maeve-fpf =)

PR #992, opened 2020-12-15
Remove site URI from API image URL responses

### LINKED ISSUE #988: API generating bad urls for images
It seems like something in our API was assuming media paths were relative. Now that we're using a separate domain for media files, we're getting urls like this in API responses:

```
https://pressfreedomtracker.ushttps://media.pressfreedomtracker.us/media/images/Munhoz_201104_0570.2e16d0ba.fill-1330x880.jpg
```

where we should have

```
https://media.pressfreedomtracker.us/media/images/Munhoz_201104_0570.2e16d0ba.fill-1330x880.jpg
```

Here's a fairly minimal API query you can use to reproduce this bug: https://pressfreedomtracker.us/all-incidents/export/?tags=111&fields=title,slug,teaser_image,categories,date

chigby:
Fixes #988 

This pull request removes the site `root_url` from being prepended to image URL fields in API responses (both CSV and JSON formats).

We are in the situation where our production environment is different from the development environment because our media (i.e. our `MEDIA_ROOT` setting) is stored on a different domain in production, so the URLs attached to the images are _always_ absolute in production. 

I've made the decision here to alter the results for these image fields so that we only ever return the URL that Django/Wagtail have associated with the image. Previously, I think we were trying to be clever so that we could get absolute URLs in development mode. But I think the value of correct data outweighs the value of absolute URLs in development mode. It's also not necessarily clear that the consumer of this data in dev mode will want an absolute URL at all. So that use case is being removed, essentially. Which I think is fine.

### DIFF
--- a/incident/models/export.py
+++ b/incident/models/export.py
@@ -92,11 +92,12 @@ def to_json(obj, allowed_fields=[]):
 
 
 def _serialize_field(obj, field):
-    site = obj.get_site()
     if field.name == 'teaser_image':
-        val = getattr(obj, field.name)
-        if val:
-            val = site.root_url + val.get_rendition('fill-1330x880').url
+        teaser_image = getattr(obj, field.name)
+        if teaser_image:
+            val = teaser_image.get_rendition('fill-1330x880').url
+        else:
+            val = None
     elif field.name == 'slug':
         val = obj.get_full_url()
     elif type(field) == models.ForeignKey:
--- a/incident/models/topic_page.py
+++ b/incident/models/topic_page.py
@@ -33,9 +33,8 @@ class IncidentSchema(Schema):
     description = fields.Method('get_description')
 
     def get_image(self, obj):
-        site = obj.get_site()
-        if obj.teaser_image and site.root_url:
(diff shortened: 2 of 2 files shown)


harrislapiroff:
Hmm... you don't think there's an elegant way to _always_ get absolute URLs is there? Or would the solution to that be to provide an absolute URL as the `MEDIA_URL` setting? I agree that it probably doesn't matter for dev, but I fear running into this again if we ever switched back to a shared domain for media/htmls—unlikely, though.

chigby:
The only way I can think of to do it is look at the URL returned by the image object and:

1. if the URL is absolute (based on some criteria), then return that
2. if it's a relative URL, prepend the site domain. This is what we were doing before in all cases.

chigby:
I think the reason I leaned against this type of solution in the first place was looking up the site does have the cost of a db query, though I suppose in production we would expect to never run into that case.

harrislapiroff:
Fair enough! Let's go with it as is then! Thanks for indulging my hypothetical 😇

harrislapiroff (approved):
Seems to do what it says—let's get it on staging!

Issue #986, opened 2020-12-07
Make PFT settings consistent with other apps


maeve-fpf:
Following up on https://github.com/freedomofpress/tracker/pull/984: There are a few quirks in each app's settings that we'd like to make a little more unified. Things we can do on PFT:

- Instead of `DJANGO_NO_DB`, check for the presence of `DJANGO_DB_HOST` like other apps
- Create a setting for `DJANGO_ONION_HOSTNAME` (in case we want to use it in a template, i.e. footer link)
- Stub out a CSP variable, to add entries later
- Instead of `DJANGO_LOG_CONSOLE` and `DJANGO_LOG_PATH`, just log to console if a log path is not set. Don't require `DJANGO_LOG_JSON`, default it appropriately.
- Treat env vars that should be "booleans" consistently (I like `.get` without a default, for type soundness)

SaptakS:
@maeve-fpf @harrislapiroff this issue can be closed?

eloquence:
Closed per web sync today.

PR #995, opened 2020-12-15
Make some settings consistent with other apps

maeve-fpf:
Works towards #986 

Another PR in the spirit of https://github.com/freedomofpress/securedrop.org/pull/786 and https://github.com/freedomofpress/securethenews/pull/299 - we also needed to upgrade our logging dependencies here.


chigby (approved):
Changes look good to me, very similar to the situation over on STN and they seem to work here as described.

Issue #1004, opened 2021-02-01
Investigate improving API


chigby:
[Goals and potential approaches document](https://docs.google.com/document/d/1-u5lYT24XZwP0MRQu0NE2B9coZWVAfnk8f0F1GXuDRo/edit?ts=6018736f#)

Refs #1007

chigby:
Some insights gained via some profiling I've done with [django-silk](https://github.com/jazzband/django-silk).  Most of the slownesses I found were related to database queries being executed.

1. The `export` view is not very well optimized to minimize DB queries, particularly the JSON-formatted view. This probably should be fixed no matter what implementation route we want to take. But it is also slow because we are returning many incidents at once.
2. Retrieving image renditions, e.g. for teaser images, requires querying the database.
3. Rendering `StreamFields` is somewhat performance intensive. The process can include making DB queries for things like images, which is something I didn't realize before.
4. Getting the URL of a page from wagtail involves querying the database at least once. 
5. The topic page API view is fairly well optimized but still involves rendering streamfields and finding URLs. We are helped by the fact that the number of incidents returns is limited.

I'm inclined to see what we can gain by trying to pre-generate (as described in the approaches document) the export view. This technique is less disruptive to our codebase than bringing on board something like Django Rest Framework, and I think we could accomplish a lot of our stated goals with a little thought to how we want to design some different, yet limited, datasets to pre-generate. 

The topic page view I'm less certain about, and we should discuss more.

chigby:
Having deployed one batch of performance improvements to the current export API, here's my current thinking on where we ought to look next.

Given that:

- In production, we are using the export API (with filters and a subset of fields) to power the data viz. We're achieving acceptable performance via Cloudflare caching.
- The topic page API and the export API use almost completely distinct code.
- The topic page API might be better thought of as the Categories API, since it primarily delivers information about our incident categories.
- The export API primarily delivers information about Incidents themselves.
- There is a limit to our ability to optimize exporting all incident information, particularly the article text fields. These are always going to be somewhat slow and that slowness will be linearly proportional to the number of incidents being exported.

[...]

PR #1008, opened 2021-02-10
Export performance and profiling

chigby:
This pull request:

1. Adds two profiling tools to the dev environment.  I have been finding them somewhat helpful in diagnosing where performance slowdowns are occurring in the API views. There is some overlap between what they offer and the django-debug-toolbar, but the latter package does not work, as far as I know, with views that return non-HTML content, such as JSON. So I don't think it's redundant to have these around. I've disabled them by default, but they can be activated via environment variable, like this:  `DJANGO_PROFILE=yes docker-compose up` 
2. Adds a number of prefetch and select related modifiers to the queryset in the incident export view.  This significantly reduces the number of queries needed per result. I would like to be able to reduce it more, but wagtail doesn't seem to allow prefetching image renditions that occur inside streamfields (probably because streamfields are stored as JSON internally, and thus it's hard to figure out what images would need to be fetched at query-time). This change should not change the output of the API at all, just the performance.

Any comments or things you notice about this are appreciated.

Refs #1004

### DIFF
--- a/incident/models/export.py
+++ b/incident/models/export.py
@@ -102,10 +102,13 @@ def to_json(obj, allowed_fields=[]):
 def _serialize_field(obj, field):
     if field.name == 'teaser_image':
         teaser_image = getattr(obj, field.name)
+        val = None
         if teaser_image:
-            val = teaser_image.get_rendition('fill-1330x880').url
-        else:
-            val = None
+            for rend in teaser_image.renditions.all():
+                if rend.filter_spec == 'fill-1330x880':
+                    val = rend.url
+            if not val:
+                val = teaser_image.get_rendition('fill-1330x880').url
     elif field.name == 'slug':
         val = obj.get_full_url()
     elif type(field) == models.ForeignKey:
--- a/incident/models/incident_index_page.py
+++ b/incident/models/incident_index_page.py
@@ -74,7 +74,26 @@ class IncidentIndexPage(RoutablePageMixin, MetadataPageMixin, Page):
         else:
             allowed_fields = allowed_fields.split(',')
         incident_filter = IncidentFilter(request.GET)
-        incidents = incident_filter.get_queryset()
+        incidents = incident_filter.get_queryset() \
+            .select_related('teaser_image', 'state', 'arresting_authority') \
+            .prefetch_related(
+                'authors',
+                'categories__category',
+                'current_charges',
+                'dropped_charges',
+                'equipment_broken__equipment',
+                'equipment_seized__equipment',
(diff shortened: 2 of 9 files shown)


incident/models/export.py
```diff
@@ -102,10 +102,13 @@ def to_json(obj, allowed_fields=[]):
...
+        val = None
         if teaser_image:
-            val = teaser_image.get_rendition('fill-1330x880').url
-        else:
-            val = None
+            for rend in teaser_image.renditions.all():
+                if rend.filter_spec == 'fill-1330x880':
+                    val = rend.url
+            if not val:
+                val = teaser_image.get_rendition('fill-1330x880').url
```
harrislapiroff, line 111: I'm so fascinated that this is faster!

incident/models/incident_index_page.py
```diff
@@ -74,7 +74,26 @@ def export_view_GET(self, request: 'HttpRequest') -> HttpResponse:
...
+                'politicians_or_public_figures_involved',
+                'tags',
+                'target_nationality',
+                'targeted_institutions',
+                'targeted_journalists',
+                'teaser_image__renditions',
+                'updates',
+                'venue',
+                'workers_whose_communications_were_obtained',
+        )
```
harrislapiroff, line 96: image

tracker/urls.py
```diff
@@ -50,3 +51,6 @@
                        ] + urlpatterns
     except ImportError:
         pass
+
+if apps.is_installed('silk'):
```
harrislapiroff, line 55: TIL

harrislapiroff (approved):
These profiling tools are incredible and make me want to ask you to optimize everything now. Only way they could be better is if they auto-generated an interactive flame chart 🤔 

Let's get this up on staging an see what the improvements look like with real data!

incident/models/export.py
chigby: It's faster because the for loop can examine prefetched renditions, apparently!

harrislapiroff:
Should we try to do some speed testing on the live or staging site before we merge this? Do we have an easy way to do that? (since it would require bypassing the cf cache—though we could add a randomized querystring or something to achieve that)

chigby:
I'd support testing this out in some way.. if Cloudflare is caching it by URL and appending stuff to the querystring that will be ignored by the export view will bypass that, then it seems like a decent way to get some before-and-after data.

harrislapiroff:
Recorded some data on current timings of the API on staging. Gonna merge and compare!

PR #1014, opened 2021-03-03
Add prefetching to incident queryset for incident index page

chigby:
This is identical to the query changes made to the export views in https://github.com/freedomofpress/pressfreedomtracker.us/pull/1008


harrislapiroff (approved):
Looks good and I didn't notice any issues on local. Let's get it on staging and see how it goes!

PR #1019, opened 2021-03-16
Initial API for incidents

chigby:
This pull request sketches out an initial implementation of a Django Rest Framework-based API for incidents. This is designed to replace the functionality of the export views on the `IncidentIndexPage` models. I've also included tests for this, some improvements to the incident data factories (and some bug fixes), and some enhancements to the `createincidents` management command to make creation of related objects (Authors, Venues, etc.) easier.

The URL is `/api/v2/incidents`, though this can be changed. Actually, all of this can be changed. The main point of the PR is to lay out a basis for a path forward from our current API to an improved one.

Features the API in this PR include:

* Filtering. Uses the same filter code as the rest of the site.
* Dynamic fields. You have the power to choose which fields are returned with the `fields=field1,field2` query parameter.
* Structured relation data. Where possible, I've tried to give the JSON data more structure when it comes to data "related" to incidents, such as tags, equipment, or targets.
  * Lists of names are rendered as JSON lists
  * Links and equipment damaged/seized are rendered as JSON objects. Formerly, we might render equipment damaged as `"iphone: count of 2"`, but now we can have `{equipment: "iphone", quantity: 2}`
  * Empty/missing values are properly given the JSON value of `null`
* CSV support. 
    * Using the query parameter `format=csv` will return the same data as a CSV file
    * Here I have attempted to maintain the unstructured. "flat" nature of the current CSV export.
    * Though this is not the default behavior of the DRF CSV renderer. We might want to consider if we want to provide a different format for our CSV. I have no problem with what we have now, but this is an opportunity to change it if we want to. 

Features not included yet, but warrant discussion:

* Pagination. Our current API does not have this and there are a few different possible ways to do it.

It's possible to play around with the new incidents commend with:
```
docker-compose exec django ./manage.py createincidents --help
```

Refs #1004

### DIFF
--- a/incident/tests/test_pages.py
+++ b/incident/tests/test_pages.py
@@ -19,6 +19,7 @@ from common.tests.factories import (
     CategoryPageFactory,
     PersonPageFactory,
     CommonTagFactory,
+    CustomImageFactory,
 )
 from home.tests.factories import HomePageFactory
 from incident.models.incident_page import IncidentPage
@@ -27,9 +28,11 @@ from incident.models.export import is_exportable, to_row
 from .factories import (
     IncidentIndexPageFactory,
     IncidentPageFactory,
+    IncidentLinkFactory,
     IncidentUpdateFactory,
     TopicPageFactory,
     StateFactory,
+    VenueFactory,
 )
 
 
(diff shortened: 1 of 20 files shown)


incident/tests/test_pages.py
```diff
@@ -702,3 +705,60 @@ def test_sorts_incidents_by_incident_date_descending(self):
...
+            IncidentLinkFactory.create_batch(3, page=incident)
+
+            # Pre-generate renditions to avoid INSERT queries that we
+            # don't want to count.
+            incident.teaser_image.get_rendition('fill-1330x880')
+
+    def test_api_selects_and_prefetches(self):
+        # 1 base query, plus 20 prefetches.
+        with self.assertNumQueries(21):
+            list(IncidentPage.objects.with_public_associations())
```
harrislapiroff, line 764: Wow fancy performance test

tracker/urls.py
```diff
@@ -22,6 +23,7 @@
 urlpatterns = [
     path('django-admin/', admin.site.urls),
 
+    path('api/v2/', include(api_urls)),
```
harrislapiroff, line 26: Maybe we should put this at `api/beta/` or `api/edge/` for the moment so as not to make promises about its permanence?

harrislapiroff (commented):
Code looks good! Moving on to testing locally.

harrislapiroff:
I do think, especially if we're going to dogfood this API everywhere, we will want pagination. I'm thinking in particular of Topic Pages where we want to display the first N incidents of every category matching a certain tag, but we don't necessarily want to have to load ALL incidents from every category matching a certain tag.

This could definitely wait for a future PR though.

harrislapiroff (approved):
The basic functionality works on my computer, but response times are pretty slow and make more detailed testing frustrating. Let's get it up on staging to see how response times are there and do some more detailed testing!

Issue #1013, opened 2021-03-03
Page load timeout on certain Incident Index page requests


chigby:
Example URL: https://staging.pressfreedomtracker.us/all-incidents/?categories=10&date_lower=2020-05-25&search=protest

Brief attempts to reproduce locally yield page load times between 5 and 10 seconds, with 200+ queries to the DB recorded. The incident index page HTML view did not receive the recent performance improvements that were made to the HTML view (see #1008), so adding that in may help.  On the other hand, I'm not sure DB query counts completely explain what might be happening here. The API views we improved were performing thousands of queries (because the export is not paginated, as the HTML views are) and still managed to render in around 30 seconds. So some more investigation might be needed.

harrislapiroff:
Thanks for the write-up here, @chigby. This would definitely be good to continue exploring.

conorsch:
Some related work is presented in # #1014, with @harrislapiroff to review next.

maeve-fpf:
Here's some more URLs this has happened on since the last update in infrastructure#3188:

- `/all-incidents/?categories=10&date_lower=2020-05-26&endpage=3&page=2&search=protest`
- `/blog/feed/`
- `/all-incidents/?date_lower=2020-05-26&categories=10&search=protest&endpage=3&page=3`
- `/all-incidents/feed/`

So, no surprises. I think both the feeds and the incidents should be addressed by the change in #1014 but if not let me know. Would it be useful to deploy that branch to the staging environment before merging it to `develop` to verify? That's a stopgap to having a `manage.py shell` that we could do.

harrislapiroff:
Still getting timeouts on the incident queries, even after merging #1014. (I wasn't able to reproduce the errors at the feed endpoints, even prior to merging.) 🙁

harrislapiroff:
It appears that #1021 resolved the timeouts at those URLs! Nice work, @chigby!

maeve-fpf:
Checking logs - unfortunately, I see some today post-deploy. Worker restarts were logged at

- Mar 30, 2021 @ 14:56:55.494
- Mar 30, 2021 @ 14:54:39.121
- Mar 30, 2021 @ 14:53:50.116

However, I don't see a corresponding nginx message with a URL for them.

maeve-fpf:
Found some URLs for timeouts today:

- 09:16:40 `/george-floyd-protests`
- 01:54:17 `/election-2020`
- 01:53:40 `/george-floyd-protests`

[...]

PR #1021, opened 2021-03-22
Improve targeted journalist/institution summary query

chigby:
This change replaces our use of `total_queryset` with `incident_ids`, i.e. replacing an entire queryset (which has the potential to contain some complex filters) with a list of IDs, which we query once and then save as a Python value.  This causes the Django ORM to use an `IN (<static list of ids>)` query rather than `IN (<potentially complicated subquery>)`.

In particular I think this saves us a lot of trouble on the institution summary, which Django was running as a left join with two
very large `OR` expressions in the where-clause.  But all the complexity in those expressions was due to the filtering, which we can change to a static list of IDs and get the same results with a lot less number crunching.

Part of an ongoing effort to solve #1013

### DIFF
--- a/incident/utils/incident_filter.py
+++ b/incident/utils/incident_filter.py
@@ -721,20 +721,20 @@ class IncidentFilter(object):
             date__year=TODAY.year,
         )
 
-        total_queryset = queryset
+        incident_ids = list(queryset.values_list('pk', flat=True))
         num_this_year = incidents_this_year.count()
         num_this_month = incidents_this_month.count()
 
-        tj_queryset = TargetedJournalist.objects.filter(incident__in=queryset)
+        tj_queryset = TargetedJournalist.objects.filter(incident__in=incident_ids)
         tj_inst_queryset = tj_queryset.filter(
             institution_id=OuterRef('pk')
         ).values_list('institution_id', flat=True)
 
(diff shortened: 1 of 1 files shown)


chigby:
Some stats for the request http://localhost:8000/all-incidents/summary/?categories=10&date_lower=2020-05-25&search=democrat 

Using the `/summary/` endpoint because it isolates the particular function that's being improved here, but the same code is used in the main incident index HTML view as well.

Before these changes:

image

With these changes:

image

And here's about how much data I'm working with locally. Note that this is not using the staging database copy. It's using factory-generated random data, hence the essentially random search term.

```
trackerdb=# SELECT  (
        SELECT COUNT(*)
        FROM incident_incidentpage
        ) AS incident_count,
        (
        SELECT COUNT(*)
        FROM incident_incidentpage_targeted_institutions
        ) AS targeted_institution_count,
        (
        SELECT COUNT(*)
        FROM incident_targetedjournalist
        ) AS targeted_journalist_count        
;
 incident_count │ targeted_institution_count │ targeted_journalist_count 
────────────────┼────────────────────────────┼───────────────────────────
           4794 │                       9588 │                      1332
(1 row)
```

SaptakS (approved):

PR #1023, opened 2021-03-23
Update to Python 3.9

chigby:
This PR updates the development environment, CI environment, and production environment to use python 3.9. As a result, I also updated the psycopg2 package to 2.8.x, as the previous 2.7.x series was not compatible.


harrislapiroff (approved):
Code here looks good and tests are passing. Let's get it on staging 👍

PR #1031, opened 2021-04-05
Improve topic page performance

chigby:
This pull request attempts to decrease the topic page load time, mainly by improving two fairly inefficient SQL queries. The first, and slowest, one had a very gnarly `GROUP BY` clause that was not needed at all, probably caused by a bug in Django. The second one was something I originally wrote as a subquery to enforce a `LIMIT` amount in the query itself and work around the limitations of prefetching. It turns out that the SQL emitted by this technique is not good, and we can save ourselves by doing the limit in python. More explanation can be found in the commit messages.

Refs #1013

I tested this locally with the data from our staging database, here's some comparisons:

Before:
image

After:
image

### DIFF
--- a/incident/models/topic_page.py
+++ b/incident/models/topic_page.py
@@ -25,6 +25,11 @@ from common.utils import (
 from incident.models import IncidentCategorization
 
 
+class NongroupingSubquery(models.Subquery):
+    def get_group_by_cols(self):
+        return []
+
+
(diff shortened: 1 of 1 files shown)


harrislapiroff (approved):
I don't understand all the changes here, but it looks good from what I do understand! Let's get it on staging and see that everything behaves the same!

PR #1035, opened 2021-04-12
Add default value for state in API CSV format

chigby:
This pull request fixes a problem in the new API's CSV format where if an incident did not have a `state` relationship, it would throw an exception. I've fixed it by adding a default string value (`''`, the empty string) for this field.


harrislapiroff (approved):
This fixed the issue for me! Let's get it on staging!

PR #1036, opened 2021-04-13
API pagination, URL rename, endpoint listing

chigby:
This pull request:

1. Moves the root API URL to `/api/edge`
2. Adds a view at that root URL that displays all endpoints
3. Adds pagination. This is done using DRF's cursor-pagination system which I think will give us the best performance and efficiency. In order to get the maximum benefit from this, I added a new field `IncidentPage.unique_date` that serves as the ordering field for our pages. More information on this is available in the commit message.

Refs #1004


incident/api/views.py
```diff
@@ -1,14 +1,23 @@
...
+from rest_framework_csv.renderers import PaginatedCSVRenderer
 
 from incident.api.serializers import IncidentSerializer, FlatIncidentSerializer
 from incident.utils.incident_filter import IncidentFilter
 
 
+class IncidentCursorPagination(CursorPagination):
+    page_size = 25
+    page_size_query_param = 'limit'
+    max_page_size = 200
```
harrislapiroff: Is it possible to set no maximum? I want API consumers to be able to get the whole database in one request if they *want* but I also want them to have the option of not getting the whole database.

chigby: Yes, I think that's totally possible! I wasn't sure what our wants/needs were exactly in this regard so I just made something up here.

harrislapiroff (approved):
These look great! Let's get 'em on staging and test with real data!

harrislapiroff:
A couple things I want to mention I noticed:

- Pagination seems to include a pagination wrapper iff `limit` is specified.
   ```json
   {
         "next": "http://localhost:8000/api/edge/incidents/?cursor=cD0yMDIxLTAyLTA2LTMxMw%3D%3D&limit=10",
         "previous": null,
         "results": [
              {
                   ...
              }
         ]
    }
    ```
  Otherwise it returns just the `results` unwrapped. This does make a certain amount of sense, but I'm not sure whether it's better to make the base case (fetching all incidents) simplified that way versus making response format consistent. Worth doing some more thinking about.
- Pagination and CSV format interact weirdly. Since there's no wrapper to tell you the `next` URL, there's no way to paginate the CSV. I'm not sure if we should try to come up with a different way for CSV pagination to work, leave it as is so you can still fetch the first N incidents, or disable pagination for CSV...

harrislapiroff:
Still gonna merge as is though.

PR #1044, opened 2021-04-30
Add CORS headers to new API

### LINKED ISSUE #1043: Add `access-control-*` headers to the new API to allow cross-origin loading (CORS)
We seem to have these on the old API but not the new one? Reproduction here: https://observablehq.com/@harrislapiroff/press-freedom-tracker-incident-fetching

harrislapiroff:
Resolves #1043

### DIFF
--- a/incident/api/views.py
+++ b/incident/api/views.py
@@ -1,3 +1,5 @@
+from typing import TYPE_CHECKING
+
 from rest_framework import viewsets
 from rest_framework.settings import api_settings
 from rest_framework.pagination import CursorPagination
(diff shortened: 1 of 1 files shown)


maeve-fpf:
Headers look correct, I don't know how to test locally though. (I assume the typing is required to override this method?)

harrislapiroff:
The typing is just because I like typing 😅. You can load up `/api/edge/incidents/?limit=10` (faster to use a limit than load the whole database) locally and use your browser's web inspector network panel to look for the correct headers.

maeve-fpf:
Looks good!

PR #1046, opened 2021-05-02
feat: Add Content Security Policies

### LINKED ISSUE #1024: Add content security policies
Like the other repos, we would need to add CSP rules using django-csp for this repo as well.

### LINKED ISSUE #1028: Google Analytics request on page with data visualizations
# Steps to reproduce
1. Open a private browser window
2. Open the network panel of the developer console
3. Ensure that all ad blocking and tracking blocking is disabled (including the "shield" icon in Firefox)
4. Load https://pressfreedomtracker.us/george-floyd-protests/

# Expected behavior

No third party analytics scripts are loaded

# Actual behavior

Google Analytics is loaded

# Additional notes

If this cannot be disabled, it could perhaps be mitigated via CSP (#1024) to exercise more explicit control over these requests

prateekj117:
Fixes #1024 
Fixes #1028

Adds Content Security Policies similar to [securedrop.org](https://github.com/freedomofpress/securedrop.org/blob/develop/securedrop/settings/base.py#L313-L367) and [securethenews](https://github.com/freedomofpress/securethenews/blob/develop/securethenews/settings/base.py#L251-L303).

### DIFF
--- a/tracker/settings/base.py
+++ b/tracker/settings/base.py
@@ -276,19 +276,32 @@ CSP_FRAME_SRC = (
     # For Twitter Widgets
     "https://platform.twitter.com",
 )
-CSP_CONNECT_SRC = (
+CSP_CONNECT_SRC = [
     "'self'",
     "https://analytics.freedom.press",
     # For Wagtail Admin
     "https://releases.wagtail.io/latest.txt",
-)
-CSP_IMG_SRC = (
+]
+CSP_IMG_SRC = [
     "'self'",
-    "analytics.freedom.press",
+    "https://analytics.freedom.press",
     "https://media.pressfreedomtracker.us",
-    # For Wagtail Admin
-    "http://www.gravatar.com",
-)
+    # For Twitter Widgets
+    "https://platform.twitter.com",
+    "https://syndication.twitter.com",
+    "https://pbs.twimg.com",
+    "https://ton.twimg.com",
+    "data:",
+]
+CSP_OBJECT_SRC = ["'self'"]
+CSP_MEDIA_SRC = ["'self'"]
+
+if os.environ.get("DJANGO_CSP_MEDIA_ORIGINS"):
+    csp_media_origins = os.environ["DJANGO_CSP_MEDIA_ORIGINS"].split()
+    CSP_MEDIA_SRC.extend(csp_media_origins)  # video files
+    CSP_IMG_SRC.extend(csp_media_origins)
+    CSP_OBJECT_SRC.extend(csp_media_origins)
+    CSP_CONNECT_SRC.extend(csp_media_origins)
 
 # Report URI must be a string, not a tuple.
 CSP_REPORT_URI = os.environ.get('DJANGO_CSP_REPORT_URI',
--- a/tracker/settings/base.py
+++ b/tracker/settings/base.py
@@ -101,6 +101,9 @@ if os.environ.get('DJANGO_WHITENOISE'):
 MIDDLEWARE.extend([
     'wagtail.contrib.redirects.middleware.RedirectMiddleware',
     'django_logging.middleware.DjangoLoggingMiddleware',
+
+    # Middleware for content security policy
+    'csp.middleware.CSPMiddleware',
 ])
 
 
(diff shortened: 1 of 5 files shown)


prateekj117:
@SaptakS I know `unsafe-inline` in `script-src` and `style-src` is the worst way to go, but these are required for Wagtail Admin. I can be more explicit though in these policies and ignore `unsafe-inline`, but I see in CSP rules of [securethenews](https://github.com/freedomofpress/securethenews/blob/develop/securethenews/settings/base.py#L251-L303) as well, we went this way. Let me know if you want me to be more explicit.

SaptakS:
Yes. For Wagtail Admin, for now, we have to include unsafe-inline. All forms have inline scripts. There is an RFC by wagtail which is meant to fix this issue but not anytime soon, so that is fine for now. I will check the rest of the code.

tracker/settings/base.py
```diff
@@ -251,6 +254,45 @@
...
+)
+CSP_CONNECT_SRC = (
+    "'self'",
+    "https://analytics.freedom.press",
+    # For Wagtail Admin
+    "https://releases.wagtail.io/latest.txt",
+)
+CSP_IMG_SRC = (
+    "'self'",
+    "analytics.freedom.press",
```
SaptakS: Please add `https://` here

tracker/settings/base.py
```diff
@@ -251,6 +254,45 @@
...
+    "https://media.pressfreedomtracker.us",
+    # For Wagtail Admin
+    "http://www.gravatar.com",
```
SaptakS: You can remove this. This just adds the default gravatar in the admin section, which we don't care about, so doesn't make sense adding an entirely new src in CSP rules for just that.

tracker/settings/base.py
```diff
@@ -251,6 +254,45 @@
...
+)
+
+# Report URI must be a string, not a tuple.
+CSP_REPORT_URI = os.environ.get('DJANGO_CSP_REPORT_URI',
+                                'https://freedomofpress.report-uri.com/r/d/csp/enforce')
```
SaptakS, line 318: You will also need to set a section like this: https://github.com/freedomofpress/securedrop.org/blob/develop/securedrop/settings/base.py#L353-L361. Some of the CSP rules are set at the deployment time with environment variables. As part of that, you need to add a `CSP_OBJECT_SRC` with self for certain PDFs, etc.

SaptakS: You can actually just add a section like:

```
if os.environ.get("DJANGO_CSP_MEDIA_ORIGINS"):
    csp_media_origins = os.environ["DJANGO_CSP_MEDIA_ORIGINS"].split()
    CSP_MEDIA_SRC.extend(csp_media_origins)  # video files
    CSP_IMG_SRC.extend(csp_media_origins)
    CSP_OBJECT_SRC.extend(csp_media_origins)
    CSP_CONNECT_SRC.extend(csp_media_origins)
```

tracker/settings/base.py
```diff
@@ -251,6 +254,45 @@
...
+    # For Wagtail Admin
+    "https://releases.wagtail.io/latest.txt",
+)
+CSP_IMG_SRC = (
+    "'self'",
+    "analytics.freedom.press",
+    "https://media.pressfreedomtracker.us",
+    # For Wagtail Admin
+    "http://www.gravatar.com",
+)
```
SaptakS: With previous experience in other repos, I think you will need:

```
# Embeds
    'https://platform.twitter.com',  # Twitter
    'https://syndication.twitter.com',  # Twitter
    'https://pbs.twimg.com',  # Twitter
    'https://ton.twimg.com',  # Twitter
    'data:',  # Twitter makes use of data svgs
```
to be added to CSP_IMG_SRC to allow twitter images.

SaptakS (changes requested):
Thanks for the PR!!! I have left few comments mostly related to updates that we needed previously in other repos. Please let me know if you feel otherwise.

Also, there are certain pages in tracker website where we use data vizualisation which needs to allow other script and style src probably. I know @harrislapiroff is working on data viz so maybe he can let you know what are the CSP rules that need to be updated.

SaptakS (changes requested):
There are few more things that need to be added for everything to work.

SCRIPT_SRC:
- 'https://www.google.com/recaptcha/',
- 'https://www.gstatic.com/recaptcha/',
- 'https://cdn.jsdelivr.net',  # Observable Notebooks
- 'https://api.observablehq.com',  # Observable Notebooks
- 'https://bundle.run',  # Observable Notebooks

CONNECT_SRC:
- 'https://cdn.jsdelivr.net',  # Observable Notebooks
- 'https://pressfreedomtracker.us',  # Observable Notebooks

Also, please rebase the PR. After that, this should be ready for merging.

prateekj117:
@SaptakS Done.

SaptakS (approved):
Looks good to me.

PR #1048, opened 2021-05-04
Add middleware for tracking and reporting page generation time

chigby:
This pull request adds a fairly simple middleware that records how long a page took to fully render. This information is used to put a notice in the log if the time recorded was above a certain threshold. This information is also added to the response headers, which I thought might be useful sometimes for interactive debugging and information gathering, but might not be something we want.

I have two "log level" thresholds defined in the settings. My reason for doing this is I didn't want to overwhelm the logs with WARN level events when this was turned on, but I did want to set a fairly low threshold to start collecting data for a large sample of pages. If this is too granular, we can change it fairly easily. Also the exact thresholds can be adjusted if we don't think they'll be good for starting off. Right now an INFO level message is logged if a page takes longer than 3 seconds, and a WARN level message is logged if longer than 6 seconds.

Refs https://github.com/freedomofpress/fpf-www-projects/issues/174

### DIFF
--- a/common/middleware.py
+++ b/common/middleware.py
@@ -1,0 +1,25 @@
+import time
+import logging
+
+from django.conf import settings
+
+
+logger = logging.getLogger(__name__)
+
+
+def page_time_middleware(get_response):
+    def middleware(request):
+        start_time = time.perf_counter_ns()
+
+        response = get_response(request)
+
+        duration = (time.perf_counter_ns() - start_time) // 1_000_000
+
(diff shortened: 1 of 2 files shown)


SaptakS (approved):
Looks good to me code wise

chigby:
Closing in favor of https://github.com/freedomofpress/django-logging/pull/7

Will need to make a new PR in this repo to update the version of `django-logging` once that is merged.

PR #1052, opened 2021-05-10
Allow rich-text statistics widget to be used for stat boxes

### LINKED ISSUE #1006: Add a stats widget for embedding stats outside of rich text fields
In #962 we added a feature to Draftail that enabled editors to embed stats in rich text via a GUI. Let's now add a widget to embed stats in non-text-fields, for example the ones on the homepage statboxes

image

chigby:
Fixes #1006 

Stat boxes formerly stored their "value" as a char field, with a text
entry widget.  We commonly use them for statistics tags, so it would
be nice to be able to use our more user-friendly "numincidents" widget
with them.  This change updates stat boxes to store their "value" as
rich text (with a very limited feature set, including the incidents
widget) and display the value inline (without block-level tags).


SaptakS (approved):
Tested the code locally. Works as expected! Looks good to me.

harrislapiroff (commented):
Code looks good—I'll check it on staging

PR #1054, opened 2021-05-12
Update to version of django-logging with request duration support

chigby:
Refs https://github.com/freedomofpress/fpf-www-projects/issues/174

This pull request enables the latest version of django-logging, which includes the duration field in the request logs.


harrislapiroff (approved):
LGTM

PR #1057, opened 2021-05-17
Disable logging of response content

chigby:
This pull request removes the "content" of a response from the log objects created by `django-logging`. In some ways it's a re-do of #927, which I think was lost when we standardized some of the logging features across sites. 

I've updated two settings. The first one tells django-logging not to try and parse JSON response data at all, this is wasting time since we don't want to log that parsed data in the first place.

The second one removes the "content" field from the set of response fields on the logging object. I've set it to the default fields but without `content`.


harrislapiroff (approved):
Did not test locally, but code change looks right

PR #1071, opened 2021-05-26
Adds observableusercontent in connect_src CSP

### LINKED ISSUE #1070: CSP additions
In testing on staging I discovered the following necessary additions to our CSP:

* `connect-src`
   * https://static.observableusercontent.com/
* `img-src` (and probably other media policies)
   * https://staging-media.pressfreedomtracker.us/ (staging only, probably worth checking how we accomplished this on other sites)

The Observable requirement seems to be part of the new runtime. To test locally, include this embed code in a raw html block on a local page:

```html
<div id="embedded-viz">
  <div class="embedded-viz-chart"></div>
</div>

<script type="module">
  const vizName = 'map'
  const filters = {
    tags: 145
  }

  import {Runtime, Inspector} from "https://cdn.jsdelivr.net/npm/@observablehq/runtime@4/dist/runtime.js";
  import define from "https://api.observablehq.com/@harrislapiroff/us-press-freedom-tracker-visualizations.js?v=3"
  const main = (new Runtime).module(define, name => {
    if (name === vizName) return Inspector.into("#embedded-viz .embedded-viz-chart")()
  });
  main.redefine('filters', filters)
</script>
```

Twitter embeds seem to work fine!

SaptakS:
I have removed 2 rules because they didn't make sense on a second look for tracker.
For adding the media links in the CSP, like other repositories, an environment variable `DJANGO_CSP_MEDIA_ORIGINS` needs to be set in the deployment environment.

Fixes #1070

### DIFF
--- a/tracker/settings/base.py
+++ b/tracker/settings/base.py
@@ -289,12 +289,11 @@ CSP_CONNECT_SRC = [
     "https://releases.wagtail.io/latest.txt",
     # Observable Notebooks
     "https://cdn.jsdelivr.net",
-    "https://pressfreedomtracker.us",
+    "https://static.observableusercontent.com/",
 ]
 CSP_IMG_SRC = [
     "'self'",
     "https://analytics.freedom.press",
-    "https://media.pressfreedomtracker.us",
     # For Twitter Widgets
     "https://platform.twitter.com",
     "https://syndication.twitter.com",


harrislapiroff (approved):
Looks good. Will test on staging.

PR #1072, opened 2021-06-01
Topic page date ranges

### LINKED ISSUE #1066: Allow topic pages to filter incidents by a date range
We would like to add a non-required upper and lower date range to topic pages. These values will filter what is shown on that page to only incidents within the range. This filtering is in addition to the normal tag-based selection of incidents for topic pages.

chigby:
Fixes #1066 

This pull request adds two fields to `TopicPage`: a start and end date, both optional, but if either one are configured, then the incidents shown on the topic page will be constrained to be within that range (inclusive). As part of this I've added tests and a simple validator for ensuring that the start date precedes the end date.

There is one open question which is how this date filter affects incidents with `exact_date_unknown == True`. The incident search feature has one way of handling this but I wasn't sure if we wanted to copy that over.


harrislapiroff:
My immediate inclination is to count them by whatever date is specified (it's a little weird and maybe we should add a note to the bool about that) but what does the incident search do?

chigby:
The incident search treats the specified date as the entire month, and any date range that overlaps the month will include that incident. I'm slightly concerned this implementation might impact the performance of the already slightly beleaguered topic page endpoint, but it's something I can test. 

The "treat it as whatever date is specified" is what is happening as this PR stands right now.

chigby:
e.g. if your incident has `exact_date_unknown = True, date = '2015-04-13'` then it is included in any search whose date range overlaps April 2015.

harrislapiroff:
Yeah, ok, that makes conceptual sense to me. Test performance and update with results. I do think it would make sense to match behavior if possible since editors may verify date ranges using the site filtering.

chigby:
All right, this should be ready to review again. The benchmarking I did on the fuzzy date logic is fine, no change in performance as far as I could tell. This is because I managed to implement it in a way where all we add is a single where clause (I was afraid this might not be possible).

SaptakS (commented):
This PR will need some merge conflict resolutions, and also need to update the migration since now there is a migration conflict.

SaptakS (commented):
This looks good to me code wise and implementation mostly, but @harrislapiroff might need to take a look in case this meets the needs for the data viz as well

harrislapiroff (approved):
This looks good to me

PR #1074, opened 2021-06-02
Use an explicit random seed in the test suite

### LINKED ISSUE #1060: Ephemeral test failures?
See:

https://app.circleci.com/pipelines/github/freedomofpress/pressfreedomtracker.us/1063/workflows/b895d848-fd24-4a83-97c9-b1b70f19d373/jobs/7788

https://app.circleci.com/pipelines/github/freedomofpress/pressfreedomtracker.us/1063/workflows/daefcd95-cce9-412c-9a16-8b7dcfa2f3cd/jobs/7783

chigby:
Adds a test runner whose purpose is to configure the seed for Factory Boy (as described [in the documentation](https://factoryboy.readthedocs.io/en/stable/recipes.html#recipe-random-management)) to prevent randomly generated objects from causing unexpected and ephemeral test failures.

Such failures are particularly common for our incident-related inline objects that consist solely of a single field that must be unique.  When randomly generating several of these for a test, if the same value is chosen for that field for more than one object (an uncommon but definitely possible occurrence) then only a single object will be created where the test expected two, thus potentially causing a test assertion against an aggregate count to fail.

My hope is that using the same seed locally and on CI will prevent us from being surprised when such problems occur.

I've placed the seed in the base settings file, which makes it consistent across development and CI environments but also allows for fairly easy customization locally should the situation call for it. I chose the number with `random.randint(0, 1_000_000_000)`.

Fixes #1060

### DIFF
--- a/common/test_runner.py
+++ b/common/test_runner.py
@@ -1,0 +1,27 @@
+from xmlrunner.extra.djangotestrunner import XMLTestRunner
+import factory.random
+
+from django.conf import settings
+from django.test.runner import DiscoverRunner
+
+
+class WithSeedMixin(object):
+    """Modifies a test runner class to use a given constant from
+    `settings.RANDOM_SEED` to seed the generation of randomized
+    factory objects.
+
+    """
+    def setup_test_environment(self):
+        seed = settings.RANDOM_SEED
+        if seed:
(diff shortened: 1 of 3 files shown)


SaptakS (approved):
Looks good to me functionally.

PR #1085, opened 2021-06-14
Adds logging for purge_page_from_cache

### LINKED ISSUE #1068: Add logging of frontend cache invalidation
We've got a pretty complex set of signals that trigger frontend (cloudflare) cache invalidations. In order to make sure that's actually happening as expected, let's add logging to those invalidation functions. We should be sure, at a minimum, to log:

- identifying information about the model triggering the invalidation (if it's a page, maybe the path to the page, if it's a settings object, maybe the class name of the object, if it's anything else, class name and id)
- the path and/or tags being invalidated

SaptakS:
Logging the page class name along with the page title and page
slug to get a better idea of which page of that particular class
is getting  purged

Resolves #1068

### DIFF
--- a/blog/signals.py
+++ b/blog/signals.py
@@ -1,9 +1,14 @@
+import logging
+
 from django.db.models.signals import post_delete
 from wagtail.core.signals import page_published
 from wagtail.contrib.frontend_cache.utils import purge_page_from_cache
 
 from blog.models import BlogIndexPage, BlogPage
 from cloudflare.utils import purge_tags_from_cache
+
+
+logger = logging.getLogger("wagtail.frontendcache")
 
 
 def purge_blog_index_page_frontend_cache(**kwargs):
@@ -18,6 +23,9 @@ def purge_blog_index_page_frontend_cache(**kwargs):
     for page in BlogIndexPage.objects.all():
         tags.append(page.get_cache_tag())
         purge_page_from_cache(page)
+        logger.info(
+            f"Purged page BlogIndexPage with title: {page.title} and slug: {page.slug}"
+        )
     purge_tags_from_cache(tags)
 
 
--- a/common/signals.py
+++ b/common/signals.py
@@ -1,3 +1,5 @@
+import logging
+
 from django.db.models.signals import post_delete, post_save, pre_delete
 from django.dispatch import receiver
 from wagtail.core.signals import page_published
@@ -11,6 +13,9 @@ from home.models import HomePage
 from blog.models import BlogPage
 
 
+logger = logging.getLogger("wagtail.frontendcache")
+
+
 def purge_category_from_frontend_cache(**kwargs):
     """
     Currently this busts all the category pages' caches. If we wanted to be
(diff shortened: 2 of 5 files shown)


chigby:
Were you able to see the logging messages in the console locally with this branch?

harrislapiroff:
@chigby or @SaptakS Can you write up what would be required to change to get these logs to show up in console? Is there a way to allowlist certain `info` level logs, but not others?

@maeve-fpf Can you confirm that only logs which show up in console in dev will get piped into Kibana?

chigby:
To get the logs to appear in my local dev console, I changed this line from `True` to `False`:

https://github.com/freedomofpress/pressfreedomtracker.us/blob/c4cc0c51c214b72bbb37e2d86559966abcc5954b/tracker/settings/base.py#L353

chigby:
Also I think specific loggers can be configured by name (i.e. by the string you pass to `logging.getLogger`) lower down in the settings file in the `LOGGING` structure. You can control what handler a specific logger is sent to, and filter by level. I have not really investigated it that much, but I could imagine a setup that changed the catchall logger to not log anything, and then allow others on a case by case basis.

chigby:
So it seems like disable_existing_loggers doesn't do what I had assumed it did. 

From the [Django docs](https://docs.djangoproject.com/en/2.2/topics/logging/#configuring-logging) (emphasis mine):

> By default, the LOGGING setting is merged with [Django’s default logging configuration](https://docs.djangoproject.com/en/2.2/topics/logging/#default-logging-configuration) using the following scheme.

> If the `disable_existing_loggers` key in the `LOGGING` dictConfig is set to `True` then all loggers **from the default configuration** will be disabled. Disabled loggers are not the same as removed; the logger will still exist, but will silently discard anything logged to it, not even propagating entries to a parent logger. **Thus you should be very careful using `'disable_existing_loggers': True`; it’s probably not what you want**. Instead, you can set `disable_existing_loggers` to `False` and redefine some or all of the default loggers

So it doesn't disable all loggers, just Django's default loggers, meaning I'm not sure it was the cause of not seeing the logs.

I was able to get the messages on this branch to appear on the console, without changing `disable_existing_loggers`, by adding this key/value pair to the `LOGGING["loggers"]` dictionary:

```
"wagtail.frontendcache": {"handlers": ["normal"], "propagate": False, "level": "INFO"}
```

maeve-fpf:
@harrislapiroff Yes, only logs that show up on the console - it would be best to try the "prod-like" settings if those exist, though, in case there's a difference from dev.

@chigby That looks like the config we want! Thanks.

This will still log a line of plain text; in the future we may want to think about adding some structure.

chigby:
The logger produces messages that look like this:

```
{"message": "Purged page HomePage with title: Home and slug: home"}
```

Which is somewhat bare bones. This is due to our use of [python-json-logger](https://pypi.org/project/python-json-logger/), which seems like it removes (or never adds) any other information. Note that it's possible to call the log function with an `extra` argument and supply a dictionary, which gets added to the JSON. For example:

```python
logger.info(
    f"Purged page BlogIndexPage",
    extra={'title': page.title, 'slug': page.slug},
)
```
yields
```
{"message": "Purged page BlogIndexPage", "title": "PFT Blog", "slug": "pft-blog"}
```

SaptakS:
Should I update the config in the base.py file as mentioned by @chigby and should that reflect in staging then? Also yes, I agree that right now it's just a plain text and can add more structure the way @chigby mentioned. Right now, there's not much information related, hence I didn't bother much around the structure.

chigby:
@SaptakS Yes, I think what you describe is the best next step here.

chigby (approved):
Messages are now definitely showing up in the console for me, should be ready to see how this is looking on the live site.

PR #1089, opened 2021-06-16
Update CSP to allow functioning ReCAPTCHA

chigby:
With these changes the captcha renders for me locally, albeit with the "testing purposes only" warning.


SaptakS (approved):
Looks good to me.

PR #1091, opened 2021-06-17
Fix a bug in generating unique dates

harrislapiroff:
Previously we were generating unique dates for incidents by concatenating
incident date with incident database id on save. Unfortunately, because
when an incident is first saved it has no database id, this was
generating unique date keys that looked like "2020-06-16-None" and were
not guaranteed to be unique.

The test was missing this error because using the IncidentPageFactory
triggered multiple calls to .save() and the key was corrected on the
second call--thus the test passed.

I've rewritten the functionality to use a generated-on-the-fly uuid
instead of database id and the test to compare two incidents against
each other to verify their unique dates are not identical. I also
rewrote the test to use IncidentPage directly instead of the factory.

Suggested Review Process
---------------------------

1. Read the above, make sure it makes sense
2. Read the code, make sure it does what I said above
3. Bonus step: run the application locally and ensure you can create two different incidents that happened on the same date (We'll also have an opportunity to test this on staging once merged)

### DIFF
--- a/incident/models/incident_page.py
+++ b/incident/models/incident_page.py
@@ -1,4 +1,5 @@
 import datetime
+import uuid
 from urllib.parse import urlencode
 
 from django import forms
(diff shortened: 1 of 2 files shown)


SaptakS (approved):
Looks good to me logically.

PR #1095, opened 2021-06-23
Paginates RSS feed for blogs and incidents

### LINKED ISSUE #1058: RSS feed is massive and unpaginated
https://pressfreedomtracker.us/all-incidents/feed/ regularly clocks in as a slow page request at >20 seconds, and that should not be surprising -- it's a 5.3M file that appears to include all incidents going back to January 2018.

It's standard practice to truncate feeds after a limited number of items (e.g., 15). We could use `<link rel="next">` to paginate the feed.

SaptakS:
Resolves #1058 

I have added pagination for both blogs and incidents. The number of items per feed is customizable from the admin interface, with a default of 20. All the paginated feeds have the following rel links to help with pagination:
<link href="http://pressfreedomtracker.us/all-incidents/feed/?p=1" rel="first"/>
<link href="http://pressfreedomtracker.us/all-incidents/feed/?p=16" rel="last"/>
<link href="http://pressfreedomtracker.us/all-incidents/feed/?p=1" rel="previous"/>
<link href="http://pressfreedomtracker.us/all-incidents/feed/?p=3" rel="next"/>

And similar with corresponding links for the blog.


blog/feeds.py
```diff
@@ -48,13 +62,22 @@ def feed_url(self):
...
-        if self.blog_index_page.feed_limit != 0:
-            return posts[:self.blog_index_page.feed_limit]
+        starting_index = ((self.page - 1) * self.feed_per_page)
+        last_index = (self.page * self.feed_per_page)
+
+        if self.blog_index_page.feed_limit != 0 and self.blog_index_page.feed_limit < last_index:
+            return posts[starting_index:self.blog_index_page.feed_limit]
 
-        return posts
+        return posts[starting_index:last_index]
```
chigby: I'm looking at this code and the code in the `get_object` function above and I'm wondering if we're not re-inventing the wheel a little. What would you say about taking advantage of [Django's built-in pagination features](https://docs.djangoproject.com/en/2.2/topics/pagination/)? It seems to me like a lot of similar functionality is exposed to what is going on here, meaning we might be able to get away with fewer manual calculations.

chigby (changes requested):
I think we can get away with more use of built-in Django functionality here. Let me know if you're concerned about that approach, though, as I might be missing something. See comment below for more information, and ask if anything is not clear.

blog/feeds.py
SaptakS: Yes. I could have used the in-built paginator check there. But I also had to check for the feed_limit logic, so I had to calculate anyways to check for that, so I just used the calculated results. I can definitely still use the in-built paginator. Let me know what you think.

chigby: I was thinking that Django's paginator could be applied after the feed limit is applied, something like

```python
from django.core.paginator import Paginator

incidents = incident_index_page.get_incidents()[:feed_limit]  # plus the check for feed_limit == 0, and so on...
paginator = Paginator(incidents, incident_index_page.feed_per_page)
```

And then the paginator gives you a lot of what we want here in terms of page calculations. But I haven't actually tried it so there might be something I'm not considering. Even though we're calculating the overall limit, I think having Django handle the page indices and related is still a fairly big gain.

chigby (approved):
Looks great to me, I think this turned out quite nicely. Thanks for making the Paginator adjustments!

PR #1097, opened 2021-06-29
Image management improvements

### LINKED ISSUE #985: Better image management
Kirstin has been using the "Title" field on Wagtail images to add searchable information about the images themselves. This isn't ideal since the title field also serves extra duty as the accessible `alt` attribute for images. This ticket covers a few issues:

- [ ] Add a note to the title tag similar to what we have on freedom.press articulating that the field is used for accessibility
- [x] Talk to Kirstin about what info she needs to track about images
- [x] Figure out to what extent that info can be managed using the already built in collections functionality
- [ ] Add a `slug` field that Kirstin and co can use to stash the filename that they use to track the image through the editorial process. This does not need to be a `SlugField`—we can allow spaces and symbols, so just a `CharField` will do.

chigby:
Fixes #985

This PR adds two features to the image admin UI:

1. A "slug" field for storing some additional text on the image.
2. A bit of help text around the title field to encourage thinking about its use as alt-text. [Implementation taken from freedom.press](https://github.com/freedomofpress/freedom.press/commit/65dae40ca9c916f88c55be8694090228bade7744)


SaptakS (commented):
I can't seem to find the help text in the form:

image

Wonder if it is an issue in just my system and I need to rebuild docker 🤔

SaptakS (changes requested):
Seems like the help text for the title field is there when we go for editing an image, but when we try to add a new image from the link in admin home (which leads to `/admin/images/add`), the title field doesn't show any help text right now. It works while adding from Images -> Add an Image

SaptakS (approved):
Looks good to me.

PR #1102, opened 2021-07-14
Makes city filter case insensitive

### LINKED ISSUE #1063: Serching by city shouldn't be case sensitive

SaptakS:
Fixes #1063


chigby (changes requested):
Do you think it would be possible to add a test for this behavior? I think it would probably be something like this: 

https://github.com/freedomofpress/pressfreedomtracker.us/blob/develop/incident/tests/test_filtering.py#L79-L93

SaptakS:
@chigby ready for a second review once the CI passes.

chigby (approved):

PR #1104, opened 2021-07-21
Adds option to add category specific SEO image for incidents

### LINKED ISSUE #1080: Per-category default SEO images
We currently use a default image for all pages if they don't have an SEO, but our current default image is pretty boring

image

We've found that posts with real images get more engagement, but we don't have the time to search down interesting images for all incidents, nor do all incidents have interesting images. To add a little variety, let's allow content editors to set a default SEO image on a per-category basis—it will become the default SEO image for incidents in that category with no other image set.

SaptakS:
Resolves #1080 

Adds an optional SEO image field in category pages. If the incident doesn't have a teaser_image, it uses the main category's SEO image if any, else falls back to the usual site meta image.


incident/models/incident_page.py
```diff
@@ -798,7 +798,11 @@ def get_court_circuit(self):
         return None
 
     def get_meta_image(self):
-        return self.teaser_image or super(IncidentPage, self).get_meta_image()
+        return (
+            self.teaser_image or
+            self.get_main_category().default_image or
+            super(IncidentPage, self).get_meta_image()
```
chigby: We might not be able to get away with calling `super().get_meta_image` here, since what that function does is either return the incident's `search_image` field _or_ the site-wide default image. So if we only call this after we check for the category default image, the incident-level search image will not be shown in any categories with a default image.

Though it's possible we are not using the `search_image` field, in which case maybe it does not matter so much.

incident/models/incident_page.py
```diff
@@ -798,7 +798,11 @@ def get_court_circuit(self):
         return None
 
     def get_meta_image(self):
-        return self.teaser_image or super(IncidentPage, self).get_meta_image()
+        return (
+            self.teaser_image or
+            self.get_main_category().default_image or
```
chigby: I worry slightly here about `get_main_category()` potentially returning `None`

SaptakS:
@chigby I have tried to address both the concerns in the function so we should be mostly safe. Let me know if that makes sense.

chigby (approved):
Love the changes, this looks great!

PR #1105, opened 2021-07-21
Geocoding latitude and longitude for incidents

### LINKED ISSUE #1062: [API] Add longitude and latitude fields to tracker
This would be useful for anyone trying to make map-based visualizations.

These fields can be purely derived from city/state. We should specify in the docs at the same time that they represent approximate, not precise, locations.

Not sure what geocoding backend makes sense to use, so research required there. If we need to rely on raw data, I've created a CSV of cities with lat/long/pop here: https://observablehq.com/d/808d887451a25d6a

chigby:
Fixes #1062

This pull request:

1. Adds latitude and longitude fields to incident pages and exposes them through the API.
2. Adds a "geonames" app that consists mostly of a set of models for city name data obtained from http://geonames.org ([export page](http://download.geonames.org/export/dump/)) containing the names of countries, regions (state-level areas), districts (county-level areas) and cities with populations over 5,000.

The geonames app has:
1. Fixtures in JSON format containing the data. I did some editing of the raw data downloaded from geonames.org site for consistency and to adapt it into our database's format. The fixtures will allow us to load it directly onto the production site.
2. A `get_city_coords` function that takes a city and a state (and, optionally, a country code) and returns latitude and longitude values if that name exists in the database. I've hooked this function up to the `save` method of `IncidentPage` so that any time a new page is added or an old one updated, the longitude and latitude are kept current.
3. A management command to update all existing incident pages with the correct coordinates for the city and state defined on that page.

## Deployment
Once this code is deployed, the only thing that will immediately change is the availability of the lat/long fields in the API, but they will not be populated yet. To populate them, we must first load the fixture data with:

```
manage.py loaddata --app geonames cities5000.json
```
After this data is loaded, any incident page that is saved will have its lat/long fields populated (if the city can be found). There is also the command:

```
manage.py populate_geodata
```

That will update all incidents at once. Ideally, this would be run once the fixtures are loaded.


harrislapiroff:
@SaptakS Could you take a swing at reviewing this? (@chigby is working on coverage by excluding apps.py)

chigby:
Fixed the coverage situation

geonames/cities.py
```diff
@@ -0,0 +1,31 @@
...
+    city_coords = (None, None)
+
+    if not city or not state:
+        return city_coords
+
+    latlong = GeoName.objects.filter(
+        name=city,
+        regcode=state,
+        isocode__iso3=iso3
+    ).values_list('latitude', 'longitude')
```
SaptakS, line 26: I wonder if case insensitive filtering makes more sense with city names since the city names in incident are just text fields, and the cases might differ from the json? 🤔

geonames/models.py
```diff
@@ -0,0 +1,51 @@
...
+        constraints = [
+            models.UniqueConstraint(fields=['isocode', 'regcode'], name='unique_country_and_region')
+        ]
+
+
+class District(models.Model):
+    name = models.TextField(null=True)
+    geonameid = models.BigIntegerField(primary_key=True)
+
+    isocode = models.IntegerField()
```
SaptakS: Shouldn't this be a foreign key to country like the other fields? or it's different from District? Also I see that District is not being used anywhere right now so I am guessing we are saving the information just in case we have a future use?

SaptakS (commented):
Looks good to me. I am guessing the primary use of the data is mainly in APIs, so the information not being shown in incident page makes sense. Left a few comments but it is not blocking.

Also, I will note that lat,long are saved only  if both state and city names are present for incidents.

geonames/models.py
chigby: Good point about the foreign key! I think it did just get left out because we aren't using District right now. 

I do think we should do something with that data eventually, though. It adds another layer of what is possible for data analysis and visualization.

geonames/cities.py
chigby: This is a good idea, too. Not everyone uses the same case for the city names. There might be some other differences but at least this one is easy to account for.

The data set I downloaded actually does include a set of "alternate names" for each GeoName, but it seemed like it might be a fair amount of work to incorporate those into the lookup for not all that much gain. But I could be wrong about that!

SaptakS: Yeah I was thinking as well that there might be other differences, but case seems easiest to address.

SaptakS (approved):
Looks good to me. Merging

PR #1108, opened 2021-07-26
Reinstate by-default pagination for incidents API, JSON format

### LINKED ISSUE #1065: Incident API: Limit number of results by default
Right now, a request to the root of the incident API returns every incident in the database. We would like it to paginate that list by default for the JSON view, with a reasonable limit in place. I'd suggest 25. The CSV view should continue to not paginate.

See aa4950d6fe689d9c4b707ae2e25fd141e9c2774d

chigby:
Was removed in aa4950d6fe689d9c4b707ae2e25fd141e9c2774d

Does not apply pagination to the CSV format

Fixes #1065


SaptakS (approved):
Looks good to me.

PR #1110, opened 2021-08-04
Rate limit email signups

chigby:
This pull request adds rate limiting to the email signups creation endpoint. Currently 20 per hour per user, but this can be easily changed.

The way this is implemented is using Django Rest Framework's (DRF) throttling system, which I think is fairly well-suited to this problem since:
1. The thing we want to throttle is basically already a REST endpoint.
2. It is a lot less code and overhead than copying over the rate-limiting system from freedom.press

So I modified the signup view slightly to make it DRF compatible without changing its functionality, then applied throttling to it.


harrislapiroff (approved):
Interesting approach to the rate limiting. I like it! Let's get it on staging.

Issue #1101, opened 2021-07-08
Image issue


harrislapiroff:
It looks like an image file overwrote an older image file with the same name. I'm surprised this could happen. We should see if we can reproduce it and if so either implement a fix or file a ticket upstream.

It's also possible the older image had a different name (@KAMcCudden says this is not consistent with how she names files) and somehow the filename was overwritten in the database, which might be harder to track down as a bug. More likely it just wasn't Kirstin who uploaded the files.

Older image:
![image](https://user-images.githubusercontent.com/296631/124980153-08600900-e002-11eb-975c-4c03797b6150.png)

Newer image:
![image](https://user-images.githubusercontent.com/296631/124980178-0e55ea00-e002-11eb-8389-00c1014a4406.png)

harrislapiroff:
These may be relevant:

https://github.com/wagtail/wagtail/issues/1686
https://github.com/wagtail/wagtail/issues/3883

If that's the issue, then:

1. You'll need to test on staging to reproduce and,
2. setting
   ```
   AWS_S3_FILE_OVERWRITE = False
   ```
   should fix it.

SaptakS:
I wonder why that's not the default behaviour and why overwrite is the default behaviour. Also I wonder if we should add a validation on application side for image filenames as well in a full_clean() or clean() method or does that get ignored as well without the settings? 🤔

harrislapiroff:
I think just the S3 setting for now. It does seem like this should be the default behavior, though I do understand why this might happen if the storage backend is responsible for creating filenames... it doesn't seem either like 1. the storage backend should default to no overwrite or 2. wagtail should place additional restrictions on filenames beyond the backend. But the resulting combination is bad and maybe is at least worth a warning on the docs.

chigby:
I verified that I can overwrite images on staging by uploading a new one with the same filename. Though I think it does not necessarily overwrite all existing renditions immediately as those file names can be different, depending on when/why they were generated, but the "original" image was definitely replaced by the more recently uploaded one.

I made a PR for updating the setting mentioned in the thread. If it succeeds in fixing the issue, we'll probably want to add this to our other sites also.

chigby:
The fix for this is merged on all web repos:

https://github.com/freedomofpress/securedrop.org/pull/882
https://github.com/freedomofpress/freedom.press/pull/972
https://github.com/freedomofpress/securethenews/pull/363
[...]

PR #1111, opened 2021-08-09
Disable Google Storage default file overwrite behavior

chigby:
Refs #1101 

This pull request will disable the default `django-storages` behavior for overwriting files unless the environment variable
`GS_FILE_OVERWRITE` is set to the string `"True"`.

See documentation here for more info on the GS_FILE_OVERWRITE option:

https://django-storages.readthedocs.io/en/latest/backends/gcloud.html

## Testing and Acceptance
Once this change is merged, we will definitely want to verify the correct behavior (as described in the issue) on staging.


SaptakS (approved):
LGTM

PR #1099, opened 2021-07-07
Put unique date generation in `full_clean`

chigby:
This PR moves the `unique_date` generation to from `save()` to `full_clean()`. It also slightly changes how we make populate the field so that the UUID is only generated once per incident.

The move to `full_clean` was done in response to some errors we noticed in production, and I was able to cause an error locally when attempting to preview an incident that had not yet been saved. The changes on this PR were able to resolve the local error for me.

Error information from production:

```
{'type': 'django.core.exceptions.ValidationError',
 'message': "{'unique_date': ['This field cannot be blank.']}",
 'traceback': [{'file': '/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py',
 'method': '_get_response',
 'line': 113},
 {'file': '/usr/local/lib/python3.9/site-packages/django/views/decorators/cache.py',
 'method': '_wrapped_view_func',
 'line': 44},
 {'file': '/usr/local/lib/python3.9/site-packages/wagtail/admin/urls/__init__.py',
 'method': 'wrapper',
 'line': 127},
 {'file': '/usr/local/lib/python3.9/site-packages/wagtail/admin/auth.py',
 'method': 'decorated_view',
 'line': 172},
 {'file': '/usr/local/lib/python3.9/site-packages/django/views/generic/base.py',
 'method': 'view',
 'line': 71},
 {'file': '/usr/local/lib/python3.9/site-packages/wagtail/admin/views/pages/edit.py',
 'method': 'dispatch',
 'line': 131},
 {'file': '/usr/local/lib/python3.9/site-packages/django/views/generic/base.py',
 'method': 'dispatch',
 'line': 97},
 {'file': '/usr/local/lib/python3.9/site-packages/wagtail/admin/views/pages/edit.py',
 'method': 'post',
 'line': 218},
 {'file': '/usr/local/lib/python3.9/site-packages/wagtail/admin/views/pages/edit.py',
 'method': 'form_valid',
 'line': 238},
 {'file': '/usr/local/lib/python3.9/site-packages/wagtail/admin/views/pages/edit.py',
 'method': 'publish_action',
 'line': 273},
 {'file': '/usr/local/lib/python3.9/site-packages/wagtail/core/models.py',
 'method': 'save_revision',
 'line': 1318},
 {'file': '/usr/local/lib/python3.9/site-packages/wagtail/core/models.py',
 'method': 'full_clean',
 'line': 976},
 {'file': '/usr/local/lib/python3.9/site-packages/django/db/models/base.py',
[...]

PR #1119, opened 2021-08-18
API pagination headers

### LINKED ISSUE #1093: [API] Add pagination headers
Let's use the `Link:` response header to add pagination info (see an example in [GitHub's API docs](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#pagination). This is a semi-standardized best practice, I believe, and it will enable pagination on CSV, where we are unable to provide an "envelope" in the content of the response.

Note:
* We can also add links for `"first"` and `"last"` in addition to `"next"` and `"previous"`
* We could consider adding custom data to headers, e.g., `X-Total-Count`
* We should make a decision as to whether to leave the envelope with pagination info in the JSON response as a redundant measure (particularly for developers who are not experienced with or don't have access to HTTP headers) or remove it for consistency

chigby:
Fixes #1093 

This pull request:

1. Makes optional the Django rest framework "pagination envelope" (where the top-level object returns contains next/previous links along side a "results" key that contains the query results). The envelope can be requested by adding the query parameter `envelope=1` to the request. I've done this because I do think having this capability could be useful for some people writing applications with this API, but that the header-based pagination is a better default.
2. Adds next, previous, and first page links to the `Link` header (describe in the issue).
3. Does _not_ add the "last page" link, as I don't believe this is possible with cursor based pagination. As a side note, if it is seeming like cursor based pagination is more trouble than it's worth (see also #1091, #1099) I'd be open to switching to limit/offset pagination.
4. Makes some minor changes to our factory classes that we use in tests to make them less random. See commit message for explanation but the short story is that using random seeds still was causing problems for me locally.

### DIFF
--- a/common/devdata.py
+++ b/common/devdata.py
@@ -1,5 +1,6 @@
 import factory
 import wagtail_factories
+from faker import Faker
 from wagtail.core import blocks
 from wagtail.core.rich_text import RichText
 
@@ -22,6 +23,9 @@ from common.models import (
     TaxonomyCategoryPage,
     TaxonomySettings,
 )
+
+
+fake = Faker()
 
 
 class DevelopmentSiteFactory(wagtail_factories.SiteFactory):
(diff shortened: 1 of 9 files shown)


SaptakS (approved):
Looks good to me. I personally liked the idea of cursor pagination so that the position of items don't move away and stay in the same cursored link.

PR #1123, opened 2021-08-24
Request middleware tests

chigby:
This pull request:

1. Adds tests for the `RequestLogMiddleware`. This required some small changes to the settings file to work properly.
2. Updated the version of `coverage` to the latest
3. Moved the coverage configuration into a `.coveragerc` file.
4. As an experiment, adds a circle ci job to compare the coverage from the PR branch to the develop branch. I'm kind of checking if this is possible and what it would look like, and if it needs to be done in a separate PR (or completely reworked if we don't like it) then that's fine.


SaptakS (approved):
Looks good to me. Really like the diff-cover check. Should at least ensure that all PRs have necessary tests to some extent.

Issue #1061, opened 2021-05-24
[API] Allow query by relation names and/or provide a lookup for object IDs


harrislapiroff:
Problem:

Presently the API accepts filter parameters as object IDs, e.g.,

```
https://pressfreedomtracker.us/api/edge/incidents/?city=San+Francisco&state=8
```

Unfortunately there's no way for an API consumer to figure out that `8` is the ID for California besides manually messing around with the purple search box on the website. This is also true of other m2m/foreignkey fields, such as:

* Journalists
* Institutions
* Tags
* etc.

Three possible solutions that we should do some combination of:

1. Allow queries by name, e.g.,
  ```
   https://pressfreedomtracker.us/api/edge/incidents/?city=San+Francisco&state=CA
   ```
2. Provide endpoints for looking up details of related objects
3. Embed ID information in the incident response (i.e., instead of just rendering the names of objects, add a nested dict to include IDs as well... this might require a fussier solution for CSV)

Feel free to spec out what solution (or combination thereof) you think is best and comment on this ticket with the proposed approach before working on it.

PR #1130, opened 2021-09-15
Filter relations by humanized name

chigby:
Refs #1061

This pull request updates the incident filter code to permit filtering by text on related items. For example, we can now call:

```python
IncidentFilter({'state': 'NM'}).get_queryset()
```
to retrieve all incidents that were in the state with abbreviation "NM," where previously we would have to know that this state has ID 36 and call 

```python
IncidentFilter({'state': '36'}).get_queryset()
```

The goal of this is to improve the human readability and constructability of API query strings. The GET parameters of URLs like `http://localhost:8000/api/edge/incidents/?state=36` are passed directly to `IncidentFilter`, so this change allows us to write URLs like this instead: `http://localhost:8000/api/edge/incidents/?state=NM`. The result will be the same.

I don't have an exhaustive list of exactly which fields are affected by this change, but here are the general rules.

1. Any field that's directly related to another object with a `title` field can be filtered by that title field. Example: `http://localhost:8000/api/edge/incidents/?arresting_authority=Keystone%20Cops`
2. This includes many-to-many relationships. Example: `http://localhost:8000/api/edge/incidents/?venue=Sea`
3. State, which does not have a `title` field, can be filtered by `name` OR `abbrevation`. Meaning: `http://localhost:8000/api/edge/incidents/?state=NM` and ``http://localhost:8000/api/edge/incidents/?state=New%20Mexico` are equivalent.
4. Categories can be filtered by title. Example: `http://localhost:8000/api/edge/incidents/?categories=Equipment Damage`. This is mostly the same as point 1, but categories do not have a "direct" relationship to incident pages and there is some additional technical baggage that goes along with this change. See commits.
5. Targeted institutions and charges can be filtered by title. These are also "indirect" relationships.
6. For filtering by multiple values, separate those values with commas. This parallels how filtering by ID works. Example: `http://localhost:8000/api/edge/incidents/?categories=Leak%20Case,Border%20Stop` should return the same as `http://localhost:8000/api/edge/incidents/?categories=9,4`. Note that this does, at present, make titles or other text queries containing commas problematic. I'm not sure exactly the best way to resolve this, though I think it does not apply extremely often.
7. Furthermore, though I'm not sure what the use-case is for this exactly, ID values and names can be combined. Example: `http://localhost:8000/api/edge/incidents/?categories=9,Border%20Stop`

I've added some tests to ensure that the new functionality works, and slightly updated a few where internal changes required it. I'm pretty confident that the filtering by IDs works exactly how it did before.

I'm not saying that this fully resolves the original issue, since there is also the suggestion of adding endpoints for some of these related items in addition to making the filtering easier.

### DIFF
--- a/incident/models/incident_index_page.py
+++ b/incident/models/incident_index_page.py
@@ -176,12 +176,14 @@ class IncidentIndexPage(RoutablePageMixin, MetadataPageMixin, Page):
             context['export_path'] = self.url + self.reverse_subpage('export_view')
 
         incident_filter.clean()
-        category_ids = incident_filter.cleaned_data.get('categories')
+        category_data = incident_filter.cleaned_data.get('categories')
 
-        if not category_ids:
+        if not category_data:
             context['categories'] = CategoryPage.objects.live()
         else:
-            context['categories'] = CategoryPage.objects.filter(live=True, pk__in=category_ids)
+            context['categories'] = CategoryPage.objects.live().filter(
+                models.Q(pk__in=category_data.pks) | models.Q(title__in=category_data.strings)
+            )
 
         incident_qs = incident_filter.get_queryset() \
             .select_related('teaser_image', 'state', 'arresting_authority') \
--- a/incident/tests/test_filtering.py
+++ b/incident/tests/test_filtering.py
@@ -26,7 +26,8 @@ from incident.tests.factories import (
     StateFactory,
     InstitutionFactory,
     JournalistFactory,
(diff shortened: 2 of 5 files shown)


SaptakS (approved):
Looks good to me

harrislapiroff:
Excited to get this feature in! I agree that we should still also provide endpoints for related items. (A `categories` endpoint would be particularly helpful, but I imagine the other ones will probably be handy as well.)

PR #1131, opened 2021-09-20
Update jest version to 27

chigby:
This PR updates the versions of the packages jest and babel-jest to the latest in the 27.x line. The motivation for doing this is that it removes the dependency on `set-value` which has a [vulnerability on version 3.x](https://github.com/advisories/GHSA-4jqc-8m5r-9rpr).

In the update process, I've I needed to change the jest configuration so that the tests pass.  This means setting the `testEnvironment` configuration option to `jsdom` (default is `node`).  One side effect of this, I discovered, was that `window.setImmediate` was removed.  I think `process.nextTick` is an appropriate substitute for what we're using it for here, even if it's not identical.


SaptakS (approved):
Looks good to me!

PR #1134, opened 2021-09-29
Upgrade version script, add /health/version

maeve-fpf:
Another PR for https://github.com/freedomofpress/fpf-www-projects/issues/220, just like the previous two. Will crosslink there.

### DIFF
--- a/common/tests/test_views.py
+++ b/common/tests/test_views.py
@@ -11,6 +11,7 @@ from incident.tests.factories import (
     IncidentPageFactory,
     TopicPageFactory,
 )
+from unittest import mock
 import json
 
 User = get_user_model()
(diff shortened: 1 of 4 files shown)


SaptakS (changes requested):
I think you forgot to add the URL in the urls.py file. That should also fix the test.

maeve-fpf:
Can someone point me to where reading in the version file was previously tested? It looks like breaking that out into a function decreased coverage (so that's failing), but only (going by line numbers) the exception-handling part, which was there before.

chigby:
Looking at this a bit and I'm actually not sure it was previously tested? It's possible the diff-ing part of `diff-cover` is getting confused and moving the function looks, from its point of view, like adding it? That's my theory about why the check is failing.

I feel like we may as well write a test for this. Here's a quick-and-easy one I came up with, but feel free to improve on this or write your own, too:

At top of file:
```python
from unittest import mock
```

The test:

```python
@mock.patch('common.views.open')
def test_version_info_url_returns_200_status_filenotfound(self, mock_open):
    mock_open.side_effect = FileNotFoundError
    self.response = self.client.get('/health/version/')
    self.assertEqual(self.response.status_code, 200)
```

SaptakS:
> It's possible the diff-ing part of diff-cover is getting confused and moving the function looks, from its point of view, like adding it

I agree. I think this is what's happening as well. Coverage is considering a complete new critical path since it's a function now.

SaptakS (approved):
Looks good to me.

PR #1135, opened 2021-09-29
Relationship endpoints

### LINKED ISSUE #1061: [API] Allow query by relation names and/or provide a lookup for object IDs

chigby:
Fixes #1061

This PR adds read-only endpoints for the following incident-related items:

| Related Entity | Example URL | 
| ------------------ | --------------------|
| Journalists | http://localhost:8000/api/edge/journalists/ |
| Institutions | http://localhost:8000/api/edge/institutions/3/ |
| Government Workers | http://localhost:8000/api/edge/governmentworkers/ |
| Charges | http://localhost:8000/api/edge/governmentworkers/ |
| Nationalities | http://localhost:8000/api/edge/nationalities/ |
| Politicians or Public Figures | http://localhost:8000/api/edge/politicians/1/ |
| Venues | http://localhost:8000/api/edge/venues/ |
| Equipment | http://localhost:8000/api/edge/equipment/6/ |

There are two "views" for each item: a list view and a detail view. The list view is the "base" view (e.g. `/venues/`) and the detail view is the view for a single item that can be reached by appending the ID of the entry (e.g. `/venues/33/`).

Right now the views are not paginated, meaning the list view lists all the entities. It's a lot less data-per-item than Incidents, but it potentially is a lot of values. If we do want to add pagination, it would not be very hard to do, though I am not sure we would want to apply cursor-based pagination to the title/name fields, since the title/name field feels like it is not as unchanging as we might want, though possibly I am overworrying about that. It might actually be fine.

Addendum: I also did some work to try to reduce some of the randomness in the test suite by moving the random/realistic incident-related factories into a file called `devdata.py`. I did something similar for the factories in the `common` app a little while ago (see 17164b02e398863ffa8d0134bc23f337008eb5d7). I hope that this will continue to reduce the annoying test failures on CI and locally due to non-deterministic tests. One thing to note is that I put `devdata.py` into the coverage omit list because (1) the factories were formerly in the `/tests/` directory which is also omitted from coverage analysis, and (2) this file is not really part of our application code, it's more of a library to help build realistic test data and it would definitely require a bit more work to ensure every code pathway is covered, which is not really the goal of our test suite anyway (it's to ensure our application works properly).

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,6 +4,7 @@ omit =
     */migrations/*.py
     */tests/*.py
     tracker/settings/*.py
+    incident/devdata.py
 skip_empty = True
 show_missing = True
 
--- a/common/management/commands/createdevdata.py
+++ b/common/management/commands/createdevdata.py
@@ -28,7 +28,7 @@ from common.devdata import (
 from forms.models import FormPage
 from home.models import HomePage, HomePageFeature
 from incident.models import IncidentIndexPage, IncidentPage
-from incident.tests.factories import IncidentIndexPageFactory, IncidentLinkFactory, MultimediaIncidentUpdateFactory, MultimediaIncidentPageFactory
+from incident.devdata import IncidentIndexPageFactory, IncidentLinkFactory, MultimediaIncidentUpdateFactory, MultimediaIncidentPageFactory
 from menus.models import Menu, MenuItem
 
 
--- a/incident/api/serializers.py
+++ b/incident/api/serializers.py
@@ -74,6 +74,11 @@ class JournalistSerializer(serializers.Serializer):
     title = serializers.CharField()
 
 
+class InstitutionSerializer(serializers.Serializer):
+    id = serializers.IntegerField()
+    title = serializers.CharField()
+
+
 class BaseIncidentSerializer(serializers.Serializer):
     title = serializers.CharField()
     url = serializers.SerializerMethodField()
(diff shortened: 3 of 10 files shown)


harrislapiroff:
I know this is handled in a slightly different way, but can we also get a `categories` endpoint in this PR? It would be good to be able to get a complete list of categories without having to analyze the incidents data.

chigby:
@harrislapiroff Added categories to the set of available endpoints. Let me know if you want different fields available, such as quick facts or statistics items.. not sure exactly if those are needed or desired for our current purpose.

harrislapiroff (commented):
Reviewed this for functionality and it looks good. @SaptakS will you give the code a closer pass just in case? I took a very cursory look.

SaptakS (approved):
Looks good to me code and implementation wise.

PR #1143, opened 2021-10-25
Updates code to make it compatible with newer version of wagtail-autocomplete

### LINKED ISSUE #1138: Wagtail autocomplete makes multiple requests for deleted items
We were seeing a spike in `/autocomplete/objects/` path when opening an all-incidents page with an autocomplete filter for an item that was deleted. When autocomplete sees that the request returned with empty items list, the autocomplete attempts to fetch info again and again, hence creating huge number of requests to `/autocomplete/objects/`

The two things we need to decide is:
- Find if we can send a better error response in case the item being filtered by is deleted.
- If we can't provide a better response, restrict the requests to `n` failures, after which it shouldn't send any more requests.

SaptakS:
Blocked on a newer release of wagtail-autocomplete including https://github.com/wagtail/wagtail-autocomplete/pull/101 . That will fix https://github.com/freedomofpress/pressfreedomtracker.us/issues/1138

In this PR, updated the code to remove and change the parts deprecated by edit handlers and in turn by wagtail-autocomplete in newer versions.

### DIFF
--- a/common/forms.py
+++ b/common/forms.py
@@ -21,7 +21,7 @@ class BaseMergeForm(forms.Form):
             widget=type(
                 '_Autocomplete',
                 (Autocomplete,),
-                dict(target_model=self.merge_model, can_create=False, is_single=False)
+                dict(target_model=self.merge_model)
             ),
             label='{} to merge'.format(capfirst(self.merge_model_name))
         )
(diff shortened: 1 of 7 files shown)


SaptakS:
Closing this since these updates have been made as part of https://github.com/freedomofpress/pressfreedomtracker.us/pull/1156 . Will need to make a separate PR to update to the fixed wagtail-autocomplete version for the deleted items.

PR #1145, opened 2021-10-25
Mailchimp emails

### LINKED ISSUE #1144: Add ability to replace current PFT mailing list form with mailchimp form
- [ ] The mailchimp form should sport the same functionality as the freedom.press mailchimp form
   - [ ] options for name and email or email only
   - [ ] ability to specify a group for subscribers to be added to
- [ ] Options should be configurable on the `EmailSettings` object
- [ ] If no mailchimp configuration is specified, display the current form, otherwise display only the mailchimp form

chigby:
Fixes #1144 

This pull request adapts our Mailchimp enabled email signup form from freedom.press (see https://github.com/freedomofpress/freedom.press/pull/1005 and https://github.com/freedomofpress/freedom.press/pull/986) to work on this site. A notable difference is that if a MC form is selected, then the legacy form will not be displayed. Otherwise the implementation is the same. I've tried to keep the spirit of the old design but added input labels when more than one field is present:

image

If one field is present, the new form looks the same as the old form.


harrislapiroff (approved):
With my style change, this is ready to merge!

Issue #1117, opened 2021-08-16
[API] Update consumer documentation


harrislapiroff:
[The documentation](https://pressfreedomtracker.us/data/) needs to be update to reflect recent changes including

* Default pagination limit

chigby:
I think the default pagination limit docs were updated in https://github.com/freedomofpress/pressfreedomtracker.us/commit/6b54699b6ff309189761bbf1447444a7df4d077c -- part of #1151. Are there other things that we ought to have for this issue?

harrislapiroff:
I think we can probably close this in that case!

PR #1151, opened 2021-11-10
API documentation and Swagger (initial pass)

chigby:
Refs #1117
 
I'm making a PR for this to get some additional eyes on the Swagger. I'd say this is ready for review but I'm not totally sure it's ready to merge, if that makes any sense. I guess this PR is looking for feedback moreso than it's looking for fast-track-to-production. But if we think it's ready then, sure, let's merge it.

Anyway, this PR adds:

* A markdown file containing some initial static documentation. I wrote this _before_ going down the rabbit hole of auto-schema generation
* DRF Spectacular, a version < 1.0 package for introspecting a schema out of a Django Rest Framework configuration. This schema is compatible with OpenAPI 3.0.x (notably _not_ OpenAPI 3.1) and can be used by various frontend UIs for browsing this "live" piece of documentation. (Though I will say the "liveness" is slightly compromised by our setup, see consideration 3 below). 
* A Swagger URL for our API at http://localhost:8000/api/schema/swagger-ui
* A Redoc URL for our API at http://localhost:8000/api/schema/redoc (these are both "included" with DRF spectacular, which is nice, but ultimately we probably only want one of them).
* A raw schema URL for our API at http://localhost:8000/api/schema/
* Some code for converting between our Incident filters and the OpenAPI formatted [parameter objects](https://swagger.io/specification/#parameter-object). This is slightly tricky, and probably took the most time out of everything I worked on for this PR because I was trying to coax a nice UI for choosing valid parameters out of Swagger, and only in some cases I was able to succeed. Particularly, we are now allowing relationship filters (such as "state") to accept values of either the name of a state or its database ID. It doesn't seem like Swagger or ReDoc have a great way for inputting this type of "either or" value. The attitude is more like: "here's a text box."
* Proper versioning for our API. Previously we were using a static URL prefix, but now we are taking advantage of DRF's actual versioning features. This lets our schema express the version correctly.

Further considerations:

1. I'm pretty sure that right now the Swagger code is being pulled from a CDN. According to the docs, it's possible to self host these static files but that's not in this PR. See the documentation: https://drf-spectacular.readthedocs.io/en/latest/readme.html#self-contained-ui-installation
2. There is definitely more "explanatory" documentation that could be added to the Swagger UI, akin to the `help_text` fields on the wagtail UI, but I have not had time to sit down and write any. So looking at the API explorer right now is somewhat utilitarian.
3. Our filters are not actually static, and cannot be determined simply from the DRF configuration. They depend on the Wagtail site settings, which is stored in the database. I think one consequence of this is that changes made to the database after the schema is computed will not be reflected immediately. In practice, I had to restart the server. It seems to me that the schema is calculated during the Django startup phase. I am not completely sure about this and some more investigation can be done.

### DIFF
--- a/api_docs/Incidents.md
+++ b/api_docs/Incidents.md
@@ -1,0 +1,271 @@
+# Incidents
+
+Incidents are occurrences of violations of the freedom of the press committed by government authorities or private individuals.  The API contains information about when the incident took place, who was involved, and what kind of violation happened.
+
+## Endpoints
+
+```
+GET incidents/
+```
+
+Gets a list of all incidents.
+
+## Parameters
+
+Parameters can be applied to the API request as part of the URL's query string.  All parameters are optional.
+
+### Query parameters
+
+These parameters can be used to customize the query and response that the API returns.
+
+
+
+| Query string parameter | Type    | Description                                                                                                                                                                           |
+|------------------------|---------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `limit`                | integer | The maximum number of results to return per page. If this parameter is not provided, then the set of results will not be paginated.                                                   |
+| `fields`               | string  | A comma-separated list of field names. The incidents returned in the response will include only the ones given in this parameter. The set of available field names is detailed below. |
+| `format`               | string  | Describes the format of the data in the response. Can be either `json` (the default) or `csv`.                                                                                        |
+
+
+### Filtering parameters
+
+These parameters can be used to filter which incidents are returned in the response.  Only incidents that match the values given will be included in the response.
+
+TODO: fill in this section.
+
+## Sample request
+
+```
+curl -X GET "https://pressfreedomtracker.us/api/edge/incidents/?limit=2&format=json"
+```
+
+## Sample response
+
+Here is a sample response:
+
+```
+{
+  "next": "http://localhost:8000/api/edge/incidents/?cursor=cD0yMDIxLTAyLTIwLTQ3Mzk%3D&format=json&limit=2",
+  "previous": null,
+  "results": [
+    {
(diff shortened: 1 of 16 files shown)


chigby:
I think CI is failing because, on CI, the database does not exist (or at least migrations have not been run) at the time that the schema is initialized. It should still work locally (it does for me, anyway) but we will have to come up with something else, perhaps, for these checks to pass.

chigby:
Working on removing the conflicts from this and verifying that it still works how we want

SaptakS (approved):
Looks good to me. I haven't reviewed the copy of the documentation a lot apart from the technalities. Maybe someone else wants to take a look at that in case they want to change anything.

One thing I am a little icky about is using CDN (which @chigby also mentioned). I am curious after seeing the drf-spectacular-sidecar repository, if we can directly use the individual npm modules? Seems like https://www.npmjs.com/package/swagger-ui-dist and https://www.npmjs.com/package/redoc are available as npm packages. But not sure how much extra work that would need. I am also fine with just using drf-spectacular-sidecar (though not the biggest fan of using python package to handle ui library). But I don't think it's a huge blocker in merging (unless we hear that jsdeliver got hacked, in which case it's a trouble).

SaptakS:
Also @chigby you will need to resolve some merge conflicts.

chigby:
It actually seems like getting sidecar configured was pretty easy! Try loading the docs locally again (after rebuilding the django container) and see if it's making requests to a 3rd party. For me, it's not anymore. :)

SaptakS:
Looks good to me. @harrislapiroff unless you want to take a look at the copy of the docs, I can merge this PR

harrislapiroff:
Go ahead and merge! I’ll look live

PR #1156, opened 2021-11-22
Update npm pip depn

SaptakS:
Refs freedomofpress/fpf-www-projects#231

- Updates npm packages without pinning
- Updates makefile to be more consistent with other repos and to add pip upgrade commands
- Updates all pip dependencies
- Makes wagtail autocomplete related changes
- Updates sass deprecations (NOTE: it still throws deprecation warnings because some of the dependencies have not updated their sass code.)
- Pins psycopg2 to below 2.9 since it doesn't work with django 2.2


chigby (approved):
Did some manual checking of the affected areas, and everything seems to be working normally. It's a shame we can't get rid of those deprecation warnings, though I suppose that's something we could also address by moving away from those libraries that did not update (which we might have to if they prevent us from updating to the sass where they remove that deprecated feature).

Issue #1148, opened 2021-11-08
Adjust legal case options


harrislapiroff:
We've got some requests from the tracker team to make the following changes to `IncidentPage`s:

- [ ] Add field "Case Number": should be a charfield, put in admin under Legal Case
- [ ] Add field "Status": should be a multiple choice field with checkbox options ongoing, settled, dismissed, unknown, and appealed, put in admin under Legal Case
- [ ] Deprecate "Case venue" (remove it from the admin, api, and display on the frontend and add a comment to the code that we should delete it in a month)
- [ ] Deprecate "Lawsuit name" (same as above)

harrislapiroff:
Would be nice to have this done by first week in Dec

harrislapiroff:
We should remove the lawsuit name and venue fields still. These aren't editable in the admin AFAIK so editors are definitely not using them.

PR #1157, opened 2021-11-23
Initial implementation of legal case options update

SaptakS:
Refs #1148 

- Adds case_number and case_statuses
- Removes and deprecates lawsuit_name and venue from most part


incident/templates/incident/_lawsuit_rows.html
```diff
@@ -1,29 +1,32 @@
...
-					<li class="data-table__list-item data-table__list-item--inline">
-						<a href="{% pageurl page.get_parent %}?venue={{ venue.pk }}">{{ venue }}</a></li>
-				{% endfor %}
+			{% for case_status in page.case_statuses %}
+			<a href="{% pageurl page.get_parent %}?case_statuses={{ case_statuses }}">
+				{% get_case_status_display case_status as case_status_display_name %}
+				{{ case_status_display_name|capfirst }}
+			</a>
+			{% if not forloop.last %}, {% endif %}
+			{% endfor %}
```
chigby, line 27: If an incident has more than one status, I'm noticing some extra spacing around the comma in the list:

image

There are probably multiple ways to address this but I was able to make it go away by removing the whitespace in the template like this:

```
{{ case_status_display_name|capfirst }}</a>{% if not forloop.last %}, {% endif %}
```

Potentially the [spaceless template tag](https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#spaceless) could be used as well.

chigby (changes requested):
I think almost everything is looking good here, except for one minor whitespace issue which I've explained below.

chigby (approved):

PR #1164, opened 2021-12-08
Adds type of case field

### LINKED ISSUE #1162: Add "Type of case" to legal case details
I missed this in the initial spec of #1148 but we also want a choice field for "Type of case": options should be "Civil" and "Class Action"

### LINKED ISSUE #1163: Make cases statuses searchable
Currently case status is linked in the sidebar of incidents, but it doesn't lead to a properly filtered page when you click on it:

image
[Example](https://pressfreedomtracker.us/all-incidents/journalist-graduate-student-stopped-secondary-screening-electronic-devices-searched/)

Normally to enable this functionality we'd add "Case statuses" as a "General incident filter" in Wagtail's admin under Settings > General incident filters, but it doesn't seem to be an option in the dropdowns there. (The other field @SaptakS added—"Case number"—is there. Not sure what needs to be done to register this field.)

SaptakS:
Fixes #1162 
Fixes #1163


harrislapiroff (approved):
Code looks good. Gonna test on staging.

PR #1168, opened 2021-12-09
Uses display name for case type

### LINKED ISSUE #1166: Render display value for Case Type in template
We should be rendering the display value (title cased) instead of the database value (all capsed) for case types on incident pages. Current appearance:

image

Should be `Type of case: Civil` instead.

SaptakS:
Fixes #1166

### DIFF
--- a/incident/api/serializers.py
+++ b/incident/api/serializers.py
@@ -112,7 +112,7 @@ class BaseIncidentSerializer(serializers.Serializer):
     detention_date = serializers.DateField()
     unnecessary_use_of_force = serializers.BooleanField()
     case_number = serializers.CharField()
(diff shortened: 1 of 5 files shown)


harrislapiroff:
@SaptakS Can you just add one more fix to this PR? The PFT editors want Case number, Case status, and Type of case to be at the bottom of the table—only Tags should be below it.

harrislapiroff:
And also move type of case above the other two fields.

SaptakS:
@harrislapiroff done!

harrislapiroff:
@SaptakS Am I correct that this PR is no longer relevant with the redesign? If so, go ahead and close it, sorry for the extra work 😅

PR #1175, opened 2021-12-13
Adds eslint and stylelint

SaptakS:


client/statistics/js/searchstats.js
```diff
@@ -1,19 +1,19 @@
...
-const React = window.React
-const Modifier = window.DraftJS.Modifier
-const EditorState = window.DraftJS.EditorState
-const TooltipEntity = window.draftail.TooltipEntity
-const Icon = window.wagtail.components.Icon
+const { React } = window
+const { Modifier } = window.DraftJS
+const { EditorState } = window.DraftJS
+const { TooltipEntity } = window.draftail
+const { Icon } = window.wagtail.components
```
harrislapiroff: I'm intrigued by what the rule is that creates this structure. Also noting that we could even shorten one of these lines to

```js
const { EditorState, Modifier } = window.DraftJS
```

which is more parallel with how we usually do imports. Just an optional nitpick though.

harrislapiroff (approved):
Looks great! Feel free to merge

client/statistics/js/searchstats.js
SaptakS: It was being caused by https://eslint.org/docs/rules/prefer-destructuring. But since I think we will be using destructuring probably quite a lot in our code, I have disabled that in eslint for now, and reverted this code changes.

PR #1178, opened 2021-12-15
Adds STYLEGUIDE variable to control showing styleguide

SaptakS:
Adds a `production-styleguide.py` with STYLEGUIDE = True. This STYLEGUIDE variable is being used by to turn on just styleguide path without using the rest of DEBUG settings.


harrislapiroff (approved):
Looks good

Issue #1173, opened 2021-12-13
Redesign: Incident Card, Part 1


harrislapiroff:
- [ ] Big
- [ ] Big w/ photo
- [ ] Small
- [ ] Small w/ photo

[Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=2710%3A22015)

When closing this ticket, open a ticket **Redesign: Database Card** and the same figma link

chigby:
Question about these cards: there are 4 designs here, but they differ in very small ways. Do we want to create four separate HTML files that can be separately invoked via an `include` statement (for example `{% include "incident_card_big_photo.html" %}`)? From looking at these, the difference between the big and small cards is the width, which I would have assumed would be non-fixed and based on the properties of the element containing the card (or something like a media query). The difference between the photo and non-photo cards is whether to use display a photo or a teaser text. But these are both optional fields on the `Incident` model, and (again) I would have assumed that the card template would have internal conditional logic for prioritizing the display of these fields. So maybe I have some misunderstandings of the main ideas here, or possibly just need some clarification of exactly what is needed for this issue to be resolved.

SaptakS:
I would think of passing some arguments that make distinction maybe? Like if `{% include "incident_card.html" with photo=incident.image and size="big" %}` is called, then we can check in the template these two parameters and decide the small changes? Or basically just add this arguments to the the class names somehow that we can different it in the sass file? What do you think @harrislapiroff ?

harrislapiroff:
After discussing in web sync, we've decided to use a single template with context variables for whether to display a photo and whether to display large or small (which only changes the heading size) as @SaptakS proposed

PR #1179, opened 2021-12-15
Incident cards

chigby:
Refs #1173

### DIFF
--- a/incident/templates/incident/incident_card.html
+++ b/incident/templates/incident/incident_card.html
@@ -1,6 +1,6 @@
 {% load wagtailimages_tags %}
 {% with size=size|default:"big" %}
-	<div class="incident-card">
+	<article class="incident-card">
 		{% if photo %}
 			{% image photo fill-665x440 as x1 %}
 			<img
@@ -54,5 +54,5 @@
 				{% endfor %}
 			</ul>
 		{% endif %}
-	</div>
+	</article>
 {% endwith %}
--- a/incident/templates/incident/incident_card.html
+++ b/incident/templates/incident/incident_card.html
@@ -28,10 +28,16 @@
 		<dl class="dl-inline">
 			<dt>Updated on</dt>
 			<dd>
-				{{ incident.latest_update|date:"F j, Y" }}
+				<time datetime="{{ incident.latest_update|date:"c" }}">
+					{{ incident.latest_update|date:"F j, Y" }}
+				</time>
 			</dd>
 			<dt>Date of Incident</dt>
-			<dd>{{ incident.date|date:"F j, Y" }}</dd>
+			<dd>
+				<time datetime="{{ incident.date|date:"c" }}">
+					{{ incident.date|date:"F j, Y" }}
+				</time>
+			</dd>
 			<dt>Location</dt>
 			<dd>{{ incident.city }}, {{ incident.state.name }}</dd>
 			{% if incident.targeted_journalists.all or incident.targeted_institutions.all %}
--- a/incident/templates/incident/incident_card.html
+++ b/incident/templates/incident/incident_card.html
@@ -9,7 +9,7 @@
 				src="{{ x1.url }}"
 			>
 		{% endif %}
-		<h1 class="card-title-{{ size }}">{{ incident.title }}</h1>
+		<h3 class="card-title-{{ size }}">{{ incident.title }}</h3>
 
 		{% if not photo %}
 			<p>
--- a/incident/templates/incident/incident_card.html
+++ b/incident/templates/incident/incident_card.html
@@ -1,0 +1,58 @@
+{% load wagtailimages_tags %}
+{% with size=size|default:"big" %}
+	<div class="incident-card">
+		{% if photo %}
+			{% image photo fill-665x440 as x1 %}
+			<img
+				class="card-image-{{ size }}"
+				alt="{{ x1.alt }}"
+				src="{{ x1.url }}"
+			>
+		{% endif %}
+		<h1 class="card-title-{{ size }}">{{ incident.title }}</h1>
+
+		{% if not photo %}
+			<p>
+				{{ incident.teaser }}
+			</p>
+		{% endif %}
+
+		<ol>
+			{% for categorization in incident.categories.all %}
+				<li>
+					{{ categorization.category.title }}
+				</li>
+			{% endfor %}
+		</ol>
+
+		<dl>
+			<dt>Updated on</dt>
+			<dd>
+				{{ incident.latest_update|date:"F j, Y" }}
+			</dd>
+			<dt>Date of Incident</dt>
+			<dd>{{ incident.date|date:"F j, Y" }}</dd>
+			<dt>Location</dt>
+			<dd>{{ incident.city }}, {{ incident.state.name }}</dd>
+			{% if incident.targeted_journalists.all or incident.targeted_institutions.all %}
+				<dt>Targets</dt>
+				{% for journalist in incident.targeted_journalists.all %}
+					<dd>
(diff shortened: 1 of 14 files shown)


SaptakS:
Updated the base to point to `redesign` instead of `develop`

chigby:
@SaptakS This is ready for review except for the failing test :sweat_smile: I'll figure out what's going on with that tomorrow, but the rest of the code is in a good spot I think.

incident/templates/incident/incident_card.html
```diff
@@ -0,0 +1,58 @@
...
+	<div class="incident-card">
+		{% if photo %}
+			{% image photo fill-665x440 as x1 %}
+			<img
+				class="card-image-{{ size }}"
+				alt="{{ x1.alt }}"
+				src="{{ x1.url }}"
+			>
+		{% endif %}
+		<h1 class="card-title-{{ size }}">{{ incident.title }}</h1>
```
SaptakS: Every HTML page should ideally have only 1 `<h1>` tag usually reserved for the page title. That helps in maintaining the structure of the page such that page title is h1, then all sub-sections have h2, and then anything inside that have h3-h6. So given how most templates look, I will recommend using `<h3>` here. Some heading best practices are mentioned here: https://accessibility.psu.edu/headingshtml/

incident/templates/incident/incident_card.html
```diff
@@ -0,0 +1,58 @@
+{% load wagtailimages_tags %}
+{% with size=size|default:"big" %}
+	<div class="incident-card">
```
SaptakS: Though I don't have super strong opinion here, but I feel [`<article>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) gives a better semantic meaning. Thought?

incident/templates/incident/incident_card.html
```diff
@@ -0,0 +1,58 @@
...
+		{% endif %}
+		<h1 class="card-title-{{ size }}">{{ incident.title }}</h1>
+
+		{% if not photo %}
+			<p>
+				{{ incident.teaser }}
+			</p>
+		{% endif %}
+
+		<ol>
```
SaptakS, line 24: Are the categories in some specific order for incidents? I remember that each incident has like a main category or something. In that case, maybe makes sense if that's the first category shown. Else maybe we want to use `<ul>` instead?

incident/templates/incident/incident_card.html
```diff
@@ -0,0 +1,58 @@
...
+				<li>
+					{{ categorization.category.title }}
+				</li>
+			{% endfor %}
+		</ol>
+
+		<dl class="dl-inline">
+			<dt>Updated on</dt>
+			<dd>
+				{{ incident.latest_update|date:"F j, Y" }}
```
SaptakS: It might be good to wrap this in [`<time>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time). So something like:
```html
<dd>
  <time datetime="{{ incident.latest_update|date:"c" }}">
    {{ incident.latest_update|date:"F j, Y" }}
  </time>
</dd>
```

incident/templates/incident/incident_card.html
```diff
@@ -0,0 +1,58 @@
...
+			</dd>
+			<dt>Date of Incident</dt>
+			<dd>{{ incident.date|date:"F j, Y" }}</dd>
```
SaptakS: Same as above.

incident/templates/incident/incident_card.html
```diff
@@ -0,0 +1,58 @@
...
+				{% for institution in incident.targeted_institutions.all %}
+					<dd>{{ institution.title }}</dd>
+				{% endfor %}
+			{% endif %}
+		</dl>
+
+		{% if incident.tags.all %}
+			<ul class="tag-list">
+				{% for tag in incident.tags.all %}
+					<li><a href="#" class="btn btn-tag">#{{ tag.title }}</a></li>
```
SaptakS, line 63: Just noting here that the href is a TODO to be added once we add urls and pages for individual tags.

incident/templates/incident/incident_index_page.html
```diff
@@ -0,0 +1,9 @@
+{% extends "super.html" %}
+
+{% load webpack_loader %}
+
+{% block body %}
+	{% for incident in entries_page %}
+		{% include "incident/incident_card.html" with incident=incident size="small" photo=incident.teaser_image only %}
+	{% endfor %}
```
SaptakS, line 8: I know this page's entire design implementation is far from finished, so I am okay merging this for now, but I think later, we should ensure a `<ul>`/`<ol>` structure depending on whether its sorted by certain attribute, or whether it's a curated list.

incident/templates/incident/incident_card.html
SaptakS: Also, I think based on the design, at least the heading part should be inside `<a>` tags pointing to the particular incident page. I am mostly neutral about whether the entire card should be inside an `<a>` tag (slightly more inclined towards no). Also given the design, the anchor tag under `.incident-card` will have slightly different design (`text-decoration: none` in beginning and `underline` on `:hover`. And unlike the default link style, `background-color: transparent` on `:hover`.

The hover styles and focus styles are the cards below the first card shown in the design respectively. Unless we do `<a>` on entire card, the focus style is messy to attain. But I am personally okay if the focus style applies only to the heading text (which it automatically will based on default `<a>`). @harrislapiroff @chigby do you have any strong preferences?

incident/templates/incident/incident_card.html
SaptakS: Also all the incident cards have a border line at the top. I think can be attained by adding a `border-top: 1px solid; border-color: $black;`

SaptakS (changes requested):
This looks good! I like that the width is kept fluid, so we can design that based on the parent element. Left a few comments, mostly some HTML semantic stuff.

incident/templates/incident/incident_card.html
chigby: The first category being the "main" category was what motivated putting these in an `<ol>`, but the more I think about it, that might not be the best way of representing that idea.

incident/templates/incident/incident_card.html
chigby: Thanks for pointing out the hover/focus styles, I completely missed that. Though one thing I'm seeing is that the `<a>` tag has a default font size that's taking precedence over the `card-title-{big,small}` typography styles. How do you think we should address that?

incident/templates/incident/incident_card.html
chigby: Updated this PR with the time tags. Great idea!

incident/templates/incident/incident_card.html
chigby: Added both of these to the PR.

incident/templates/incident/incident_card.html
chigby: I guess maybe the class could be applied to the `<a>` instead?

incident/templates/incident/incident_card.html
SaptakS: But I think it's okay to leave it `<ol>` then given there is some order. So let's keep it that way.

incident/templates/incident/incident_card.html
chigby: After some discussion, also added the hover styles into the PR.

chigby:
Should be ready for re-review and/or merging.

SaptakS:
All looks good to me. Only I think the article cards need a padding-top of 2rem based on the design. Also the link class needs a display: inline-block so that border in the :focus shows correctly. After that, this should be good to merge

chigby:
Added those

SaptakS (approved):
Looks good to me

PR #1180, opened 2021-12-20
Fixes guidelines for the show more button

SaptakS:
I noticed after the deploy that some copy-paste text was left in the guidelines for the show more button. Fixing it here.


chigby (approved):
Looks good to me!

Issue #1172, opened 2021-12-13
Redesign: Footer


harrislapiroff:
Depends on #1170 

- [ ] Desktop
- [ ] Mobile
- [ ] Tablet

[Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=4051%3A72736)

PR #1181, opened 2022-01-03
Redesign footer

SaptakS:
Refs: #1172 

Can be seen in http://127.0.0.1:8000/styleguide/#footer once run locally.

Added 3 separate template includes for sitation, email signups and footer info for reusability.


SaptakS:
@harrislapiroff the thing I was telling might benefit from some decision of yours is that the input fields dont have two separate clicked and focused states. Even when we are clicking the input element, it's getting focussed. So `:focus` is what gets applied. There are [two different designs](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3086%3A26844) however. Right now I have applied the focus design since otherwise tabbing through doesn't make it prominently different from non-focused state.

This is how it looks for both clicked and focussed
image

common/templates/common/_email-signup.html
```diff
@@ -0,0 +1,57 @@
+{% with email_settings=settings.emails.EmailSettings %}
+	{% for group in email_settings.mailchimp_groups.all %}
+		<section class="emails-signup">
+			<h2 class="paragraph-subtitle">{{ email_settings.signup_prompt }}</h2>
+			<form action="https://press.us6.list-manage.com/subscribe/post?u=19b87eae477821267dea69438&amp;id=c8d8f610fb" method="post" id="mc-embedded-subscribe-form" name="mc-embedded-subscribe-form" class="validate emails-signup__form" target="_blank" novalidate>
```
chigby, line 5: So, this is maybe not something we need to worry about right now, but it's probably something we will want to do before the redesign is done, but on freedom.press we are in the process of changing over from forms that submit directly to Mailchimp to ones that use the API. See here: https://github.com/freedomofpress/freedom.press/pull/1036 

Once that PR is approved and merged we should probably apply the same changes to this site.

client/common/sass/components/_sr-only.sass
```diff
@@ -0,0 +1,8 @@
+.sr-only
```
chigby, line 1: Nice! I added something similar to this in the Menus PR, but I think we should probably use this one as all I did was copy the one from the old design.

chigby (approved):
The code looks good to me. I don't really have any general comments other than that! Nor do I have any thoughts on the focus/design question you raised. I'm in favor of merging but I'll also wait for @harrislapiroff to weigh in.

common/templates/common/_email-signup.html
SaptakS: Yes. I was thinking the same. Will need to make that change once it is merged.

PR #1182, opened 2022-01-03
Menus

### LINKED ISSUE #1171: Redesign: Menu
Depends on #1170 

- [ ] Desktop
- [ ] Mobile
- [ ] Tablet

Include javascript for mobile and tablet

[Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3128%3A29842)

chigby:
Fixes #1171 

This PR implements the mobile, tablet, and desktop top site menu as outlined in the issue/design. I've put the menu in the styleguide, as there's not really any other page to apply it to, so please go there to test it out.

Potential issues I know about:

1. There is no animation for the open/close icon changing state. Right now it just changes. I have not really looked into how to make that animation between two SVGs work, and I don't know how I'd go about it. Might be worth making into another issue.
2. The mobile/tablet menu list does not have a hover state. I was looking at the design document for this and it seems slightly tricky and like it might want to become its own component (e.g., it needs to be able to display logos on hover), so I have basically implemented the minimum necessary for this PR. Also potentially it doesn't matter since mobile/tablet interfaces don't really support a hover state (that I'm aware of).
3. The Styleguide page has horizontal scrollbars at mobile widths, which affects the menu. My assumption is that when more of the page layout code is complete this kind of thing will be prevented, so I have not attempted to provide my own slapdash fix for the Styleguide page. 
4. When the viewport is around 770-800 pixels wide, the buttons are a bit scrunched. My opinion here is we might want to think a little bit about how wide we consider a "tablet" for the purposes of our media queries. According to Firefox, the iPad's effective pixel width is 810 pixels, and right now that width is being shown the "desktop" layout (smaller than this and the scrunch starts to occur). This could also happen if the number/size of the menu items changes (see below).

So, if you have thoughts on any of the above, or if you have feedback about the implementation markup, SASS, Javascript, please let me know. I've tried to take into account relevant attributes for assistive technologies, ~added a `sr-only` class that we might want to use elsewhere (I just copied it from what was on the site before)~, and made the "vertical wipe" animation in the simplest way I could figure out. There might be other ways though. Any and all comments welcome regarding all of this: I want these foundational elements to be the best they can be.

Open question:

Are the menu items intended to be hard-coded or loaded from Wagtail settings? They are hard-coded right now, but we do have a system for allowing admins to input menu items. We would have to change that system a little bit, I think, to incorporate the ideas of "primary" and "secondary" navigation.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -21,6 +21,7 @@ module.exports = {
 		"window": true,
 		"HTMLInputElement": true,
 		"fetch": true,
+		"document": true,
 	},
 
 	"rules": {
--- a/client/common/icons/icon-menu-close.svg
+++ b/client/common/icons/icon-menu-close.svg
@@ -1,0 +1,5 @@
+<svg width="44" height="44" viewBox="0 0 44 44" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M14 31L30.9706 14.0294" stroke="black" stroke-width="3"/>
+<path d="M14 14L21.0711 21.0711" stroke="black" stroke-width="3"/>
+<path d="M21 21L30.8995 30.8995" stroke="black" stroke-width="3"/>
+</svg>
--- a/client/common/icons/icon-menu-open.svg
+++ b/client/common/icons/icon-menu-open.svg
@@ -1,0 +1,5 @@
+<svg width="44" height="44" viewBox="0 0 44 44" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M10 10H20" stroke="black" stroke-width="3"/>
+<path d="M10 22H34" stroke="black" stroke-width="3"/>
+<path d="M10 34H24" stroke="black" stroke-width="3"/>
+</svg>
--- a/client/common/js/common.js
+++ b/client/common/js/common.js
@@ -1,2 +1,2 @@
 import '../sass/common.sass'
-import './menu.js'
+import './menu'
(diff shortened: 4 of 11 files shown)


SaptakS (changes requested):
For the most part it looks great! Few things I noticed, in the mobile width, there seems to be an overflow and the "Go to the page" seems not visible. Are you noticing this too? I checked in Firefox. Also, in desktop mode, the "U.S. Press Freedom Tracker" seems to be wrapping for me instead of fitting in a single line.

Also, I think we need a design like [this](https://www.figma.com/proto/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3912%3A63655&scaling=min-zoom&page-id=3897%3A59348&starting-point-node-id=3897%3A59349) for the menu links in smaller screens.

chigby:
Pushed some changes to this PR to:

1. Prevent the title link from wrapping 
2. Prevent overflow on the styleguide page (which should make it so that the menu properly fits on the page on small screens and does not extend so far that the "Go to this page" is visible).

@SaptakS Do you think this PR should include a design for the menu links with extra info? Or should that be a separate PR for that version of the lists component?

chigby:
@SaptakS This is rebased now, conflicts resolved

SaptakS (approved):
Looks good to me. Let's get this merged.

PR #1184, opened 2022-01-05
Initial search bar component

chigby:
This pull request adds the initial scaffolding and styles for the search bar component. Here's what it looks like in the styleguide:

image

I recognize that this component is destined to gain its fancy search-hints functionality via React, but the specifications for that don't really exist yet, and the fallback for non-JS users will be a plain HTML form, which is what I've done here.

### DIFF
--- a/client/common/sass/components/_search-bar.sass
+++ b/client/common/sass/components/_search-bar.sass
@@ -3,7 +3,10 @@
 
 .search-button
 	position: absolute
-	right: 4px
-	top: 5px
+	right: 2px
+	top: 3.5px
 	border: 0
-	font-weight: 700
+	font-weight: 500
+
+	&:focus
+		top: 1px
--- a/client/common/sass/components/_search-bar.sass
+++ b/client/common/sass/components/_search-bar.sass
@@ -1,0 +1,9 @@
+.search-form
+	position: relative
+
+.search-button
+	position: absolute
+	right: 4px
+	top: 5px
+	border: 0
+	font-weight: 700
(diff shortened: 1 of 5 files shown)


client/common/sass/components/_search-bar.sass
```diff
@@ -0,0 +1,9 @@
+.search-form
+	position: relative
+
+.search-button
+	position: absolute
+	right: 4px
+	top: 5px
+	border: 0
+	font-weight: 700
```
SaptakS: Based on the designs, the font-weight should be 500. We are still falling back to Helvetica instead of Helvetica Nueue, so the font-weight 500 looks different in local implementation compared to the design given, but I think sticking with 500 will make it look more proper once we have the correct font.

client/common/sass/components/_search-bar.sass
```diff
@@ -0,0 +1,9 @@
+.search-form
+	position: relative
+
+.search-button
+	position: absolute
+	right: 4px
+	top: 5px
```
SaptakS: I think `top: 3.5px` and `right: 2px` (since border is 2px) looks more close to the design.

SaptakS (changes requested):
Left 2 minor nitpicking fixes. Apart from that, looks good to me.

client/common/sass/components/_search-bar.sass
SaptakS: Also, add:

```sass
&:focus
  top: 1px
```

client/common/sass/components/_search-bar.sass
chigby: That's a good point! I didn't consider a weight between 400 and 700 but it didn't really look right at either. This seems to work better, for sure.

chigby:
I added the adjusted spacing and font weight to this PR.

SaptakS (approved):
Looks good to me.

PR #1185, opened 2022-01-05
Blog card components

chigby:
This PR adds blog card components mostly based off of the model already established by the incident card components. I have added some additional helper styles for the images so that they don't end up on the same line as the card title when the width of the card is very wide (which might not happen anywhere besides the styleguide, but it's good to be prepared).

The component is demonstrated in the styleguide as well.

### DIFF
--- a/blog/templates/blog/blog_card.html
+++ b/blog/templates/blog/blog_card.html
@@ -32,5 +32,6 @@
 					{{ post.author.title }}
 				</dd>
 			{% endif %}
+		</dl>
 	</article>
 {% endwith %}
--- a/blog/templates/blog/blog_card.html
+++ b/blog/templates/blog/blog_card.html
@@ -10,16 +10,16 @@
 			>
 		{% endif %}
 		<a class="text-link--card" href="{% pageurl post %}">
-			<h3 class="card-title-{{ size }}">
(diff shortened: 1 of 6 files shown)


blog/templates/blog/blog_card.html
```diff
@@ -0,0 +1,36 @@
...
+				<time datetime="{{ post.publication_datetime|date:"c" }}">
+					{{ post.publication_datetime|date:"F j, Y" }}
+				</time>
+			</dd>
+			{% if post.author %}
+				<dt>Written by</dt>
+				<dd>
+					{{ post.author.title }}
+				</dd>
+			{% endif %}
```
SaptakS, line 34: Ending tag for `<dl>` is missing.

SaptakS (changes requested):
Left a nitpick comment. Apart from that, I see that the blog cards are missing the top border as shown in the figma design. Also the spacing between the heading and the teaser text is a bit much. This is because of the default h3 and p spacing. So would need to adjust for this particular case, such that the space is 12px as shown in the figma designs.

chigby:
Rebased this (to fix safety check) and also addressed the points mentioned. Ready for another review.

SaptakS (approved):
LGTM

PR #1189, opened 2022-01-10
Upgrades urllib3 to fix safety and docker build breaking

SaptakS:
Without updating urllib3, the dev-requirements was expecting a separtae version because of setuptools and since that was not already satisfied, both the local docker compose build and [safety ci check](https://app.circleci.com/pipelines/github/freedomofpress/pressfreedomtracker.us/1736/workflows/6ebf7100-702d-41f0-9442-d577fd9edabf/jobs/9890) was failing. This fixes that issue.


chigby (approved):
Looks like it fixes the check, therefore let's merge it

PR #1191, opened 2022-01-10
Adds categories symbol library

SaptakS:
- Adds all the svg icons
- Adds the icons as pseudo-elements. So, in future development, we don't need to add separate markups again for the icons, but just mention the class names as the category slug. 
 
PS: If a new category is added, we need to decide what icon that category gets, and set accordingly. Else it will appear as just normal links

I have listed how all the categories look with icons in the styleguide.

image


client/common/sass/components/_text-link.sass
```diff
@@ -18,3 +18,41 @@
 		&:hover
 			background-color: transparent
 			text-decoration: underline
+
+	&--category
+		&::before
+			vertical-align: middle
+			margin-right: 4px
+
+		&.arrest::before
```
chigby: So one thing I'm realizing here is that the category page `slug` fields in the dev data do not match the ones elsewhere. On staging, they are:

* `arrest-criminal-charge`
* `border-stop`
* `chilling-statement`
* `denial-access`
* `equipment-damage`
* `equipment-search-seizure-or-damage`
* `leak-case`
* `other-incident`
* `physical-attack`
* `prior-restraint`
* `subpoena`

This makes me wonder if it might be wise to de-couple of the name of the CSS class from the page slug? Like, maybe it could be a separate field that defines a set list of choices. Maybe though that's overdoing it if the slugs aren't expected to change at all, and we can update the dev data to match what's in prod/staging. On the other hand, I guess it depends a little on how we're going to use these icons, and if this is only confusing because of how we're implementing showing a complete icon list on the Styleguide page, and maybe elsewhere the icon class names will not be loaded from the CategoryPage model object at all.

What do you think about this? I think you are more familiar with the use of the icons across the whole site as described in the Figma design.

chigby (changes requested):
See comment: I think we might want to better define exactly how the icon class names are stored in the DB (or change them), if only because if we merge this as-is, the styleguide won't show the correct icons on staging.

client/common/sass/components/_text-link.sass
SaptakS: Yes. I was giving this thought as well. I am guessing, everytime we will have a CategoryPage model object when we are trying to show a category link itself? For the rest of the places like filters, etc., I feel the implementation for the rest places apart from the category link will need to be separate from this anyways.

As of the field, I was also thinking if it makes sense to add a separate field, but then again since there was already a unique field, I just wanted to use slug. I feel we can generalize to use the same slug everywhere? I am not sure how often we plan to change slugs, but if not very often, I would just keep using that. But if we plan to change slug pretty often, I think we can add a separate field. @harrislapiroff any ideas? I definitely think adding a separate field is probably the safe way to go about it.

SaptakS:
I have modified the code such that there is now a settings to select the icons from each category page similar to the page color settings. Also removed all page color related code in the process. That does bring some change in the API schema, but we may need to worry that later?

For now, I have kept the icon names as the category names corresponding to the icons provided in design. I have also updated createdevdata to assign icons (though this happens randomly, so may not assign the icon that should probably be with the category, but can be changed from settings). I have also listed all the icons with their names in the styleguide. Also, decoupled the design from the text-links, so that the same classes can be used with or without it being a link.

image

chigby (approved):

chigby:
CI failing due to conflicting migrations:

```
CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0081_adds_index_entries, 0081_auto_20220119_1522 in common).
```

Can be merged once this is fixed and passing.

SaptakS:
Should be ready for merging now

PR #1192, opened 2022-01-11
Adds microanimations for components

SaptakS:
- Adds microanimations for all the buttons (The text links have instant changes and no animations)
- Adds microanimations for incident cards and blog cards (Added the animation at the link level instead of the whole card parent element. This is because we can't control the parent element's animation based on the child element's hover state)


chigby (approved):
Buttons look great! And the card animations might need to be adjusted depending on how they interact with a larger flex layout container, but we can deal with that if it happens.

Issue #1193, opened 2022-01-11
Categories block


chigby:
Create categories block as outlined in the Figma design.

Desktop:
![image](https://user-images.githubusercontent.com/561931/149016608-e62dd5ac-5ff1-4b55-adfe-188c4466388c.png)

Mobile:
![image](https://user-images.githubusercontent.com/561931/149016958-0f4ea60c-491c-4203-9139-a941f64f6d94.png)

Tablet:
![image](https://user-images.githubusercontent.com/561931/149017012-72cbb64c-74c2-476c-90e2-10b9d41be11d.png)

chigby:
Question for @harrislapiroff and @SaptakS and anyone else: is the above layout actually possible using CSS only? I am having a hard time getting the borders on the category items to "collapse" as shown -- meaning arranging it so that the borders between the items is the same as the border around the outside. Everything I've tried ends up with "doubled" internal borders or changing the design somehow, such as forcing each row to take up the same width, as shown:

![image](https://user-images.githubusercontent.com/561931/149019641-5c21f976-d117-4022-ab2b-607c3345a91d.png)

Any thoughts on this one? My CSS may not be up to the task here.

SaptakS:
This should be possible to do. Let me look into it a bit more and might have a better solution.

PR #1194, opened 2022-01-12
Categories block

chigby:
Refs #1193

### DIFF
--- a/client/common/sass/components/_categories-block.sass
+++ b/client/common/sass/components/_categories-block.sass
@@ -6,16 +6,16 @@
 	margin: 72px 5rem
 	padding: 0
 
-.categories-block--item
-	list-style-type: none
-	display: inline-block
-	text-indent: 0
-	padding: 1rem
-	border: solid 3px
-	border-color: $black
-	margin-left: -3px
-	margin-top: -3px
-	font-weight: 700
+	&--item
+		list-style-type: none
+		display: inline-block
+		text-indent: 0
+		padding: 1rem
+		border: solid 3px
+		border-color: $black
+		margin-left: -3px
+		margin-top: -3px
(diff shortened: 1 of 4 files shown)


SaptakS:
I was able to achieve something of these sorts:

image


Basically what I did was remove the `flex: 1` in the child elements which I guess you did so that each row takes up 100% space. And then to each of those elements, I added a `margin-left: -3px; margin-top: -3px` if the border width is 3px, to compensate for the double bordering. Then you can adjust the parent element's margin to compensate for the overall 3px shift to top and left. Also, the actual design being exactly same as the design would definitely depend on the order in which the categories are shown, which I think is fine. Also, you will need to add `justify-content: center` to the parent element.

chigby:
@SaptakS I put the finishing touches on this one and it's ready for review. Let me know what you think.

chigby:
Rebased this on the latest `category-symbol` branch. It's ready for review (and should be merged after that branch).

client/common/sass/components/_categories-block.sass
```diff
@@ -0,0 +1,22 @@
+.categories-block
+	display: flex
+	flex-flow: row wrap
+	justify-content: center
+	list-style: none
+	margin: 72px 5rem
+	padding: 0
+
+.categories-block--item
```
SaptakS: This should be `&--item` instead.

SaptakS (changes requested):
Only one small change. After that, this can be merged

client/common/sass/components/_categories-block.sass
chigby: Good catch here, I've added it.

SaptakS (approved):

PR #1202, opened 2022-01-19
Update npm dependencies to fix vulnerabilities.

SaptakS:
Updates follow-redirects. I was unable to update markdown-it, since the latest version of modernizr depends on a lower major version of markdown-it. So added it to ignore. I am not sure if we are actually using modernizr anywhere. It's not being used in our codebase it seems. In that case, we may want to completely remove modernizr dependency as well. @harrislapiroff any ideas?


chigby (approved):
Looks good to me

Issue #1210, opened 2022-01-26
Implement incident with the same event feature


SaptakS:
As mentioned in #1187 , we now have a new table for incidents with same event in the design for the incident page. This is a new feature and involves quite some additions in the backend since we don't have this data right now.

Deliverables:
- [ ] Implement the template styles as mentioned in [figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3805%3A61121)
- [ ] Create new model/model field to interconnect incidents belonging to the same event
- [ ] Sent context data from backend with a list of incident objects

harrislapiroff:
Confirmed with @KAMcCudden that this is a desirable feature and they currently struggle a lot with tying incidents together in the existing design.

harrislapiroff:
The original idea here was to have an "event" database object that editors manually tag incidents with, but @KAMcCudden suggested that maybe a better MVP would be to start by auto-detecting incidents as having the same event if they share these criteria:

* date
* location

A future implementation could have manual events (maybe smart events that would search for matching filters so they didn't have to be manually curated)

Jendunham122:
A few questions from the team:

How will this appear when feature is turned on? Will “incidents from the same event” appear above “related incidents” (or vice versa)? 

And if this ends up looking too crowded, will there be a way to make only one or the other appear?

harrislapiroff:
Here's a screenshot from the original mockups (we don't use the sources table, so that will never be there)

<img width="1437" height="1382" alt="Image" src="https://github.com/user-attachments/assets/34484a9e-707e-4daf-b9ff-0626130089c9" />

Issue #1209, opened 2022-01-26
Create sources table for incident page


SaptakS:
As mentioned in #1187 , we now have a new sources table in the design for the incident page. This is relatively a new feature compared to what we have right now.

Deliverables:
- [ ] Implement the template styles as mentioned in [figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3805%3A61121)
- [ ] Figure out which model element to use in the backend for the storing of source information
- [ ] Sent context data from backend with source informations (link to the source article, title of the source article, publisher of the source article)

Issue #1187, opened 2022-01-10
Incident detail page


harrislapiroff:
This ticket is for building and piecing together the components of the incident page _except_ for stream field blocks, which are specified in #1186.

You can see example templates incident pages in [Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3565%3A41473).

The components needed on the page include:

- [ ] Incident details table (new component, [Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3805%3A61121)) - @SaptakS - #1211
- [ ] Sources table (new component, [Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3805%3A61121)) - @SaptakS - #1211
- [ ] Incidents from same event table (new component, [Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3805%3A61121)) - @SaptakS - #1211
- [x] Page title (use component from [style guide](https://redesign.pressfreedomtracker.us/styleguide/#typography))
- [x] Tags (use component from [style guide](https://redesign.pressfreedomtracker.us/styleguide/#buttons))
- [ ] Introduction (new component, [Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3805%3A60362), may require new field)
- [ ] Cover photo/video (new component, [Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3805%3A72698))
- [ ] Related incidents (see example in [Figma template](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3565%3A41473))

SaptakS:
I have started working on the list table component as shown in the [Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3805%3A61121), that has similar styles for incident details, sources table and other tables in both incident detail page and blog detail page.

urlsangel:
[...]

PR #1211, opened 2022-01-26
List table components for incident page

SaptakS:
This PR started off aiming to implement the elements using the list table components, i.e. incident details section, sources table and incident with the same events table.

The scope of this PR has changed after a discussion to:
- [x] fully finish the implementation of incident details sections (both template styling and backend linking)
- [x] basic style implementation of sources table
- [x] basic style implement of incident with the same events table
- [x] Make everything responsive

Refs #1187 
Refs #1209 
Refs #1210


SaptakS:
There are few decisions I made, which I would love to hear if anyone has any strong feelings about:
- For incident details and category details I have used a `<dl><dt><dd></dl>` markup instead of a table since everything is a key/description kind of value.
- I have created separate functions and a dict with function maps to map to functions for each of the category fields and there I handle if needs any specific code, or just the generic function
- I am returning the entire HTML (with the link) for the values instead of just the value. This is because all the links also have some small intricacies which I wanted to address in backend itself.
- After discussion with @harrislapiroff I have updated the createctegories file to use the same slug as prod, so we have the same values that can be used in the category map.

SaptakS:
The things remaining in this PR are:
- [x] Check for responsiveness (mobile/tablet)
- [x] Make the CI pass (add tests, etc.)

SaptakS:
I still need to add tests for the new functions, but this should be ready for review. I will add the tests once we are sure that the way I have implemented is what we want.

styleguide/templates/styleguide/index.html
```diff
@@ -274,6 +274,39 @@ <h4 class="heading-regular">Blog Card (Small, Photo)</h4>
...
+						<dt class="details-table__label">
+							{{ detail.name }}
+						</dt>
+						<dd class="details-table__value">
+							{{ detail.html | safe }}
+						</dd>
+					</div>
+					{% endif %}
+				{% endfor %}
+			</dl>
```
chigby: Is this intended to render as an actual details item on the styleguide page? Because I'm not seeing anything locally (`category_detail` is not defined in the styleguide context).

styleguide/templates/styleguide/index.html
```diff
@@ -274,6 +274,39 @@ <h4 class="heading-regular">Blog Card (Small, Photo)</h4>
...
+		</pre></code>
+		<p>
+			All different card details have their own template include files.
+			For example, incident details are accessed like:
+			&lbrace;% include "incident/_details-table.html" with incident=page only %&rbrace;
```
chigby: Nitpick: the code fragrment here should be in code tag, probably.

incident/utils/category_field_values.py
```diff
@@ -0,0 +1,283 @@
...
+            getattr(page, field)
+        )
+        value = 'Yes'
+    else:
+        link = '{}?{}={}&categories={}'.format(
+            page.get_parent().get_url(),
+            field,
+            getattr(page, field),
+            main_category.title
+        )
```
chigby: There may be a nicer way to handle the querystring part of this (and similar) constructions. See [urlencode](https://docs.python.org/3.9/library/urllib.parse.html#urllib.parse.urlencode):

```python
from urllib.parse import urlencode

field = 'charge'
value = 1
category_title = 'Arrest/Criminal Charge'

urlencode({field: value, "categories": category_title})
# Returns 'charge=1&categories=Arrest%2FCriminal+Charge'
```

To me, setting up an explicit dictionary is slightly more readable, and we also get the benefit here of url-encoding an characters that need it.

incident/utils/category_field_values.py
```diff
@@ -0,0 +1,283 @@
...
+def detention_date_html_val(page, field):
+	return date_html_val(page, field)
+
+def release_date_html_val(page, field):
+	return date_html_val(page, field)
+
+def unnecessary_use_of_force_html_val(page, field):
+	# If no value for the attribute and not main category, return blank
+    main_category = page.get_main_category()
+    if not getattr(page, field) and main_category.slug != 'arrest-detention':
```
chigby: Inside this function feels a little bit to me like the wrong place to be checking if the main category is correct before continuing. It seems like we should be able to avoid calling this function at all if we are on an incident with the inappropriate category.

incident/utils/category_field_values.py
```diff
@@ -0,0 +1,283 @@
...
+            item.pk
+        )
+        value = item.title.capitalize()
+        val_list.append(
+            '<li class="details-table__list-item"><a href="{}" class="text-link">{}</a></li>'.format(
+            link,
+            value
+        ))
+
+    return '<ul class="details-table__list">{}</ul>'.format('\n'.join(val_list))
```
chigby: As the HTML involved in these becomes more complicated, I begin to wonder if we can gain some advantage by using the Django template system selectively to make what is going on more legible, while keeping the complicated logic, `getattr` calls, linkbuilding, and such, in the python. For example, what do you think of something like:

```python
def list_html_val(page, field):
    # If no value for the attribute, return blank
    items = getattr(page, field).all()
    if not items:
        return ''

    return render_to_string('incidents/_list_field.html', {
        'page': page,
        'field': field,
        'items': items,        
    })
```
With `incidents/_list_field.html`:

```django
<ul class="details-table__list">
  {% for item in items %}
    <li class="details-table__list-item">
      <a href="{% pageurl page.get_parent %}?{{ field }}={{ item.pk }}" class="text-link">{{ item.title|capfirst }}</a>
    </li>    
  {% endfor %}
</ul>
```

What do you think of this? To complicated? Better separation of concerns? I might slightly lean towards preferring this because most of what we are doing in these functions is assembling HTML code, and the templates feel like they're still pretty good at that so we should try to employ them where possible, but I can also see the appeal of a pure python approach.

incident/utils/category_field_values.py
chigby: I guess I'm also trying to avoid this hard-coded slug here, as well as avoiding calling `get_main_category` (which runs a DB query) more than once if we don't have to.

incident/models/incident_page.py
```diff
@@ -767,6 +770,21 @@ def get_main_category(self):
...
+            category_details[category.category] = []
+            for field in category_fields:
+                display_html = CAT_FIELD_VALUES[field[0]](self, field[0])
+                category_details[category.category].append(
+                    {
+                        'name': field[1],
+                        'html': display_html,
+                    }
+                )
+        return category_details
```
chigby, line 786: Regarding something I said in another comment, _this_ function seems like the place where, if we want to, we can skip details that don't belong to an incident's primary categories (or based on other logic). 

As I think about it, what I'm describing might involve adding more information to the data in `CATEGORY_FIELD_MAP`, like returning an object instead of a tuple, and then we could have `primary_only=True` or something on certain fields. Though, maybe that's complicating the matter too much.

incident/utils/category_field_values.py
chigby: A similar strategy could be employed on other functions here, particularly the ones that involve assembling lists.

chigby (commented):
Left some comments below, mostly focusing on the implementation of the relationship between categories and incident fields. As always, I definitely don't have all the answers here, these are just my thoughts & I'm interested in yours, too.

incident/utils/category_field_values.py
SaptakS: Nice idea. Totally agree. Would depend on whether we use template or python strings to get the html string as commented by you below.

incident/utils/category_field_values.py
SaptakS: So what I was trying to do here is [this](https://github.com/freedomofpress/pressfreedomtracker.us/blob/develop/incident/templates/incident/_arrest_rows.html#L105). It did make me squirm a bit too, but I wanted to preserve the current behaviour so I added it. The current behaviour seems to be if either it is main category or the value exists, then show the field in the table. So I just copied exactly the logic in the template because I think there were only 2-3 such cases.

incident/utils/category_field_values.py
SaptakS: Definitely agree that it would be much cleaner. I did think about this. I was just thinking if a `render_to_string` would further affect the performance of our backend. But I don't know if there would be any substantial degradation. What do you think?

home/templatetags/case_tags.py
```diff
@@ -0,0 +1,10 @@
+from django import template
+from incident.models import choices
+
+
+register = template.Library()
+
+
+@register.simple_tag
+def get_case_status_display(case_status):
+    return dict(choices.CASE_STATUS).get(case_status)
```
chigby: Duplicate of this? https://github.com/freedomofpress/pressfreedomtracker.us/blob/dbbc75a02e4bf2d358442fb882e6bdcf3f6ad89d/incident/templatetags/case_tags.py#L8-L10

SaptakS: Weird. I feel like the template was giving me error for missing the tags. But maybe I am wrong. Definitely a duplicate. Will remove it.

PR #1212, opened 2022-01-27
1187 incident detail page

urlsangel:
@harrislapiroff @SaptakS 

This branch has added the following component templates and fields where necessary:

 - Page title
 - Tags
 - Introduction
 - Cover photo/video
 - Related incidents
 
This should hopefully be separate to what you've been working on @SaptakS?

Part of #1187


chigby (changes requested):
So I think this is a good start on the incident detail page. As I go through it and compare it with the Figma design, I do notice some areas where I think additional work probably needs to be done for the implementation to match the design.

1. I am not sure if `page-title` is the correct class for the, well, page title. This converts the text to upper-case and doesn't seem to me to match the font size in the design.
2. Category icons should be in the page title. This is actually something I did some development on and I think it can be merged in later, so it doesn't need to be part of this PR.
3. The whitespace around the tag list needs some adjusting. [This section of the design](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3565%3A41043) has some more detailed specifications, but from what I can tell the left edge should be aligned with the header and the spacing above and below should match what's in the link.
4. In the mobile view, the details table does not match the design. I am not sure if this means that component needs to be adjusted fundamentally, or if we need to design a new component specifically for the incident details page. The "list table" style is used in a lot of places on the site and I don't think I have my head fully around each of them and the places where the design differs and where the design is the same. But it does seem to me that the mobile view shows the field and the value of each row on different lines:
image
Whereas, currently, ours has them on the same line (along with some differences in the line coloring, it looks like):
image
Again, I'm not fully sure of the tactic that needs to be taken to address this. Maybe @SaptakS or @harrislapiroff has some ideas.
5. It looks like we might need a little more responsive design added to the styling of the cover photo. For me, it overflows out of the page if the width goes under ~640px.
6. The font size on the introduction field feels too large to me on table and mobile layouts. The Figma design has it at 44px on tablet and 32px on mobile -- though I'm not sure what this mapst to in rem units.
7. The related incidents block probably needs some kind of layout or structure applied to it to constraint how the cards are shown, but that might be something better suited to another PR? It's possible that's something we want to make into a reusable component if the card pattern on this page is used elsewhere on the site.

These are mainly what stood out to me when comparing the Figma design to what's in this PR. I think all of these will eventually have to be addressed in some way, but not all of them necessarily ought to be done as part of this PR right now.

SaptakS:
> The "list table" style is used in a lot of places on the site and I don't think I have my head fully around each of them and the places where the design differs and where the design is the same. But it does seem to me that the mobile view shows the field and the value of each row on different lines:

This is weird. I remember making the design like in figma (i.e. in next line). Not sure what changed that makes it appear in the same line. But I think this list table is primarily being used in this section only, so I think ideally we can change the fundamental list detail component itself, and not need to create separate component since it's used only in incident details and category details as of now.

About the border color, it was a miss on my part. Apologies!

SaptakS:
Okay, after looking into it, point 4. is actually working below 376px width. This is because our mobile screens are defined below 376px. And based on the figma design, in tablets, it should still be side by side. So that is why somewhere around 400px, things got messy. Now for this particular issue, the easiest fix is adding `flex-wrap: wrap` which is what I will do and push to the branch. But I think we need to have a discussion regarding this screen width

SaptakS:
I have updated the border colors for tablets and below. And also added a `flex-wrap: wrap` so that even in width greater that tablet boundary, it wraps properly, instead of appearing messy

SaptakS:
(Working on the remaining tasks)

SaptakS:
@chigby I have addressed all your requested changes. Also made the layouts (media, body, related_incidents) match the figma design in all the different screen resolutions. The updates section was missing, have added that as well.

chigby:
I would say this is looking very good. As I was experimenting with the page, I did notice two things:

1. The lines on the details table, in desktop mode, don't extend all the way to the edge of the page. In the design they do.
2. The related incident cards hover effect is, similar to other components in the past, is causing the footer of the page to slide upward as well as the card.

This PR has been going for a while. @SaptakS let me know if you'd rather merge now and I can make tickets for these two points and someone can fix them later, or if you want to handle them prior to merging then that's also fine.

Also I think I know how to fix the coverage problem. I can add a commit for that.

SaptakS:
> The lines on the details table, in desktop mode, don't extend all the way to the edge of the page. In the design they do.

I did that purposefully. In the design, the screen size is restricted to the desktop contrain, which is 1440px. So I didn't extend them beyond that. It looked a bit weird to me extending full width, so interpreted it that way

> The related incident cards hover effect is, similar to other components in the past, is causing the footer of the page to slide upward as well as the card.

I am almost convinced that both this hover effect and the row hover effect on tables need to be done with some JS or some other way and not this simple way. I created an issue regarding that since that brings up some discussion points: #1248

chigby (approved):

PR #1225, opened 2022-02-03
Correct error in test assertion

chigby:
Fixes a date-formatting problem in a test assertion.

The only reason this ever passed is that it was written on the second day of the second month :sweat_smile:


harrislapiroff (approved):
Looks great

PR #1226, opened 2022-02-03
Add setting to toggle the django debug toolbar

### LINKED ISSUE #1213: Disable debug toolbar by default
@urlsangel observed that debug toolbar seems to be slowing down page loads on dev instances substantially. Most of the time developers probably don't need it running. Let's turn it off by default and put instructions for how to enable it (`tracker/settings/local.py`) in the readme.

chigby:
This pull request updates the version of the django debug toolbar, disables it by default, and provides a setting to turn it on locally.

Fixes #1213


SaptakS (approved):
Looks good to me

PR #1228, opened 2022-02-04
1218 Database page: Filters sidebar, basic

### LINKED ISSUE #1218: Database page: Filters sidebar, basic
Sidebar form for the database page. Use native HTML component instead of react/chart equivalents. Filters, for reference: [Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=4008%3A76112)

urlsangel:
Uses `serialized_filters` from `get_context` method of `IncidentIndexPage`  to convert filters to form fields.

Resolves #1218

### DIFF
--- a/incident/utils/forms.py
+++ b/incident/utils/forms.py
@@ -31,6 +31,12 @@ class FilterForm(forms.Form):
                 if _type == 'date':
                     field = forms.DateField
                     kwargs['widget'] = forms.DateInput
+                    kwargs['label'] = label.replace('between', 'before')
+                    self.fields[f'{name}_upper'] = field(**kwargs)
+                    kwargs['label'] = label.replace('between', 'after')
+                    self.fields[f'{name}_lower'] = field(**kwargs)
+                    field.field_type = _type
+                    continue
 
                 if _type == 'bool':
                     field = forms.ChoiceField
--- a/incident/utils/forms.py
+++ b/incident/utils/forms.py
@@ -37,6 +37,11 @@ class FilterForm(forms.Form):
                     kwargs['widget'] = forms.RadioSelect
                     kwargs['choices'] = [('Yes', 'Yes',), ('No', 'No',)]
 
+                if _type == 'checkbox':
+                    field = forms.MultipleChoiceField
+                    kwargs['widget'] = forms.CheckboxSelectMultiple
+                    kwargs['choices'] = [[x[0], x[1].capitalize()] for x in item.get('choices', [])]
+
                 if _type == 'choice':
                     field = forms.ChoiceField
                     kwargs['widget'] = forms.Select
@@ -54,7 +59,35 @@ class FilterForm(forms.Form):
 
 def get_filter_forms(request, serialized_filters):
     filter_forms = []
+    categories = []
+
     for item in serialized_filters:
+
+        # if the item has a filters object then create a form from the item
+        if item.get('filters', []):
+            filter_forms.append(
+                FilterForm(
+                    request.GET,
+                    data=item
+                )
+            )
+
+        # otherwise it's a category page, so add it to the list
+        else:
+            categories.append(item)
+
+    # if we have category items, create form filter for them
+    if categories:
+        item = {
+            'title': 'Limit to',
+            'filters': [{
+                'title': 'Category',
+                'type': 'checkbox',
+                'name': 'categories',
+                'choices': [[x.get('id', ''), x.get('title', '')] for x in categories]
+            }]
+
+        }
         filter_forms.append(
             FilterForm(
                 request.GET,
--- a/incident/utils/forms.py
+++ b/incident/utils/forms.py
@@ -1,0 +1,63 @@
+from django import forms
+
+
+class FilterForm(forms.Form):
+
+    def __init__(self, *args, **kwargs):
+
+        data = kwargs.pop('data', None)
+        filters = data.get('filters', None)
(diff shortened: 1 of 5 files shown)


harrislapiroff:
Copying over some chatter from Slack that may be relevant to someone working on this next week.

Russell:

> Filters of type radio don’t have any choices, I’m not sure where they would be available from?
> I’m not sure how to get the values back from GET to pre-populate  field values?

Cameron:

> Without Javascript I guess I'm not sure it's possible to pre-populate them?

Russell:

> The values will be in the GET request - if we used Django forms then it would be pretty much automatically handled
> I tend to build filters using Django forms as a base, then everything works sans JS, and can be progressively enhanced from there 

Harris:

> Having looked at the code, do you think there’s a good way to use django forms without duplicating the work that’s being done by the IncidentFilter? (Maybe IncidentFilter can be responsible for building a form?) It’s been a long time since I’ve looked at this code, so you two probably have a better sense of that?

Russell:

> > Maybe IncidentFilter can be responsible for building a form
>
> I would do exactly that

incident/utils/forms.py
```diff
@@ -0,0 +1,98 @@
...
+                    'required': False,
+                    'label': label,
+                }
+
+                if _type == 'text' or _type == 'autocomplete':
+                    field = forms.CharField
+
+                if _type == 'date':
+                    field = forms.DateField
+                    kwargs['widget'] = forms.DateInput
```
chigby, line 33: The thing about the date-type filter is it actually kind of wants to be represented as _two_ fields. The filtering backend is looking for two parameters: `<date_field>_upper` and `<date_field>_lower` (e.g. `detention_date_upper` and `detention_date_lower`). As it stands now, `<date_field>` by itself in the GET params will not do anything without also changing the filtering code.

And I actually don't think changing how this works is necessarily something we don't want to do, but that's perhaps a different conversation. In any case, is it possible to support two date fields per date filter here?

incident/utils/forms.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def get_filter_forms(request, serialized_filters):
+    filter_forms = []
+    categories = []
+
+    for item in serialized_filters:
+
+        # if the item has a filters object then create a form from the item
+        if item.get('filters', []):
```
chigby, line 77: As you mention in your comments on #1218, there is a need a "limit to a category" option, and your instinct is correct that these filters are not explicitly defined in the `filters` key, but what we have here is not quite the logic for determining that (and it's definitely not surprising you didn't infer this is how it works, there's not really documentation on the format of what's returned from `get_serialized_filters`, so no worries there). 

Here's what's happening: every item in `serialized_filters` "belongs to" a category _except_ the "general" filters item. The way to tell the difference is that each item has the `id` key set to the the primary key of the corresponding category page (which you are using to set the value of the appropriate checkbox below), and the general filters have `id` set to `-1`. The items with without filters (or with `filters == []`) are categories without specific filters that belong to them. 

So the code you've written here is very close to the correct behavior, but it will only create category checkboxes for certain categories, namely ones in the latter group without filters. There may be a more clever way to do this, but essentially the `categories` variable is every item in `serialized_filters` with an `id` key that's not `-1`, or, like, 

```python
[item for item in serialized_filters if item.get('id', -1) != -1]
```

Though how you want to express this I think is up to you. Let me know if that makes sense! The code that generates this is here: 

https://github.com/freedomofpress/pressfreedomtracker.us/blob/42bff62725af8b77b8cd220d5c5894b7c168541e/incident/utils/incident_filter.py#L529-L553

Which, though it's not the clearest, you can see the first part is a block of "general filters" and the second block loops through CategoryPages renders the filters that "belong" to each of those.

chigby (dismissed):
I'm generally pretty happy with the "Django forms" method that's being used here, though there are a couple of changes, noted in the code comments, I think ought to be resolved before this is merged. Let me know if you have further questions and I'll be happy to help or talk more about what's going on.

urlsangel:
> I'm generally pretty happy with the "Django forms" method that's being used here, though there are a couple of changes, noted in the code comments, I think ought to be resolved before this is merged. Let me know if you have further questions and I'll be happy to help or talk more about what's going on.

Hi @chigby 

Thanks for the review and the comments - both are very clear, and I have a plan on how to implement them.

@harrislapiroff - I'll get onto these changes when I'm back on Wednesday unless you direct me otherwise 👍

harrislapiroff (approved):
@urlsangel: Cameron made some changes to this PR. If you're able to review when you sign on today, that would be great. If Cameron's changes pass your review—or even if you just need to make small fixes—feel free to merge once you're done!

(If rebase and merge proves challenging, feel free to use your merging the `redesign` branch back in approach—or just mark as ready to merge and someone else will get it in the morning ET.)

incident/utils/forms.py
urlsangel: Looks good to me, pretty much how I would have written it I think 👍

incident/utils/forms.py
urlsangel: Makes sense now, and the way it's now implemented looks sensible 👍

urlsangel (commented):
@harrislapiroff @chigby 

The changes that @chigby added look good to me - it looks like I can't approve the PR though as I'm the author of the PR?

harrislapiroff:
I forgot it wouldn't allow merge until change requests were dismissed!

PR #1229, opened 2022-02-07
Blog index page redesign

### LINKED ISSUE #1214: Blog index page
See: [Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3698%3A70349)

SaptakS:
- [x] Implements page title
- [x] Summary section
- [x] Featured section
- [x] Latest blog section
- ~~[ ] Special section~~ (Decided to skip till later) https://github.com/freedomofpress/pressfreedomtracker.us/issues/1233
- ~~[ ] Filtered sections~~ (Decided to skip till later) https://github.com/freedomofpress/pressfreedomtracker.us/issues/1234
- [x] Backend modification for adding featured blogs 
- [x] Pagination and/or JS loading of more articles 
- [x] Make blog cards responsive for this particular page
- [x] Add tests

Refs #1214


SaptakS:
@harrislapiroff Few questions and discussions points I had while implementing the index page regarding the filters:
image

- As we can see above, when we have all posts, we have everything - blog articles, special section listings, newsletter articles. Now in the template everything looks very organized because exactly after 8 blog articles/newsletter articles (basically items with card layout), we have the special sections. But what if in the order, we have 7 card layout articles, and then special sections? Do we leave the position for the 8th empty?
- Right now there is no distinction between what is a blog article and what is a newsletter article. So we need to add some field for doing that, and some form of data migration/ manual selection to chose which articles are blog and which are newsletter for filtering purposes
- What will be there in Special Sections? Is this like the Topic Page? Because right now topic page is under incidents and not blogs.

harrislapiroff:
> But what if in the order, we have 7 card layout articles, and then special sections? Do we leave the position for the 8th empty?

Let's ask Accurat about intended behavior here. My instinct is we might want to use something like `grid-auto-flow: dense` to achieve this (or something similar—not sure what your implementation looks like here).

> Right now there is no distinction between what is a blog article and what is a newsletter article. So we need to add some field for doing that, and some form of data migration/ manual selection to chose which articles are blog and which are newsletter for filtering purposes

Yeah, we'll need to discuss whether newsletters and special sections are the same page type as blog posts (with a choice field to distinguish) or distinct page types. Maybe we can dig into this a bit during the check-in today. 

> What will be there in Special Sections? Is this like the Topic Page?

Special sections will replace topic pages, but I don't think we should attempt to migrate from old topic pages to new special sections. Let's discuss when we discuss the newsletter thing.

SaptakS:
`grid-auto-flow: dense` sounds like a reasonable idea to me. Only thing is then the order might not make sense.

blog/models.py
```diff
@@ -135,6 +147,15 @@ def get_meta_description(self):
         )
 
 
+class BlogIndexPageFeature(Orderable):
+    blog_index_page = ParentalKey('blog.BlogIndexPage', related_name='featured_blogs')
+    page = models.ForeignKey(Page, on_delete=models.CASCADE)
```
chigby: Would it be possible to make this a foreign key to `blog.BlogPage` directly? If so, I think we might be able to avoid the call to `specific` in `get_context`.

incident/models/export.py
```diff
@@ -52,6 +52,7 @@
     'index_entries',
     'subscribers',
     'wagtail_admin_comments',
+    'blogindexpagefeature',
```
chigby: This is something else that might not be needed if we set up blog page features as specifically linked to BlogPages only.

client/common/sass/_responsive.sass
```diff
@@ -17,6 +17,10 @@ $desktop-min: $tablet-max + 1px
 	@media(max-width: $desktop-min)
 		@content
 
+=mobile-down
```
chigby, line 20: This is maybe a larger question than this PR, and I'm not really asking for any changes here, but I have been kind of wondering about if we should retire using `=` and `+` for Sass mixins? This syntax has always been a little mysterious to me, probably due to lack of familiarity with our Sass codebase, but I think it's interesting that [the documentation](https://sass-lang.com/documentation/at-rules/mixin#indented-mixin-syntax) says "users are encouraged to avoid" defining and using mixins like this. 

Just a thought! Wondering if you have any opinions here.

blog/models.py
```diff
@@ -95,10 +108,9 @@ def get_context(self, request, *args, **kwargs):
         context['entries_page'] = entries
         context['paginator'] = paginator
 
-        if request.is_ajax():
-            context['layout_template'] = 'base.ajax.html'
-        else:
-            context['layout_template'] = 'base.html'
+        context['featured_blogs'] = [
+            f.page.specific for f in self.featured_blogs.all()
```
chigby: I think using `self.featured_blogs.select_related('page')...` here will reduce the number of queries if we know we're always going to want `.page` in there (which I think we do).

blog/templates/blog/blog_index_page.html
```diff
@@ -0,0 +1,25 @@
...
+{% block main %}
+
+	<h1 class="page-title">
+		Blog
+	</h1>
+
+	{% if page.body %}
+		<div class="site-summary site-summary--blog">
+			{% for block in page.body %}
+				{% include_block block %}
```
chigby, line 14: For me, the image block sort of breaks the layout if it's too wide for the viewport. Though that's potentially something that could be fixed elsewhere. Example:

image

client/common/sass/components/_blog-card.sass
```diff
@@ -5,9 +5,12 @@
 	&--info
 		margin-top: 12px
 
-	&--teaser
+	&__teaser
 		margin-top: 12px
 		margin-bottom: 12px
 
+		&--sm
```
chigby, line 14: Just wondering what "sm" refers to here, I'm not familiar with this notation.

client/common/js/articleScroller.js
```diff
@@ -0,0 +1,93 @@
...
+		this.articleParent.style.transform = 'translateX(' + shift + ')'
+	}
+
+	prevExists() {
+		return !ArticleScroller.isVisible(this.articles[0])
+	}
+
+	nextExists() {
+		return !ArticleScroller.isVisible(this.articles[this.articles.length - 1])
+	}
```
chigby, line 48: So I'm not sure if this is worth handling right now, but I did notice that these functions aren't run again if the window is resized, so if the window changes from where the first and/or last article is visible to where it's not visible, the corresponding button will remain inactive and the person won't be able to scroll. This might be an edge case, though, but I just wanted to make a note of that here.

chigby (changes requested):
So I think overall this is very well done and we are very close to merging. The only change I am really requesting is the `select_related` addition to the featured blogs query (as there's no reason not to have it), but I did leave some other feedback which might lead to other changes if others think they're possible or desired.

One other thing I did notice, and I wasn't really sure where to put it, is that the blog cards are displaying some slightly strange hover behavior for me where sometimes hovering on a card would cause other cards to shift upwards as well, but for other cards it did not. I'm attaching a GIF of what I'm seeing, maybe it will help troubleshoot it:
x

Anyway, let me know what you think of all this.

client/common/sass/_responsive.sass
SaptakS: I don't have any strong opinion on this. I also saw this in the documentation, but since it was already being used with `=` and `+` notation, I just kind of rolled with it.

blog/templates/blog/blog_index_page.html
SaptakS: I noticed this too. I am not sure about the design here since in the design it's always just a paragraph. I guess I can add `max-width: 100%` anyways to give some sane output.

client/common/sass/components/_blog-card.sass
SaptakS: So I just used `sm` as small similar to bootstrap notation. This is basically because I wanted the teaser to appear in smaller screens in blog listing page as mentioned in the blog index page design, but not show them in any of the other blog listing pages. I could have used a separate css, but this was the only change, so I went with it. This is related to https://github.com/freedomofpress/pressfreedomtracker.us/pull/1229/files/26ef4422272de8ac0725f1bca5aaa4805b5407d2#diff-de7a921470ee8b60b88019e99f2aaeb91b280301a6c29ae469f7bc06ae998478R51-R53 and https://github.com/freedomofpress/pressfreedomtracker.us/pull/1229/files/26ef4422272de8ac0725f1bca5aaa4805b5407d2#diff-94a3f3f742eaf4f8c35ee2cff00afb0dbba0e983977f2794cb8835bb92271e49R18-R20

client/common/js/articleScroller.js
SaptakS: Yes. I kind of didn't handle it because of the edge case. But since we don't expect a lot of users to keep resizing their window, I didn't put too much time in that. Let me know if it's necessary to be put in this PR for merging, or you think we can tackle this in a separate PR. Also, I know @harrislapiroff said that he had a similar implementation needed for homepage for horizontal scrolling.

chigby: I don't think it's needed for this PR, I just didn't want to forget about it

SaptakS: maybe we can create an issue and capture this before merging.

chigby (approved):

PR #1230, opened 2022-02-07
Redesign homepage, part 2

harrislapiroff:
Homepage items still pending are recorded in #1236


harrislapiroff:
Let's review this as is. I think it's already generating duplicate work with some other features (in particular #1229) so we should review and integrate and I'll open a new ticket for my remaining checklist items.

harrislapiroff:
This is ready for @chigby or @SaptakS to review—whoever can get to it first 😇

client/common/sass/components/_article-carousel.sass
```diff
@@ -0,0 +1,20 @@
...
+// The "Latest Incidents" carousel collapses to multiple rows
+// over smaller viewport sizes
+.article-carousel-item--latest
+	grid-column: 1 / -1 // span the entire grid
+
+	+tablet-up
+		grid-column: span 4 // ½ of grid
+
+	+desktop-up
+		grid-column: span 3 // ¼ of grid
```
SaptakS: This implementation is very close to [what I did with blog featured articles](https://github.com/freedomofpress/pressfreedomtracker.us/blob/redesign/client/common/sass/components/_blog-list.sass#L17-L62) using grid. I fell back on flex for the scrolling part, but since it's not part of this PR, I think this implementation aligns with my implementation. Once the scrolling part is implemented, I think we can safely swap up the classes from markup and everything should 90% work perfectly.

client/common/sass/components/_card-image.sass
```diff
@@ -1,8 +1,10 @@
 .card-image
 	&-big
+		max-width: 100%
```
SaptakS: I think this is repeated code from below, since I needed to add the same lines in blog index page 😅 Can be removed.

client/common/sass/components/_card-image.sass
```diff
@@ -1,8 +1,10 @@
...
 		display: block
 		max-width: 100%
 
 	&-small
+		max-width: 100%
```
SaptakS: Same as above.

home/models.py
```diff
@@ -174,6 +168,25 @@ def get_context(self, request, *args, **kwargs):
...
+    page = models.ForeignKey(Page, on_delete=models.CASCADE, related_name='+')
+
+    panels = [
+        PageChooserPanel('page', ('incident.IncidentPage',)),
+    ]
+
+
+class FeaturedBlogPost(Orderable):
+    home_page = ParentalKey('home.HomePage', related_name='featured_blog_posts')
+    page = models.ForeignKey(Page, on_delete=models.CASCADE, related_name='+')
```
SaptakS: Something that @chigby pointed out to me in my blog index page PR and I think applies here as well is we can just use `blog.BlogPage` here itself and not mention in PageChooserPanel. Same applies to `FeaturedIncident` above.

SaptakS (changes requested):
This PR looks ready to be merged. Some small changes are needed, which I will finish off and then merge this PR.

SaptakS:
I see that incident cards have a "Updated on" label all the time, but there's no value as of now. I think we need to update how we show "Updated on" as either update date or the creation date. Or else just not show that label if update date is not available.

But this is not important for completion of this PR, just wanted to note it here.

SaptakS (approved):

Issue #1219, opened 2022-01-31
Database page: Header and Layout


harrislapiroff:
Pull together components from #1217 and #1218

Filters do not need to be functional for completion of this ticket.

harrislapiroff:
This is done

PR #1231, opened 2022-02-07
Horizontal incident cards

### LINKED ISSUE #1217: Database page: Horizontal incident cards
- [x] Template for horizontal incident cards ([Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=2710%3A22015) "Incident Card, Database Card")
- [x] Template for list of such cards

chigby:
Resolves #1217 

This PR adds horizontal incident database cards as specified in the issue. Some thoughts:

1. I've tried to get the spacing on the lines to match the design as much as possible, but it's going to depend on the exact content.
2. A lot of the design for this card is similar to, but does not exactly match, the details table component. Rather than try to abstract out those styles and markup, I've basically duplicated that code and changed what needed to be changed. I don't really have a problem with this strategy but if there's a more "elegant" way of doing it, I'd be interested to learn it.
3. That said, I did make some changes to the details table prior to duplicating that code, cleaning up some rough edges, basically.
4. I found that our definition of what a "desktop" view might be a little different than what the design shows. We consider anything viewport with width greater than 768px to be a desktop. This is probably fine (though according to Firefox an iPad has a width of 810px?) but the desktop design for these cards has a width of roughly 1050 pixels, and as you shrink down to 800 or 900 pixels it starts to look a little cramped, so I made an executive decision to modify the layout and "stack" the main incident details and the category details tables if there's not enough room to show them comfortably side-by-side. It looks a little like this:
image
5. It's not clear to me if these cards as a whole are supposed to be links. If so, I'll need to modify them a little bit to reflect that.

### DIFF
--- a/client/common/sass/common.sass
+++ b/client/common/sass/common.sass
@@ -32,3 +32,5 @@
 @import components/media
 @import components/quote
 @import components/aside
+@import components/incident-index
+@import components/incident-database-card
--- a/client/common/sass/components/_details-table.sass
+++ b/client/common/sass/components/_details-table.sass
@@ -1,6 +1,5 @@
 .details-table
 	position: relative
-	+constrain-to-desktop
 
 	&__incident-details
 		margin-top: 0
--- a/client/common/sass/components/_incident-database-card.sass
+++ b/client/common/sass/components/_incident-database-card.sass
@@ -29,6 +29,10 @@
 		margin-top: 0
 		margin-bottom: 0
 
+	&__category-details
+		margin-top: 0
+		margin-bottom: 0
+
 	&__row
 		padding: 0.5rem 0
 		border-bottom: 1px solid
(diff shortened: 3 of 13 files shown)


SaptakS:
(This PR is under review)

SaptakS (approved):
I made some adjustments in the flex layout to match the gaping more with the figma design. And made the flex-direction: column for tablet and below. I think as discussed we will anyways adjust the responsive sizes so should be fine. 

Based on the design, i would think that only the title should be clickable to take to the particular incident, though the focus style makes me think they want entire incident card to be focused? But I think this discussion can be had as part of the #1219 . Based on the principle of merging quickly, I think this is good to be merged.

SaptakS:
I am also ignoring the npm_audit fail right now, since that needs to be updated in dev and then rebased and might be a little time consuming while not really hurting if we merge.

PR #1249, opened 2022-02-22
Database page layout

chigby:
#1219 

Initial draft PR

[Scroll behavior](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3897%3A59348)

### DIFF
--- a/client/common/js/details-table.js
+++ b/client/common/js/details-table.js
@@ -1,12 +1,12 @@
-document.querySelectorAll('.incident-database-card__details-toggle').forEach(button => {
+document.querySelectorAll('.incident-database-card__details-toggle').forEach((button) => {
 	button.addEventListener('click', (e) => {
-		let isVisible = Boolean(e.target.dataset.visible === 'true')
-		let newVisible = !isVisible
-		let newText = newVisible ? "Show Less" : "Show More"
+		const isVisible = Boolean(e.target.dataset.visible === 'true')
+		const newVisible = !isVisible
+		const newText = newVisible ? 'Show Less' : 'Show More'
 		e.target.innerText = newText
 		e.target.dataset.visible = newVisible.toString()
 		e.target.setAttribute('aria-expanded', newVisible.toString())
-		e.target.getAttribute('aria-controls').split(' ').forEach(id => {
+		e.target.getAttribute('aria-controls').split(' ').forEach((id) => {
 			document.getElementById(id).dataset.visible = newVisible.toString()
 		})
 	})
--- a/client/common/js/details-table.js
+++ b/client/common/js/details-table.js
@@ -1,20 +1,13 @@
-if (document.querySelector('#details-toggle-btn')) {
-	const detailsBtn = document.querySelector('#details-toggle-btn')
-	const detailsDiv = document.querySelector('#incident-details')
-
-	detailsDiv.dataset.visible = 'false'
-	detailsBtn.setAttribute('aria-expanded', 'false')
-	detailsBtn.classList.add('has-js')
-
-	detailsBtn.addEventListener('click', () => {
-		if (detailsDiv.dataset.visible === 'false') {
-			detailsDiv.dataset.visible = 'true'
-			detailsBtn.setAttribute('aria-expanded', 'true')
-			detailsBtn.innerText = 'Show Less'
-		} else if (detailsDiv.dataset.visible === 'true') {
-			detailsDiv.dataset.visible = 'false'
-			detailsBtn.setAttribute('aria-expanded', 'false')
-			detailsBtn.innerText = 'Show More'
-		}
+document.querySelectorAll('.incident-database-card__details-toggle').forEach(button => {
+	button.addEventListener('click', (e) => {
+		let isVisible = Boolean(e.target.dataset.visible === 'true')
+		let newVisible = !isVisible
+		let newText = newVisible ? "Show Less" : "Show More"
+		e.target.innerText = newText
+		e.target.dataset.visible = newVisible.toString()
+		e.target.setAttribute('aria-expanded', newVisible.toString())
+		e.target.getAttribute('aria-controls').split(' ').forEach(id => {
+			document.getElementById(id).dataset.visible = newVisible.toString()
+		})
 	})
-}
+})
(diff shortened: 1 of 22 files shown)


chigby:
More on scroll behavior. Here's what's in the Figma:

image

image

I am able to get the sidebar and results section to scroll separately by adding this CSS to each container:

```css
maximum-height: 100vh;
overflow-y: scroll;
```

There may be another way to do this that I don't know about, though. But having added that, what I'm not sure how to achieve is the "applied once the beginning of the filters section reaches the top of the screen." I've thought about how to do this with CSS only and I am unable to come up with a way. With JS, I am thinking it might be possible to add an event to `onscroll` that checks if the results section's Y position on the page has reached the top, and _then_ apply the above CSS (or similar CSS) to both containers, but this leads me to consider edge cases like: what if they scroll significantly beyond the top of the filters box? What if they want to scroll the page back up to above the filters sidebar? So I haven't messed with it that much and wanted to consult with others if they might have other ideas.

harrislapiroff:
@chigby I've just taken a cursory look, so I'm not 100% sure on this, but have you played with `position: sticky` on the filters bar?

chigby:
I haven't, thanks for the suggestion.

chigby:
Poked around a bit with `position: sticky` and I have some thoughts. This property is definitely close, or at least near to, what we want, but the behavior of various elements on the page when I apply sticky positioning to has me further questioning exactly what the design is describing. Is the design showing us something where if you scroll down to a certain point you are not permitted to scroll the entire page further, like, if you wanted to scroll to the footer on the DB page, are you allowed to? If you're not allowed to, I'm not sure position sticky will work, since "sticking" an element means (as I understand it) sticking it with in a container, and there's nothing stopping someone from scrolling past whatever contains the filters & search results (and since right now these two containers are already at the "bottom" of the main container, the stickiness is not apparent in that case). I'm thinking one way around this might be to stick the elements to `<body>`, though I haven't toyed with this enough to figure out of that's possible, or if it is how to do that, since these elements are not immediate children of `<body>`. Though then you're running into the problem of these elements sticking over-top of the footer.

Which is similar to answering the earlier question as "no, you are _not_ permitted to scroll past the filters/results." In which case maybe the answer is not positioning but simply to make them the last actual element on the page?

It could be I'm missing something here, though.

chigby:
Q for @harrislapiroff (or anyone with opinions) regarding the HTML version of the filter sidebar. Right now, the grouping of the fields on the filter sidebar does not really match the hierarchy defined in the Figma design, which, as I read it, is to give each "General" filter (date of incident, venue, tags, etc.) its own section. The current design on this branch is that "General" is a section that contains each of these filters. This can be fairly straightforwardly changed to match the specification in Figma.

What is less straightforward to me is how to implement the category-fields as a pure HTML form. Here's the design:

image

Is our goal for the HTML form to match this design as exactly as possible, with the clickable category headers opening a second level of nesting with the field names, which open out to the form fields themselves? The part that is not doable without JS is the dynamic count updates (and, for now, category level counts I don't think are available at all). But it seems to me we are probably able to get _closer_ to the design than we are right now, though without an exact specification for the pure HTML form it's not clear to me how exact we want to try and go.

harrislapiroff:
Let's not worry about matching the pure HTML layout to the design. The vast majority of people will be seeing the javascript experience, so that's where I'd like to focus. (Tomorrow, if I look at the chart code and decide it's too complicated for us to implement in a few weeks, I might revise this, but for now, let's merge without fussing.)

client/common/js/details-table.js
```diff
@@ -1,20 +1,13 @@
...
-	detailsBtn.addEventListener('click', () => {
-		if (detailsDiv.dataset.visible === 'false') {
-			detailsDiv.dataset.visible = 'true'
-			detailsBtn.setAttribute('aria-expanded', 'true')
-			detailsBtn.innerText = 'Show Less'
-		} else if (detailsDiv.dataset.visible === 'true') {
-			detailsDiv.dataset.visible = 'false'
-			detailsBtn.setAttribute('aria-expanded', 'false')
-			detailsBtn.innerText = 'Show More'
-		}
```
SaptakS, line 18: Removing this code makes the individual incident pages not function properly in smaller screens.I guess we can keep both the sections?

client/common/sass/_typography.sass
```diff
@@ -48,3 +48,28 @@ body
...
+	.rich-text &
+		font-size: inherit
+		line-height: inherit
+
+	&:not(.btn):focus
+		border: 3px solid
+		border-color: $black
+
+	&:not(.btn):not(.text-link--card):not(.incident-index__download-item):hover
+		background-color: $accent-color
```
SaptakS: I removed this generic `p` and `a` styles because they were conflicting too much. Wherever the common link style is needed, we can add `text-link` class. Is there any other reason to bring this back?

client/common/sass/components/_incident-index.sass
```diff
@@ -1,2 +1,332 @@
...
+.incident-index__results
+	grid-column: 1 / -1
+
+	+tablet-up
+		grid-column: 4 / -1
+		display: grid
+		+grid-columns(9, 2rem)
+		position: sticky
+		top: 0
+		overflow-y: scroll
```
SaptakS: I think the sticky and scroll logic here needs some adjusting based on the figma design. I don't think the entire result section needs to be sticky, but the heading of the result, and the count, sort, etc are what's sticky based on [this figma presentation](https://www.figma.com/proto/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=4474%3A116770&scaling=min-zoom&page-id=3897%3A59348&starting-point-node-id=3897%3A59349)

SaptakS (changes requested):
Mostly looks good to me. Left a few comments. The sidebar sticky looks great to me, though I feel the height can be full 100vh instead of `100vh-120px`. Any specific reason for that? For the result section, I think the sticky logic needs som adjusting maybe to match the scrolling example in figma presentation

client/common/sass/_typography.sass
chigby: Ohh, I think this got re-added during rebase because I changed a line somewhere in it. I'll take it back out.

SaptakS (approved):
Looks good to me

PR #1250, opened 2022-02-23
1240 Contact form

### LINKED ISSUE #1240: Contact form/Submit an incident form
[Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3689%3A60388)

Update the `FormPage` model, templates, and styles to match the Figma design.

urlsangel:
Adds grouped form fields to `FormPage` model, adds selectable templates for differing output styles, adds form page and landing templates.


client/common/sass/_a11y.sass
```diff
@@ -0,0 +1,8 @@
+=a11yhide
+    position: absolute !important
+    clip: rect(1px, 1px, 1px, 1px)
+    padding: 0px
+    border: 0 !important
+    height: 1px !important
+    width: 1px !important
+    overflow: hidden
```
chigby: I'm wondering if this serves a similar enough function to the [`sr-only` class](https://github.com/freedomofpress/pressfreedomtracker.us/blob/52ecaf85702982b9f115cb136fb21c95295f8e46/client/common/sass/components/_sr-only.sass#L1-L8) that these should be reconciled or combinedf somehow, so as not to have two implementations of the same general idea?

forms/templates/forms/form_page.html
```diff
@@ -0,0 +1,41 @@
...
+            {% csrf_token %}
+            {% for group in page.groups %}
+                <fieldset class="fieldset fieldset--spaced">
+                    <h4 class="heading-regular">{{ group.title }}</h4>
+                    {% if group.description %}
+                        <p class="hint">{{ group.description }}</p>
+                    {% endif %}
+                    {% with template='forms/_'|add:group.template|add:'.html' %}
+                        {% include template %}
+                    {% endwith %}
```
chigby, line 32: Tell me a little more about your idea here for the `template` data on the `FieldGroup` model. I am wondering if having these pre-defined templates makes form customization a little too inflexible. I think that having field groups is great (and think that nested inlines are great for this purpose). But, and please correct me here if I'm misunderstanding how this is supposed to work, is there something about the design or structure of the form overall that prevents form groups from being more generic container with titles and descriptions that are capable of rendering an arbitrary collection of form fields?

Because from reading the code and playing around a little bit locally, it seems like if I create a group with the "Choices" template and add a URL field to it, the field will not be shown. This particular combination of fields (that is, a choices field and a date field) is not in the design, but I was imagining that if someone _did_ want to make a group like that, it would be possible without adding a new template. If there is a technical limitation preventing this, I'm curious to know what it is.

In general, I do like the idea of giving site editors greater ability to customize the appearance of forms, but I'm feeling like, as implemented here, the templates are in fact restricting what's possible in terms of form creation. Let me know if I'm misunderstanding something, though, as I've definitely had that happen before!

forms/models.py
```diff
@@ -62,20 +73,68 @@ class Meta(AbstractFormField.Meta):
     )
 
     panels = AbstractFormField.panels + [
+        FieldPanel('placeholder'),
         FieldPanel('append_to_subject'),
         FieldPanel('use_as_reply_to'),
     ]
 
 
+class FieldGroup(ClusterableModel, Orderable):
```
chigby, line 83: In spite of this being an `Orderable`, I'm having a hard time reordering these, locally. Or at least getting the order to stick? If I change the order in the admin, then publish, the order on the rendered page doesn't seem to change (and reverts back to its previous state in the admin). Are you seeing this? I'm not sure what's going on, actually. Might be something weird on my end?

chigby (commented):
Left some comments about a couple issues I experienced in my local testing, as well as a bigger question on the architecture. I don't want to say I'm specifically requesting changes in that area, but I'm not sure it should be merged as-is either. Thanks for bearing with me.

client/common/sass/_a11y.sass
urlsangel: @chigby 

My bad - I didn't realise there was already a class for this purpose, my addition can be replaced with the existing `sr-only` class 👍

forms/templates/forms/form_page.html
urlsangel: @chigby 

I implemented this based on the design - I couldn't see any other way to implement the date without a dedicated template (how would the fields know that they need a forward slash between them/need to be on a single line etc.?).

I separated the templates out based on the design, but I think it's only the date that really needs its own template, the others could use a default loop as you've said.

I was working to the figma design - it would be good to have a design that shows all of the possible implementations (such as select etc.), or these could be removed as options from the edit field if they are not supported?

forms/models.py
urlsangel: @chigby 

I think I've seen this before - I've added an explicit `sort_order` to the model's Meta ordering, and this looks to have solved the issue.

forms/templates/forms/form_page.html
chigby: Definitely get what you're saying about matching the design. I'm also a little bit stumped as to how to implement the three year/month/day fields without some kind of workaround. Hmm. So, if I understand correctly, the way to set up a date field that matches the design would be to create a group with the correct template and add three number fields to it, _and_ this type of date field can only exist in a group by itself. 

I mean, I guess the design is the design, but it seems like no matter how we go about this, it's somewhat of a regression in terms of effort and flexibility on the part of the site editor.

urlsangel: @chigby 

> So, if I understand correctly, the way to set up a date field that matches the design would be to create a group with the correct template and add three number fields to it, and this type of date field can only exist in a group by itself.

That's the way it currently works, yes.

> I mean, I guess the design is the design, but it seems like no matter how we go about this, it's somewhat of a regression in terms of effort and flexibility on the part of the site editor.

I completely agree - the design doesn't really take into account a content-managed contact form, it's very specific in terms of layout. So I think either the design needs to be amended to work with the grain of a CMS, or a more involved solution is required (which is where we are at currently).

I've pushed some updates today which simplify the available group templates to be either `Default`, which will print out every type of field in order, or `Single date`, which outputs fields inline with a `/` divider. I've also switched to the `sr-only` class for labels, and removed the duplicate class.

That's all I can do for now - I'm away now until I get back to the UK in a few weeks. Hopefully this is now generic _and_ specific enough for the project 🙂

chigby:
I think the latest batch of changes look good here, and the "Default" template behaves basically how the old form page template behaved, which is great. 

I am having a little trouble using the single-line date template to get the exact design as shown on Figma. This is the closest I was able to get:

image

whereas the Figma looks like this:

image

Maybe I don't have it set up quite right?

urlsangel:
@chigby 

> Maybe I don't have it set up quite right?

My mistake (I’m away from laptop, so can’t fix this currently) - the `label` element in the date template should have the class `form-label—hidden` removed and `sr-only` added.

chigby:
Looks great to me, I think this can be merged once the checks are passing. The safety check looks like it can be fixed by rebasing or merging the latest `redesign` branch, and I think the dev check is failing due to a migration conflict.

PR #1258, opened 2022-03-02
Adds skiplink and title id to all h1

SaptakS:
For testing, load any of the pages, press tab, and "Skip to content" button should appear. Unless tabbed, it stays hidden. When tabbed, it's the first focusable item. All the "Skip to content" buttons link to `#title`. I have added `id="title"` in all the `<h1>` tags in the page templates. The purpose of this link is to skip through the links in the menu and jump to the content. This is a [WCAG AA requirement](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-skip.html)

### DIFF
--- a/blog/templates/blog/blog_index_page.html
+++ b/blog/templates/blog/blog_index_page.html
@@ -4,7 +4,7 @@
 
 {% block main %}
 
-	<h1 class="page-title">
+	<h1 class="page-title" id="title">
 		Blog
 	</h1>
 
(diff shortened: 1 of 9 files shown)


chigby (approved):
Great change! It would be nice if there were a systematic way (like a part of the test suite) to verify that all our pages had a page element with `id="title"`. I actually have been thinking of a way to do this but it might have to wait until we get a few more of the templates implemented.

Anyway, I'm going to merge this.

PR #1261, opened 2022-03-07
Homepage 3

### LINKED ISSUE #1236: Homepage, part 3
- [x] Deprecate outdated fields and remove them from admin editor
- [x] Add text field for categories block
- [x] Add mobile styles
- [x] Add new components to styleguide
- [x] Add graph component (#1215)
- [x] Unify article carousel styles with blog list styles

chigby:
This pull request reconciles the blog and incident carousel by adapting the home page styles/code to match that on the blog index page. I created a slightly more general purpose "article carousel" component template. It might be more generalizable than what I've done but I'm not convinced that's needed. I've also hit the other points on the issue except for the charts, as discussed.

Fixes  #1236


home/templates/home/_featured_incidents_section.html
```diff
@@ -1,21 +1,15 @@
 {% load wagtailcore_tags %}
 
-<h2 class="homepage-section-header">Featured Incidents</h2>
+<h2 class="homepage-section-header">{{ label }}</h2>
```
SaptakS: In this template and similar template files, can we have a default value if label is not mentioned? As in {{ label|default:"Featured Incidents" }} ?

chigby: Good idea. I have added them.

SaptakS (approved):

PR #1262, opened 2022-03-08
Add link handler to apply category symbols to Category Page links

chigby:
This PR adds a link handler to wagtail's rich text processing set of hooks to apply our special category symbol CSS classes when it detects that the link is being made to a `CategoryPage`.


SaptakS (approved):
Looks good to me.

PR #1268, opened 2022-03-11
Fixes recent incidents regression in style

SaptakS:
While merging of the article scroller logic in the #1261 , some of the style codes for the latest incidents were lost, resulting in the recent incidents not appearing correctly. This PR fixes that.


SaptakS:
Self merging since it's relatively small.

Issue #1257, opened 2022-03-02
Information on incident cards look cramped


SaptakS:
![image](https://user-images.githubusercontent.com/9530293/156394241-27a2a190-d0be-415d-ad96-621ac538aa96.png)

As seen above, in smaller laptop screen (checked in 1280px width), the information on the incident cards look little cramped and wrapped. This is mainly in a 4 column grid. I don't know an exact solution for this. I wish container queries start getting adopted by all browsers very soon. This would be so much easier to address.

harrislapiroff:
I think the font-size here might be a little larger than we're seeing in the mockups and that might make it work a little better. I'll look into this.

chigby:
I wonder if part of the problem here is that the data rows are supposed to be aligned like that? Here's the home page from the Figma templates:

![image](https://user-images.githubusercontent.com/561931/157966963-86d1ee39-f7c2-4d60-92cd-e43f1ede7d09.png)

Note how the data table is not rendered in aligned columns, in contrast to @SaptakS's screenshot. Maybe that is contributing to the cramped feeling?

PR #1270, opened 2022-03-11
Incident card details alignment

chigby:
Refs #1257 

Addresses potential cramped text and other discrepancies from the spec on the "Incident Card" component. Basically I have reformatted the incident details so they match the "database card" component. This looks more like the [Figma design](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3897%3A60123) in terms of alignment and font size, and as a result looks less cramped. Though I don't know if it fixes all the issues regarding the overall presentation of these cards raised in issue #1257

### DIFF
--- a/client/common/sass/components/_incident-database-card.sass
+++ b/client/common/sass/components/_incident-database-card.sass
@@ -107,7 +107,6 @@
 	&__value
 		margin-left: 0
 		font-size: $font-size-small-p
-		white-space: nowrap
 		line-height: 1.43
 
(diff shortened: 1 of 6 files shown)


SaptakS:
I think this looks much better and I think for the time this according to me can be considered resolving  #1257 fully (at least the intention with which I created the issue)

SaptakS (approved):

PR #1271, opened 2022-03-14
Category page

SaptakS:
This PR adds template for the category pages. Few things that I have not done that are part of the design:
- I have not added links to the statistics items, since we don't have a system in the admin for the statistic items to create a link from the filters added (correct me @chigby if I am wrong)
- The introduction layout looks a little different because I have not added any chart to the template. Once that is added, the template should look the same as the design. I am thinking that will be a separate PR though.

Everything else should be working. I have reused the featured incident, recent incident and featured blogs template from the home page, since I think they are exactly the same.


common/models/pages.py
```diff
@@ -224,10 +231,17 @@ class StatisticsItem(Orderable):
         blank=True,
         help_text='Whitespace-separated list of arguments to be passed to the statistics function',
     )
+    link = models.CharField(
```
chigby: Would it be possible to make this a URLField, for extra validation?

common/models/pages.py
```diff
@@ -261,8 +275,18 @@ def clean(self):
             })
 
 
+class CategoryMethodologyItem(Orderable):
+    page = ParentalKey('common.CategoryPage', related_name='methodology_items')
+    label = models.CharField(max_length=255)
+    description = models.CharField(max_length=255)
```
chigby: Do you think 255 characters might be too limiting for this description field? In the design mockups it didn't seem like anything was over this limit, but I could also imagine someone wanting to write more and it wouldn't be too hard to change it to a `TextField` here and let users write as much as they want. Also, `TextField` would be rendered as a textarea element in the admin which I think might be nicer for editing short paragraphs of text, which seems to be what this field is intended to hold. Just an idea, though.

chigby (approved):
I'm approving this to expedite merging, but I did leave two comments that might be "nice to have" but definitely are not blockers. This page is looking really good and the functionality is working exactly as expected.

SaptakS:
Updated the model fields as commented @chigby

PR #1277, opened 2022-03-15
Filter autocomplete fields and other improvements

chigby:
This pull request:

1. Updates the filter sidebar on the database page to use `<intput>` and `<datalist>` elements with for autocomplete filters.
2. Updates the cursor to become a pointer when hovering over the expandable filter headers 
3. Adds a number of autocompletable items to the `createdevdata` command, so that we can more easily test that functionality locally.
4. Fixes a bug where boolean fields submitted from the filter sidebar would not actually work.
5. Adds the search text to the search results header (as shown in the Figma design).

Originally this PR used `<select>` elements for autocomplete filters, and I think the intput/datalist is nicer, but either one is an enhancement over what is on the redesign branch now. I think this should be merged and we should discuss further improvements to this and see what is feasible for launch.

Note: with the datalist elements, we expect the page size to be larger. For me, locally, it is 177kB with the following counts:

| Item | Count |
| ----- | ----- |
| States | 57 |
| Targeted Journalists | 139 |
| Tags | 925 |
| Venues | 92 |
| Law-enforcement organizations | 113 |
| Charges | 153 |
| Nationalities | 89 |
| Politicians/public figures | 225 |
| Equipment | 14 |

### DIFF
--- a/incident/devdata.py
+++ b/incident/devdata.py
@@ -4,6 +4,7 @@ import factory
 import random
 import wagtail_factories
 from wagtail.core.rich_text import RichText
+from faker.providers import BaseProvider
 
 from incident.models import (
     Charge,
@@ -26,6 +27,7 @@ from incident.models import (
     TargetedJournalist,
     GovernmentWorker,
     TopicPage,
+    Venue,
 )
 from common.models import CustomImage
 from common.tests.factories import (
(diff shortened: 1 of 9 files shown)


incident/devdata.py
```diff
@@ -40,6 +42,82 @@
...
+            'encryption',
+            'calumny',
+            'plagiarism',
+            'forgery',
+            'bootlegging',
+            'misdeeds',
+            'gambling',
+            'espionage',
+            'problematization',
+        ])
```
harrislapiroff, line 85: This is amazing

SaptakS (approved):

PR #1283, opened 2022-03-21
Add "Learn More Page" setting

chigby:
This pull request adds a setting to the Search Settings that contains a reference to the page that should be linked to by the "Learn More" option on the database page download data links, as seen here:

image

If the setting is left blank, that link will not be rendered at all.

This PR is based on #1277 and should be reviewed and merged after that one.


incident/templates/incident/_filters.html
```diff
@@ -2,7 +2,7 @@
 	<header class="filters__header">
 		Filters
 
-		<a href="{{ page.url }}">Clear All</a>
+		<a class="text-link" href="{{ page.url }}">Clear All</a>
```
SaptakS, line 5: Maybe let's change this from a link to `<input type="reset" value="Clear All" />`. That will avoid the reloading of the page which created some confusion. Though not necessarily need to be part of this PR.

PR #1285, opened 2022-03-21
CI cleanup

chigby:
This PR changes some of our configuration and code so that CI will pass. The changes are:

1. Fix a small flake8 error.
2. Tell eslint to ignore the `charts` directory javascript. This is a known problem and we don't need to see 200+ problems every time it's run.
3. Ignores the latest [`minimist` vulnerability](https://github.com/advisories/GHSA-xvch-5gv4-984h). See commit message for further elaboration but the short answer is I don't think it's going to go away soon, given how many dependencies of ours it affects.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,7 +55,7 @@ jobs:
           command: |
             mkdir -p test-results
             npm ci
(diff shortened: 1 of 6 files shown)


harrislapiroff (approved):
Looks good!

PR #1292, opened 2022-03-24
Update rich text rendering

### LINKED ISSUE #1276: Redesign: org links at bottom of page
Hi all! Many congrats on the redesign. Just a small note on something that perhaps you're already aware of: in this text at the bottom of the page - "U.S. Press Freedom Tracker is led by the [Freedom of the Press Foundation](https://freedom.press/) and the [Committee to Protect Journalists](https://www.cpj.org/), in collaboration with leading press freedom groups" - the org names are black, which is invisible on the black background. If you hover over the links, you can see the names of the orgs, but it might be good if they were visible without hovering.

### LINKED ISSUE #1288: Footer fixes

chigby:
Fixes #1288
Fixes #1276

This pull request:

1. Adds a fair amount of rich-text related content to the dev data, to make it easier to preview exactly how various parts of the site that use rich text will render links.
2. Removes the Wagtail's "legacy rich text" setting from our site, meaning the `richtext` filter will no longer wrap its content in `<div class="rich-text">`
3. Adds that div back in manually, where it can be done and the styles do not need to change.
4. Adds field-specific CSS for parts of the site where rich text requires different styling, namely blog page AsideBlock content and the footer (both of these sections have a dark background color).

See commit messages for more tech details.

### DIFF
--- a/blog/tests/factories.py
+++ b/blog/tests/factories.py
@@ -56,6 +56,7 @@ class BlogPageFactory(wagtail_factories.PageFactory):
                     'styled_text',
                     'blockquote',
                     'styled_text',
+                    'aside',
                     'list',
                     'styled_text',
                 ],
--- a/client/common/sass/components/_aside.sass
+++ b/client/common/sass/components/_aside.sass
@@ -6,8 +6,12 @@
 		*
 			color: inherit
 
+		a
+			color: $white
+
 		a:focus
 			border-color: $white
 
 		a:hover
+			background-color: $accent-color
 			color: $black
--- a/client/common/sass/components/_citation.sass
+++ b/client/common/sass/components/_citation.sass
@@ -1,7 +1,13 @@
 .citation
+	border-top: 5px solid
+	border-color: $black
+
+	&--inner
+		padding: 32px 0
+		+constrain-to-desktop
+
+	&-content
+		max-width: 670px
+
 	&-link
 		text-align: right
-
-		.btn-primary
-			position: relative
-			right: 46px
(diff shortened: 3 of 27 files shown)


SaptakS:
The richtext parts look good, but the footer still seems broken to me. Is the footer fix part of this PR? Because it's marked in the description.

chigby:
It was supposed to be part of the PR! The thing I fixed was the text color. Maybe there was something else, though? I'll re-check.

chigby:
So, I'm noticing two more things that are different between Figma and how it is in this PR:

1. Font size of footer nav links needs to be reduced to become the same size as the info text.
2. The gray text right now is #757575, but in the Figma it is #8F8F8F.

Anything else I might be missing?

harrislapiroff:
Haven't checked over the PR so not sure if you got this, but the major issue was that the top two sections here are supposed to be on a white background:

image

Here's what it looks like on the current site:

image

Should also constrain them to the container width.

I think probably those two sections deserve to be their own components—they're not really linked with the black footer except conceptually they're all "footer"

SaptakS:
I did discuss with @chigby about this clearing what was meant by the issue. And they are indeed separate components, and not part of footer.html, though they are imported in the footer.html file so that all three components don't need to be imported separately. So maybe some class in parent is the issue. I am also checking it right now.

SaptakS:
@harrislapiroff I have updated the code to fix that. Sadly I had to use two nested `<div>` since there are two constrains on width for citation and email signups as seen in figma. So not too happy about the markup, but it's not invalid or anything either. @chigby I think the color difference is because we chose that darker gray for table call-to-action links since that didn't meet WCAG standard. I verified that even with what we have, the contrast standards pass in the footer, so I am okay with it.

Issue #1287, opened 2022-03-22
Homepage charts refinement


harrislapiroff:
- [x] Hover on treemap seems broken (broken on `/charts/` as well, so this is in the React component, not the integration) #1289 
- [x] Clicking on charts should actually take you to the database instead of raising an alert #1289 
- [x] Tags aren't showing up as clickable (bug)
- [x] Clicking on state on map should bring you to that state (currently broken, bug) #1297
- [x] Add ability to choose a start year or date range (we have a few incidents recorded from before 2017, but we should not actually show buttons for those years)
- [x] Adjust spacing between elements to match mockups
- [x] Add a better loading indicator _or_ bake the data into the initial page request instead
- [x] "Incidents recorded outside of US" should not show if count is zero
- [ ] **Stretch goal:** Adjust CSS to use our breakpoints
- [ ] **Stretch goal:** Adjust CSS to use our grid maybe?
- [ ] **Stretch goal:** Replace mouseup events with `<a href>` to get browser-default accessible link behaviors (cmd click, screenreader sees links, url preview, etc.)
- [ ] **Stretch goal:** Homepage map "Incidents outside of US" link should link to database filtered to incidents not in the US

harrislapiroff:
Okay, actually tags _are_ showing up as clickable, it's just that we don't have many incidents from the last six months on the redesign server, and particularly not many with the tags I selected to preview with. It would be a nice stretch goal to have a message on the disabled buttons on hover that explains why they're disabled.

harrislapiroff:
I'm going to close this. There's a few stretch goals that I'm not sure if we hit or not, but we should open new tickets for the ones that we want to do.

PR #1297, opened 2022-03-25
Fix charts map

harrislapiroff:
* Make map dots link to filtered database by state
* Fix rendering and linking of "Incidents outside the US" (see commit for details)

Addresses issues in #1287


chigby (approved):
These changes work as described, for me. Merging it.

PR #1299, opened 2022-03-25
Add date limits to homepage viz

harrislapiroff:
This will be necessary because we have a couple incidents from pre 2017 in the database, but overall we consider our database to have started in 2017, therefore we only want to visualize years 2017+ on the homepage.

In the process I also fixed a proptype issue on a component and switched all javascript date handling to UTC because it's easier. (I wonder if there's a linter that would limit us to _only_ UTC dates 😅 ... or maybe time to start using `Temporal`?)

Reading through this PR commit by commit should be an easy way to review it.


harrislapiroff:
Hmmm my fiddling with dates has predictably created a mess. Going to try to untangle this on Monday.

chigby (approved):
Definitely very positive date math changes. 

I'll note an edge case where where if the start date or end date is within six months of "now," then the "past six months" filter gets cut off (i.e. incidents from that six month interval are not shown), but this is not a case we expect to come up in production.

PR #1301, opened 2022-03-26
Adds and fixes hover effect for cards and table rows

### LINKED ISSUE #1248: Hover effect on table rows and cards
All the table rows in the new redesign has a [hover effect](https://www.figma.com/proto/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3897%3A62716&scaling=min-zoom&page-id=3897%3A59348&starting-point-node-id=3897%3A59349) that transitions the entire row. 

1. Now 1 solution is to put the entire row inside an anchor tag, but some rows have links inside the row, e.g. in the about page when there's an email. 
2. The other solution is to make it into a div and control it with JS, which is a terrible way of making a link and accessibility nightmare.
3. The third solution would be to have an absolutely positioned pseudo element inside a link that also covers the entire size of row (see 3rd example here: https://css-tricks.com/nested-links/). The issue with this is, in a table row, adding this link makes no sense in any of the columns and the markup looks very messy.

I think we should have a discussion around what's the best solution. I would go with solution 1, in case the internal link in the table and the link for the entire row points to the same link. But not sure what's the best solution if that's not the case.

I think we need to have a similar discussion regarding the card hover effect as well since the implementation right now makes the layout a bit iffy sometimes, and any other way would bring up the exact same questions as above.

SaptakS:
Fixes #1248 

- This adds a hover effect for the table rows
- Fixes the hover effect on cards so there is no reflow of content
- Makes the entire card/table rows clickable (PS: if JS is disabled, only the title in the card, and the link in the table rows will be clickable)


client/common/sass/components/_blog-card.sass
```diff
@@ -1,12 +1,23 @@
...
 	&--info
 		margin-top: 12px
 		margin-bottom: 0
 		transition: margin-bottom 0.09s ease-in
 
+	&--inner
+		transition: transform 0.15s ease-in-out
+		&:hover
+			transform: translateY(-0.5rem)
+			cursor: pointer
```
chigby, line 15: I did notice that when Javascript is disabled, there are parts of the card where the cursor is a pointer but it is not clickable. I'm not sure how to fix this, or if it's even worth doing, but it was something I noticed.

chigby (approved):
I think the hover effect on the cards is (a) nicer now that the whole card can be clicked, and (b) no longer page-layout breaking. So this is a big win. I left one comment about the progressive enhancement situation but it's not a blocker at all.

I do think the sliding table rows with disappearing text look a little funny, but that's a design decision, and so I figure I can't really comment on that. I will point out that, from comparing the page effect to what's in the Figma, it looks like there are some slight differences compared with this PR. Namely:
1. The right arrow symbol becomes slightly larger when the row is in the hovered state
2. The "Call to action label" (the rightmost column text) seems to slide smoothly out of the column, rather than instantly vanishing. We do seem to have sliding text on "unhover" (when the text slides back in).

I'm approving this PR to speed up merging. @SaptakS when you log on, feel free to merge, or add changes to this based on the above comments, or we can open separate issues for them if we feel they are important (or just ignore them entirely if they aren't, which may be the case in fact).

client/common/sass/components/_blog-card.sass
SaptakS: Yes. I did mention this. I don't know any easy way of not doing it either.

SaptakS:
> The right arrow symbol becomes slightly larger when the row is in the hovered state

I did notice that. The issue is they are using svg and we are using `&rarr;` (which I feel is the right approach). If I am increasing the font-size as shown in the design, then in some rows, it is causing a reflow. So to avoid that I made the increase in font-size very small. I am not sure if we want to go the svg route since that will also cause more jerky animation I think.

> The "Call to action label" (the rightmost column text) seems to slide smoothly out of the column, rather than instantly vanishing. We do seem to have sliding text on "unhover" (when the text slides back in).

I noticed this too. Honestly, I don't know of a good easy way of not making it vanish instantly and make it look like the text is sliding out and the arrow is sliding in. It can be done, but I feel the markup will be really messy. So I just left it like this. If anyone has any strong opinion on this, I can check.

harrislapiroff (approved):
I think it would be nice to get this closer to the mockup animation but let's call that a stretch goal. This looks good as is.

PR #1303, opened 2022-03-28
1296 Fix search button hover state

urlsangel:
Fixes the white top and bottom border on search button hover by removing fixed position and making button 100% height of its container.


chigby (approved):
Looks good to me!

PR #1308, opened 2022-03-29
Fix CSS warnings

### LINKED ISSUE #1293: Home page charts CSS errors
When hovering the mouse cursor over the category boxes on the home page chart controls, in Firefox, I'm noticing a large number of errors in the console (visible if CSS warnings are turned on):

image

harrislapiroff:
* Remove `fill` attribute from a ReactAnimatedDataset that generates
  lines. Lines don't have fills
* Use d3.event from the old version of d3-selection (1.4.1) that is
  bundled with react-animated-dataset in order to provide
  event info to the tooltip callback


chigby (approved):
Looks good to me! This fixed the errors locally and brought back the tooltip.

PR #1321, opened 2022-03-30
Tweak spacing issues around tags on incident cards

urlsangel:
Apologies @SaptakS - I know you just reviewed and merged a tweak for these!

After doing some browser testing I've tweaked a couple of CSS properties for a better layout of these tags.

### DIFF
--- a/client/common/sass/components/_incident-database-card.sass
+++ b/client/common/sass/components/_incident-database-card.sass
@@ -30,9 +30,6 @@
 		+desktop-up
 			grid-column: span 3
 
-	&__tags
-		line-height: 2.2
-
 	&__details-title
 		+desktop-up
 			display: none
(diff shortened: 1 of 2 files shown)


SaptakS:
This seems to remove the margin from the top of the tags, which is needed based on the figma design. Maybe you can add a margin top in `.tag-list`?

urlsangel:
> This seems to remove the margin from the top of the tags, which is needed based on the figma design. Maybe you can add a margin top in `.tag-list`?

Hey @SaptakS 

I think this is a bit whack-a-mole, as the top margin isn't necessary in some contexts, but might be in others. I'll do some local dev later this afternoon to try to capture all contexts and make the list look good wherever it's used - ignore this PR for now 👍

SaptakS:
I feel applying the top margin on the parent ul instead of individual li should actually fix for almost all cases if I am not wrong.

urlsangel:
> I feel applying the top margin on the parent ul instead of individual li should actually fix for almost all cases if I am not wrong.

@SaptakS - yep, I think that works for all the use cases I can see , I've added that now 👍

SaptakS (approved):

PR #1328, opened 2022-03-30
Category page fixes

### LINKED ISSUE #1311: Rich text fixes
- [x] The category methodology code is rendering HTML tags. I think maybe it needs to be run through `|richtext`?
    image

- [x] We should add stat tag parsing everywhere we have rich text. Currently not present on, e.g. category description
    image

SaptakS:
Resolves #1311 

- Adds richtext filter for methodology description
- Uses render_as_template for rendering category description so as to render the statistics properly.

### DIFF
--- a/common/templates/common/_category_methodology_section.html
+++ b/common/templates/common/_category_methodology_section.html
@@ -2,7 +2,7 @@
 
 <h2 class="methodology-section-header">Data Tracked For This Category</h2>
 
-<div class="methodology-summary">{{ methodology.description }}</div>
+<div class="methodology-summary rich-text">{{ methodology.description|richtext }}</div>
 
 <dl class="methodology-details-table">
 	{% for item in methodology.data_items %}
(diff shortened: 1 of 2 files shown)


common/templates/common/_category_methodology_section.html
```diff
@@ -2,7 +2,7 @@
 
 <h2 class="methodology-section-header">Data Tracked For This Category</h2>
 
-<div class="methodology-summary">{{ methodology.description }}</div>
+<div class="methodology-summary rich-text">{{ methodology.description|richtext }}</div>
```
chigby: I know this isn't a thing you added, but now I'm really confused by what this `.description` is doing here. `methodology` is a rich text field, right? What does adding `.description` do? Can we remove it? If I remove it locally, the output doesn't seem to change.

SaptakS: Good point. I think in the original PR, some unecessary bit got left behind. removed it.

PR #1329, opened 2022-03-30
Style cleanup database page

### LINKED ISSUE #1305: Database page style cleanup
- [x] Get rid of bullets and make checkboxes and radio buttons flush with left text edge
- [x] Style input fields with black border
- [x] Add position sticky to form section headers

chigby:
Fixes #1305 

This PR adjusts the styles of the DB page in accordance with the above issue. I also added an additional change to remove browser-level autocomplete from trying to fill in the "City" field on top of the datalist autocomplete options our site provides.

I'm actually not sure what adding position sticky is accomplishing on the section headers. I did some scrolling around the page and they didn't seem to stick for me. Maybe it's because it's already nested in a sticky positioned element? I'm not sure what the overall design goal of that change is either, so I'm slightly at a loss for alternatives.

### DIFF
--- a/client/common/sass/components/_filters.sass
+++ b/client/common/sass/components/_filters.sass
@@ -21,6 +21,7 @@
 	background-position: right
 	padding-right: 17px
 	cursor: pointer
+	position: sticky
 
 .filters__group
 	border-bottom: 1px solid rgba(0, 0, 0, 0.1)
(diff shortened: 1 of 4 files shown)


urlsangel:
@chigby 

I've made a couple of amends to `_filters.html` to avoid duplicate labels for radio/checkbox elements and added `<strong>` to differentiate between filter elements a little.

Issue #1307, opened 2022-03-29
Browser testing issues


urlsangel:
@harrislapiroff 

I'm going to put all of my findings into one issue for now to avoid polluting Github, they can then be pulled out into separate issues based on priority etc.

urlsangel:
### ~Social media nav in footer too close to footer links~

Resolved #1386 

* Desktop/Chrome/Mac
* https://redesign.pressfreedomtracker.us/

<img width="813" alt="Screenshot 2022-03-29 at 09 49 42" src="https://user-images.githubusercontent.com/560463/160573405-c443c39c-65f3-4cca-af49-7f6cb3c3ffa5.png">

* Mobile/Chrome/Galaxy 22
* https://redesign.pressfreedomtracker.us/

<img width="276" alt="Screenshot 2022-03-29 at 14 29 29" src="https://user-images.githubusercontent.com/560463/160622506-21cbd4be-95a5-4e5e-b58e-57d2c45ebc26.png">

urlsangel:
### ~Mobile nav icon margin incorrect~

**Edit: resolved here:** https://github.com/freedomofpress/pressfreedomtracker.us/pull/1399

* Mobile/Chrome/Mac (replicated on mobile device via BS)
* https://redesign.pressfreedomtracker.us/

Figma has the margins as `margin-top: 8px` and `margin-right: 16px`, the actual is `17px` and `33px` respectively.

<img width="285" alt="Screenshot 2022-03-29 at 10 12 03" src="https://user-images.githubusercontent.com/560463/160578114-88d8682b-43f3-47f4-b8ff-b3a0dc2ff8e8.png">

urlsangel:
### Chart filters spacing appears inconsistent

- Safari/iPhone 13
- https://redesign.pressfreedomtracker.us/

The Figma design doesn't have stacked filters (not sure what was meant in the design), but the rows look a little close together, and are missing an overline (which may lead to it feeling like a big gap above the top filter buttons row).

<img width="379" alt="Screenshot 2022-03-29 at 10 19 38" src="https://user-images.githubusercontent.com/560463/160580780-d48e5262-5459-4be8-a5e8-b05a1eafa6e3.png">

urlsangel:
### ~Button bottom border and search placeholder overlaps~

**Edit: fixed**

* Safari/iPhone 13
* https://redesign.pressfreedomtracker.us/

[...]

PR #1335, opened 2022-03-31
iOS CSS fixes

### LINKED ISSUE #1362: Button primary is a little pokey
Browser: Safari Version 15.4 (17613.1.17.1.6) macOS

image

urlsangel:
This PR fixes or partially fixes these issues noted here:

- https://github.com/freedomofpress/pressfreedomtracker.us/issues/1307#issuecomment-1081675555
- https://github.com/freedomofpress/pressfreedomtracker.us/issues/1307#issuecomment-1081645663
- https://github.com/freedomofpress/pressfreedomtracker.us/issues/1307#issuecomment-1081707289
- https://github.com/freedomofpress/pressfreedomtracker.us/issues/1307#issuecomment-1083174288

### DIFF
--- a/client/common/sass/components/_buttons.sass
+++ b/client/common/sass/components/_buttons.sass
@@ -29,7 +29,8 @@
 			content: url(../icons/arrow-left.svg)
 			height: 24px
 			width: 16px
-			border: 3px solid
+			// 2.95px fixes border overlap issue on iOS
+			border: 2.95px solid
 			border-color: $black
 			position: absolute
 			right: -3px
--- a/client/common/sass/components/_buttons.sass
+++ b/client/common/sass/components/_buttons.sass
@@ -4,7 +4,7 @@
 	font-family: $font-base
 	font-size: $font-size-small-p
 	font-weight: 500
-	line-height: 1.714
+	line-height: 1.7145
 	padding: 0.5rem 0.75rem
 	background-color: $white
 	color: $black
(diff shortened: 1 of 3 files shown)


client/common/sass/components/_buttons.sass
```diff
@@ -28,7 +28,8 @@
 			content: url(../icons/arrow-left.svg)
 			height: 24px
 			width: 16px
-			border: 3px solid
+			// 2.95px fixes border overlap issue on iOS
+			border: 2.95px solid
```
SaptakS: With 2.95px, it looks overlapping in firefox and chrome. I think maybe we can keep 3px for now, and find a better way to solve this later? @harrislapiroff

SaptakS (commented):
Left one comment. Apart from that all looks good.

SaptakS:
Updated this PR to increase the line-height slightly (without that slight change safari was computing it as 23.99 instead of 24px and causing issues by taking the floor value instead of round-off value). This should now fix https://github.com/freedomofpress/pressfreedomtracker.us/issues/1362

SaptakS:
Updated this PR to increase the line-height slightly (without that slight change safari was computing it as 23.99 instead of 24px and causing issues by taking the floor value instead of round-off value). This should now fix https://github.com/freedomofpress/pressfreedomtracker.us/issues/1362

PR #1339, opened 2022-03-31
Add targets to incident database cards and incident detail page

chigby:
This PR adds the display of the "Targets" data item (and appropriate links to the DB page) to incident database cards and the incident detail page. It also enables filtering the DB page by the name of the targeted journalist (where before it was only possible by ID).


SaptakS (approved):
LGTM

PR #1340, opened 2022-04-01
Order categories on homepage chart alphabetically

### LINKED ISSUE #1315: How are categories ordered on the home page charts?
Should probably either be by size or alphabetically. Currently appears random.

harrislapiroff:
I wanted to order them largest to smallest, but the way the lines and
rects are specified in the treemap component was complicated to
untangle so I'd like to save that for a future task

Resolves #1315


chigby (approved):
Looks good to me!

PR #1341, opened 2022-04-01
About page polish

### LINKED ISSUE #1312: About page table polish
- [ ] Tables should collapse on mobile ([Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3689%3A53132))
- [ ] On desktop, hovering over a table row should show its image ([Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3922%3A65418))

chigby:
Fixes #1312 

The CSS in this is a little jank, I think, but I'm not sure there's another way to do this? I was also having a hard time getting the image to fade _out_, as having `display: none` applied makes it vanish instantly (this is my theory anyway). So, let me know if you think there's a better way here.

### DIFF
--- a/common/templates/common/blocks/info_table/_url.html
+++ b/common/templates/common/blocks/info_table/_url.html
@@ -6,8 +6,8 @@
             {% if item.image %}
                 <div class="list-table__col info-table-row__hover-only">
                     <figure class="info-table__figure">
-						{% image item.image fill-66x66 as small_img %}
-						{% image item.image fill-120x120 as large_img %}
+						{% image item.image height-66 as small_img %}
+						{% image item.image height-120 as large_img %}
 						<img
 							alt="{{ item.title }}"
 							srcset="{{ large_img.url }} 120w, {{ small_img.url }} 66w"
--- a/common/templates/common/blocks/info_table/_url.html
+++ b/common/templates/common/blocks/info_table/_url.html
@@ -16,7 +16,7 @@
                 </div>
             {% endif %}
             <div class="list-table__col info-table__title">{{ item.title }}</div>
-            <div class="list-table__col">
+            <div class="list-table__col info-table__url">
 				<a href="{{ item.url }}" class="text-link" rel="nofollow">{{ item.url }}</a>
 			</div>
             <div class="list-table__link info-table__col-right info-table__col-bottom">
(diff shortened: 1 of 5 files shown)


common/templates/common/blocks/info_table/_url.html
```diff
@@ -4,9 +4,14 @@
      <li class="list-table__row">
         <div class="anim-table-row list-table__item info-table--col-3 {% if item.image %}info-table--row-3 info-table--col-3-desktop{% endif %}">
             {% if item.image %}
-                <div class="list-table__col list-table__small_screen_only">
+                <div class="list-table__col info-table-row__hover-only">
                     <figure class="info-table__figure">
-                        {% image item.image fill-66x66 alt=item.title %}
+						{% image item.image fill-66x66 as small_img %}
+						{% image item.image fill-120x120 as large_img %}
```
SaptakS: `fill-` will crop the image. Do we want the image to crop or scale to fit?

SaptakS (approved):
Let what small comment. Not sure what should be the decision. Apart from that looks good to me.

common/templates/common/blocks/info_table/_url.html
chigby: Hmm, maybe we should change to scaling, as you have done in #1345? (i.e. replace `fill` with `width` here)

chigby: Having done some testing, I think actually height is the correct thing to scale to, as the design expects a fixed height.

SaptakS: Makes sense to me.

harrislapiroff:
@chigby Haven't looked at the code so not sure if this is applicable, but could you use `opacity: 0; pointer-events: none;` instead of `display: none;`?

chigby:
@harrislapiroff Interesting idea, let me give that a try

chigby:
I pushed a commit that _mostly_ gets the "fade out" transition working, but sometimes I notice a ghost image appearing for a fraction of a second at the _top_ of the info table during that "dehover" animation. I've tried to figure our what's causing that and I can't really. Maybe someone else can? @harrislapiroff @SaptakS

SaptakS (approved):
I think this looks good enough to me for now.

PR #1345, opened 2022-04-04
Makes primary image and richtext images scale and not crop

### LINKED ISSUE #1343: Weird image sizing issues
Getting a lot of weird image behaviors. See [this page](https://redesign.pressfreedomtracker.us/blog/between-the-bookends-1-year-of-press-freedom-violations/) for examples.

There might need to be a variety of approaches to resolving this, but here are some initial things that should happen:

- [x] Images in stream fields and rich text fields should be scaled to fit, not crop
- [x] Primary images on blog detail pages and incident detail pages should be scaled to fit, not crop
- [x] Images on blog/incident cards should continue to be scaled to crop

Screen Shot 2022-04-01 at 17 12 51

SaptakS:
Fixes #1343

### DIFF
--- a/common/templates/common/_video_or_image.html
+++ b/common/templates/common/_video_or_image.html
@@ -5,7 +5,7 @@
         {% if video %}
             {% embed video %}
         {% elif image %}
-            {% image image fill-910x510 alt=image.attribution|default:image.title %}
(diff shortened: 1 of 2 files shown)


harrislapiroff (approved):
Code looks good, getting it onto staging

PR #1350, opened 2022-04-04
Fix category page rendering

chigby:
This PR fixes the 500 error on category pages. 

I'll note that this problem would have been caught earlier by tests, but we had marked the ones that would have caught it as `skip`, so they were not being run :sweat_smile:


harrislapiroff:
Hmmm I wonder if we should unmark some of those skip tests or if that would just open pandora's box...

harrislapiroff (approved):
Code looks good, let's get it staged

PR #1351, opened 2022-04-04
Switch to Helvetica Neue with system font backups everywhere

harrislapiroff:
Helvetica Neue is the preferred font for this design, but system
fonts are the ones most likely to be present and have the range
of weights we would like to use if Neue is not


harrislapiroff:
Self merging

PR #1352, opened 2022-04-04
Correctly display `case_type` field

chigby:
We have been showing the "value" for this field, rather than the display text.


harrislapiroff (approved):
💯

PR #1357, opened 2022-04-05
Makes mobile menu fixes

### LINKED ISSUE #1342: Mobile menu fixes
Current look:

Screen Shot 2022-04-01 at 16 03 26

[Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3128%3A29842)

Fixes:

- [x] background color should be off white
- [ ] buttons should have some margin on the bottom, currently flush
- [ ] add 4px spacing between text and border on menu items

SaptakS:
fixes #1342 

image

I made the flex wrap so that instead of the text in button wrapping, it goes to next line, I feel that looks better than having the buttons side-by-side and then the text inside breaking the button layout. Let me know if you feel strongly otherwise.

### DIFF
--- a/client/common/sass/components/_menu.sass
+++ b/client/common/sass/components/_menu.sass
@@ -85,6 +85,7 @@
 			text-decoration: none
 			justify-content: space-between
 			border-bottom: solid 1px #c1c1c1
+			padding-bottom: 4px
 			font-weight: 700
 
 			&:hover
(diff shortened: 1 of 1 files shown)


chigby (approved):
I think wrapping is fine here, it allows for more flexibility actually in case we ever want to add more items to this menu. The rest also looks good :+1:

PR #1365, opened 2022-04-05
Fix footer margin

### LINKED ISSUE #1346: White space between related incidents and footer on incident detail page
Screen Shot 2022-04-04 at 11 54 04 AM

Should be no white space, to match figma

SaptakS:
Fixes #1346

### DIFF
--- a/client/common/sass/common.sass
+++ b/client/common/sass/common.sass
@@ -22,7 +22,6 @@
 @import components/blog-card
 @import components/blog-list
 @import components/text-field
-@import components/footer
 @import components/footer-info
 @import components/emails-signup
 @import components/citation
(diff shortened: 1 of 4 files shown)


chigby (commented):
So I think this does fix the overly large footer margin on the home page, but I'm not sure if what's here matches [the Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=4120%3A80644):

Desktop: 104px (plus some whitespace around the button too??) 
image

Mobile: 64px
image

Tablet: 72px
image
Also apparently the button is centered (which it's not in what we have now, but also probably not a high priority).

Do you think it's possible to make this margin responsive as in the design or should we just keep it like it is now?

SaptakS:
@chigby updated to make the spacing responsive

chigby (approved):

PR #1370, opened 2022-04-05
Fixes search button for safari

### LINKED ISSUE #1361: Search button in webkit is taller than box
It seems the solution in #1303 fixed the search button for Gecko, but threw it off for WebKit:

Screen Shot 2022-04-05 at 11 06 50 AM

Possibly we need a `box-sizing: border-box` declaration here?

SaptakS:
Fixes #1361 

Seems like safari was adding a marging of 2px around the entire input field and the search bar as a browser default appearance. So I added margin: 0 in both to ensure that no margin is added.

### DIFF
--- a/client/common/sass/components/_search-bar.sass
+++ b/client/common/sass/components/_search-bar.sass
@@ -7,5 +7,6 @@
 	right: 2px
 	top: 2px
 	border: 0
+	margin: 0
 	font-weight: 500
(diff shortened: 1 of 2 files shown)


chigby:
I don't have an easy way to review this in Safari

SaptakS:
I tested it in my personal macbook. Looked okay.

PR #1371, opened 2022-04-05
Maintains aspect ratio for article content images

### LINKED ISSUE #1364: Combination of max-width and image dimensions is creating distorted images
Screen Shot 2022-04-05 at 11 29 07
https://redesign.pressfreedomtracker.us/blog/bevy-of-court-rulings-for-public-records-and-press-freedom/

I think the solution here is to remove the width and height attribute from the HTML, but if there's a way to preserve the ratio before the image is loaded, to prevent reflow, that would also be great.

SaptakS:
Fixes #1364

### DIFF
--- a/client/common/sass/components/_article-content.sass
+++ b/client/common/sass/components/_article-content.sass
@@ -14,6 +14,7 @@
 
 	img
 		max-width: 100%
+		height: auto
 
 	.details-table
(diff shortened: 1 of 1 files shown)


chigby (approved):
Fixed them for me, let's merge! :sparkles:

PR #1372, opened 2022-04-05
Rename "Physical Attack" category to "Assault"

### LINKED ISSUE #1367: Change hard coded references from "Physical Attack" to "Assault"
Editors changed the name of the category, so we need to update anywhere in the code that relies on the old name. I'm hoping this is *just* data vizzes, but we should do a global search to be sure.

(Someday the categories api will make changes like this trivial 😅 )

chigby:
The slug for this category page is changing, so we're moving
everything we can here from the old name to the new name.  Not
everything changed in this commit is required for the site to still
work, but it makes it easier to change as much as we can in one go.

Fixes #1367


harrislapiroff (approved):

harrislapiroff (approved):
I amended changes for the chart javascript. This is ready

PR #1381, opened 2022-04-05
Limit "recent incidents" on category page to only that category

### LINKED ISSUE #1378: Category pages showing *all* incidents rather than category-specific ones
They should show _only_ incidents for the category under "Recent Incidents"

chigby:
This, I think, is the most minimal change needed here to make this
section display the correct data.  I think a lot of what's in the
`context` dict of the category page is obsolete, but that's best
solved another time.

Fixes #1378


SaptakS (approved):
Looks good to me.

PR #1383, opened 2022-04-05
Replace reset button with "clear all" link

chigby:
The `<button type="reset">` restores the form to the state it was in
_when the page loaded_, meaning the pre-populated values (from the GET
parameters) are still there.  What we want is "clear all"
functionality, which would put the blank value into all fields, even
if they were pre-populated.  This change accomplishes that.

### DIFF
--- a/incident/templates/incident/_filters.html
+++ b/incident/templates/incident/_filters.html
@@ -2,7 +2,7 @@
 	<header class="filters__header">
 		Filters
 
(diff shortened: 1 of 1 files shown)


SaptakS (approved):
Yes. Reset does always reset to the initial state of the form when page loaded. So probably this is the best solution for now. I think in future we should progressively enhance this using javascript so that it doesn't reload the page but just empty the form.

PR #1384, opened 2022-04-05
Add prop for tooltip (when unselectable) to our `Button` components

chigby:
This might be kind of a hack, but I couldn't really figure out how
else to display this!  But on the other hand I also feel like it's
just adding some flexibility to our "Button" and "ButtonsRow"
components even if we're not using it anywhere but one specific set of
buttons just yet.

Reviewer: Let me know if you don't like this implementation and possibly we can come up with something else together.

### DIFF
--- a/client/charts/js/components/Button.jsx
+++ b/client/charts/js/components/Button.jsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import Tooltip from './Tooltip'
 
 const textStyle = {
 	fontFamily: 'var(--font-base)',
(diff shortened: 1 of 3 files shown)


harrislapiroff (approved):
I think this is actually a pleasant solution! I hadn't seen this sort of thing solved in a React-y way (most tooltip libraries seem to do their own logic for rendering and positioning using refs). I would have worried all the rerenders could be costly, but it was perfectly smooth in my testing.

PR #1391, opened 2022-04-06
Restore wagtail userbar

chigby:
See documentation:

https://docs.wagtail.org/en/stable/topics/writing_templates.html#wagtail-user-bar


harrislapiroff (approved):
Looks good

SaptakS (commented):
The docs mention `You should consider putting the tag after any [skip links](https://webaim.org/techniques/skipnav/) but before the navigation and main content of your page.` So maybe a better place to put this is the _menu.html after the skiplink and before `<header>` ?

SaptakS (approved):

PR #1403, opened 2022-04-07
Restore loading of piwik/matomo JS file when analytics is enabled

### LINKED ISSUE #1397: Reinstate analytics
It looks like the Matomo tracking code is currently missing. Let's reinstate so we can monitor the success of the new site effectively :)

chigby:
This was removed when we re-did the templates, and I'm bringing back the old code for this verbatim.

Fixes #1397


harrislapiroff (approved):
Perfect!

PR #1405, opened 2022-04-07
Homepage API performance proof of concept

chigby:
This pull request creates a separate API "action" (in Django Rest Framework's terminology) to return the data required by the home page with as little overhead as I can manage. Namely, I'm running as few DB queries as possible and not using any serializers.

The action lives at http://localhost:8000/api/edge/incidents/homepage_csv/ and you can test it with something like:

```bash
curl -H 'Accept: text/csv' "http://localhost:8000/api/edge/incidents/homepage_csv/"
```

When I do this it returns data quite quickly.

The purpose of doing this is (1) to help debug what's slowing down the "regular" API endpoint but showing what's possible under a "best-case" scenario, and (2) potentially create a stopgap (or even long-term) solution for improving the performance of the API for the home page, which we know is being requested quite a bit.

This PR also updates the front end to use the new URL.

### DIFF
--- a/client/common/js/home.jsx
+++ b/client/common/js/home.jsx
@@ -35,7 +35,7 @@ chartContainers.forEach((node) => {
 
 	ReactDOM.render((
 		<DataLoader
-			dataUrl={`/api/edge/incidents/?${params.toString()}`}
+			dataUrl={`/api/edge/incidents/homepage_csv/?${params.toString()}`}
 			loadingComponent={(
 				<HomepageMainCharts selectedTags={selectedTags} databasePath={databasePath} data={[]} loading={true} />
 			)}
--- a/home/templates/home/home_page.html
+++ b/home/templates/home/home_page.html
@@ -34,7 +34,7 @@
 				data-tags="{{ data_viz_tags_json|escape }}"
 				{% if search_page %}data-database-path="{% pageurl search_page %}"{% endif %}
 				{% if page.viz_data_start %}data-start-date="{{ page.viz_data_start|date:'c' }}"{% endif %}
-				{% if page.viz_data_end %}data-start-date="{{ page.viz_data_end|date:'c' }}"{% endif %}
+				{% if page.viz_data_end %}data-end-date="{{ page.viz_data_end|date:'c' }}"{% endif %}
 			></div>
 		</div>
 	{% endwith %}
--- a/incident/api/views.py
+++ b/incident/api/views.py
@@ -100,8 +100,10 @@ class IncidentViewSet(viewsets.ReadOnlyModelViewSet):
         context = super().get_renderer_context()
 
         # Get set of fields from serializer that has been pruned
-        # according to request's query-string parameters.
-        if 'homepage_csv' not in self.request.path:
+        # according to request's query-string parameters.  In the
+        # homepage_csv action, this set is pre-determined and not
+        # affected by the request, in that case we skip this step.
+        if self.action != 'homepage_csv':
             context['header'] = list(self.get_serializer().fields.keys())
         return context
 
--- a/incident/api/views.py
+++ b/incident/api/views.py
@@ -133,10 +133,10 @@ class IncidentViewSet(viewsets.ReadOnlyModelViewSet):
         lower_date = request.GET.get('lower_date')
         upper_date = request.GET.get('upper_date')
 
-        tag_summary = models.IncidentPage.objects.annotate(
+        tag_summary = models.IncidentPage.objects.only('tags').annotate(
             tag_summary=StringAgg('tags__title', delimiter=', ')
         ).filter(pk=OuterRef('pk'))
-        category_summary = models.IncidentPage.objects.annotate(
+        category_summary = models.IncidentPage.objects.only('categories').annotate(
             category_summary=StringAgg('categories__category__title', delimiter=', ')
         ).filter(pk=OuterRef('pk'))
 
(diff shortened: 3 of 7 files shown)


urlsangel:
@chigby 

I've just pulled and tested this locally, and it is blazingly fast, the charts are available almost as quickly as the page HTML renders 👍 

I'm going to switch back to where I was with the original API call, and see if I can pin the bottleneck down, but I think what you said in Slack is correct from what I could see, namely that "for the API, Neither the CSV generation nor JSON generation is our code" - it looks like it's DRF apart from the serializer definition itself.

urlsangel:
Just adding a few more notes here as I go:

Testing @chigby's new endpoint on this branch, I get an average response time of around 200ms:

Screenshot 2022-04-08 at 09 33 12

Testing against the initial investigative changes I made yesterday (deferring fields that aren't required), I get an average response time of of around 4.5 - 5 seconds:

Screenshot 2022-04-08 at 09 34 20

This morning, I've refactored to use a simple DRF serializer that only defines the fields required for the CSV (`date, city, state, latitude, longitude, categories, tags`), and commented out all other fields that were being pre-fetched etc. elsewhere. This gives me a response of around 1400ms:

Screenshot 2022-04-08 at 09 41 49

 So, even with a heavily optimised serializer, this is still _seven times_ slower than @chigby's pared down response.
 
 I'm reading a couple of articles now to see if there are any tips on how to make DRF serializers faster.

urlsangel:
Most articles I've found on speeding up DRF serializers focus on `select_related` and `prefetch_related`, which is already being done here: https://github.com/freedomofpress/pressfreedomtracker.us/blob/develop/incident/models/incident_page.py#L104

This article is more in keeping with @chigby's approach: https://hakibenita.com/django-rest-framework-slow#prior-work

Screenshot 2022-04-08 at 09 56 28

urlsangel:
@chigby 

I've added `.only()` to the to `IncidentPage` querysets to reduce SQL queries - it doesn't look like it's made a difference to performance, but every little helps right?

harrislapiroff:
Hm, I like that this is fast, but I want to think about it a little more. I don't love creating custom endpoints on an ad hoc basis—I'd love a more general solution. But if this is the best practical way to speed things up, I'm open to it. The homepage is certainly worth making exceptions to keep zippy. Maybe worth some discussion at web sync.

harrislapiroff:
Chatted with @chigby about this in a call and we decided this would be a good solution to implement for an immediate improvement to our performance woes, but we'll continue to contemplate the best way to make performance improvements to the API generally, potentially using some of the lessons learned from implementing this specific endpoint and maybe even with the intention of eventually obviating it in favor of a more general solution.

@chigby will ensure this PR is ready to merge, then I'll give it a quick review, and hopefully all will be well to merge and deploy it.

home/templates/home/home_page.html
```diff
@@ -34,7 +34,7 @@ <h1 class="page-title" id="title">
 				data-tags="{{ data_viz_tags_json|escape }}"
 				{% if search_page %}data-database-path="{% pageurl search_page %}"{% endif %}
 				{% if page.viz_data_start %}data-start-date="{{ page.viz_data_start|date:'c' }}"{% endif %}
-				{% if page.viz_data_end %}data-start-date="{{ page.viz_data_end|date:'c' }}"{% endif %}
+				{% if page.viz_data_end %}data-end-date="{{ page.viz_data_end|date:'c' }}"{% endif %}
```
harrislapiroff, line 37: oops

client/common/js/home.jsx
```diff
@@ -35,7 +35,7 @@ chartContainers.forEach((node) => {
 
 	ReactDOM.render((
 		<DataLoader
-			dataUrl={`/api/edge/incidents/?${params.toString()}`}
+			dataUrl={`/api/edge/incidents/homepage_csv/?${params.toString()}`}
```
harrislapiroff, line 38: Hm—I wonder if it would make sense for us just to embed this url in the html now that I'm thinking about it. Def doesn't need to be for this PR.

incident/api/views.py
```diff
@@ -85,7 +101,8 @@ def get_renderer_context(self):
 
         # Get set of fields from serializer that has been pruned
         # according to request's query-string parameters.
-        context['header'] = list(self.get_serializer().fields.keys())
+        if 'homepage_csv' not in self.request.path:
+            context['header'] = list(self.get_serializer().fields.keys())
```
harrislapiroff: This seems like a hacky way to accomplish this. Can we rely on a view name or something instead of a URL segment?

harrislapiroff: Maybe it doesn't matter since the same class is controlling the URL, eh?

incident/models/incident_page.py
```diff
@@ -144,6 +145,27 @@ def updated_within_days(self, days):
             updated_days_ago__lte=days
         )
 
+    def fuzzy_date_filter(self, lower=None, upper=None):
```
harrislapiroff, line 148: Let's get a docstring here

harrislapiroff (approved):
Looks good. Add a docstring to `fuzzy_date_filter` and this is good to merge.

incident/api/views.py
chigby: This is a good point, though, I think there is a better way to do this, having investigated a little more about how DRF actions operate.

chigby:
@harrislapiroff Should be ready for re-review

harrislapiroff:
Looks great!

PR #1406, opened 2022-04-08
Activate carousel buttons on window resize

urlsangel:
This adds a throttled window resize event handler to set the carousel button disabled attribute based on article visibility.

Fixes note "Carousel buttons inactive if page is resized to small": https://github.com/freedomofpress/pressfreedomtracker.us/projects/4#card-80136199

### DIFF
--- a/client/common/js/ArticleLoader.js
+++ b/client/common/js/ArticleLoader.js
@@ -154,6 +154,7 @@ ArticleLoader.NUM_AUTO_FETCHES = 0
 
 document.addEventListener('DOMContentLoaded', () => {
 	if (window._articleLoader) {
+		// eslint-disable-next-line no-console
 		console.warn('An ArticleLoader instance already exists.')
 		return
 	}
--- a/client/common/js/articleScroller.js
+++ b/client/common/js/articleScroller.js
@@ -1,3 +1,5 @@
+import { throttle } from './utils'
+
 class ArticleScroller {
 	constructor(scrollerElement) {
 		// Since JS is loaded we can show the interactive horizontal scroller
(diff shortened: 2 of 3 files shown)


chigby:
Having played around with this branch locally, I think we might want to reassess how badly we want this feature, as it's more complex than it appears. The situation is that when the window resizes, a lot of stuff about the page changes that affects how the carousel scrolling works. 

1. At different screen widths, the per-article width changes, so this will also need to be recalculated or the scroller will not move the right amount to the left or right. But you can't _just_ change the per-article width, since this number is already baked into the `shift` value if scrolling has taken place prior to the resize. There is probably more than one way to solve this, but the current way of tracking scroll position will have to be changed.
2. The visibility criterion for whether or not the scroll buttons are disabled is not quite robust enough right now to handle resizing. If you're scrolled all the way to the last article, you can resize the screen such that the "scroll right" button becomes enabled again and you can scroll past the end. I have had some success by creating some different functions for determining if the button can be clicked, but it probably needs further testing:
```js
canScrollToPrev() {
    let firstArticle = this.articles[0]
    return firstArticle.getBoundingClientRect().x < 0
}

canScrollToNext() {
    let lastArticle = this.articles[this.articles.length - 1]
    return lastArticle.getBoundingClientRect().x > this.perArticleWidth
}
```
3. There needs to be a check to "reset" the translateX transform if the window starts small is resized to be large enough that the scrollbars are supposed to disappear. I was able to accomplish this by checking a media query in the JS code, i.e. something like: 

```js
const desktopWidth = window.matchMedia('(min-width: 1153px)')
if (desktopWidth.matches) { 
// remove transform
}
```

But I don't love this solution as it creates kind of an annoying coupling between our media breakpoints and our JS code. Maybe there's some other way?

Anyway, I think we ought to hold off on merging this PR until at least these situations can be handled (and, frankly, there are probably more that I didn't find).

harrislapiroff:
I think perhaps we should hold off on this, because maybe we need a bigger restructure of the carousel code. I've noticed some quirkiness with it lately that I think could be fixed by re-conceiving its architecture. I think we should be making the carousel element scrollable at an OS level (i.e., with scrollbars/touch drag) and the buttons should just be a secondary way of navigating (i.e., a shortcut for "scroll to next element").

Seen that way I _think_ it would be easier to just enable/disable scrollbar buttons according to whether the element has a larger scrollWidth than clientWidth and what the current scroll position is.

harrislapiroff:
Closing as wont merge

PR #1411, opened 2022-04-14
Improve performance of database page

urlsangel:
I've profiled the incident index page with Django Debug Toolbar, and added comments to the `get_context` method to capture a rough snapshot of where the bottlenecks are (these comments are in the first commit of this branch).

This is imprecise, as Dango won't actually hit the DB if a queryset is set up but then not used in something like a template render, but it was enough to get an idea of where the time and queries were jumping up.

Two of the big bottlenecks were using `len(qs)` on querysets rather than `qs.count()` - the latter is more efficient and is recommended by the [Django docs](https://docs.djangoproject.com/en/4.0/ref/models/querysets/):

> len(). A QuerySet is evaluated when you call len() on it. This, as you might expect, returns the length of the result list.
> 
> Note: If you only need to determine the number of records in the set (and don’t need the actual objects), it’s much more efficient to handle a count at the database level using SQL’s SELECT COUNT(*). Django provides a [count()](https://docs.djangoproject.com/en/4.0/ref/models/querysets/#django.db.models.query.QuerySet.count) method for precisely this reason.

Another improvement was adding `only('category')` to the `context['categories']` variable.

It looks like there could be more efficiencies around the `with_most_recent_update` part, but this is pretty complex, and I'm wary of breaking things as this is so low-level, and used in multiple parts of the site.

For me, these changes have improved local performance significantly, but I expect that the page render speed can likely still be improved.

### DIFF
--- a/incident/models/incident_index_page.py
+++ b/incident/models/incident_index_page.py
@@ -165,71 +165,41 @@ class IncidentIndexPage(RoutablePageMixin, MetadataPageMixin, Page):
     def get_context(self, request, *args, **kwargs):
         from common.models import CategoryPage
         context = super(IncidentIndexPage, self).get_context(request, *args, **kwargs)
-        # CPU: 13622.299 msec
-        # SQL: default 90.77 ms (47 queries including 26 similar and 22 duplicates )
 
+        # Before perf change:
         # context['all_incident_count'] = len(IncidentFilter({}).get_queryset())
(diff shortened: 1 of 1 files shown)


chigby (approved):
Testing this locally, just "by eye" and also with silk enabled, did show a dramatic speedup. Let's get this out there!

Issue #1419, opened 2022-05-09
HTTP_CF_CONNECTING_IP header is not logged in Django on PFT


mig5:
It would be great to log the HTTP_CF_CONNECTING_IP in common/middleware.py on PFT like we do on some other sites in order to get the correct IP in logs.

This request is motivated by the fact that we are seeing spikes in traffic to PFT that cause performance issues. I note that the header was added to freedom.press in order to rate limit it, and that you've in fact rate-limited parts of PFT based on this header before (https://github.com/freedomofpress/pressfreedomtracker.us/commit/c90183dc4c99d4ef291fbe411bdcdd8c348a0a06) . Maybe would be good to also rate-limit on `/all-incidents/` as that seems to be getting the spikes in traffic and thus the performance issues. https://logs.freedom.press/goto/57bb6a00-cf56-11ec-9a03-052c3afc8b82

PR #1420, opened 2022-05-09
Replace X-Real-IP with X-CF-Connecting-IP in log meta

maeve-fpf:
Addresses https://github.com/freedomofpress/pressfreedomtracker.us/issues/1419 (not closing, let's use the data once we have it to decide if rate limiting will help)

We don't need X-Real-IP as it is (in our deployment) the same as X-Forwarded-For. Also, correct the spelling of X-Scheme.


chigby (approved):
Everything seems good here, merging it.

PR #1423, opened 2022-05-16
Fix header name: no X- in CF-Connecting-IP

maeve-fpf:
This corrects #1420, which spelled the header name `X-CF-Connecting-IP`. It's actually just `CF-Connecting-IP`, see: https://developers.cloudflare.com/fundamentals/get-started/reference/http-request-headers/


chigby (approved):
Looks good to me

PR #1424, opened 2022-05-16
Fix home page date filter

chigby:
This PR fixes the date filter on the home page CSV endpoint: the querystring variable it was looking for was wrong. Also adds a basic test to ensure this filter runs without triggering an error.


harrislapiroff (approved):
Code looks straightforward!

PR #1430, opened 2022-05-23
Remove related incident cache purging

### LINKED ISSUE #1155: Rein in overly aggressive frontend cache purging
Yesterday we noticed a spike in the thousands of Tracker nginx logs of the form

```
Purged page IncidentPage with title: Portland Tribune reporter pushed by police officer while covering far-right rally and slug: portland-tribune-reporter-pushed-police-officer-while-covering-far-right-rally
```

with a number of different pages specified. I'm fairly sure this was caused by the logic here

https://github.com/freedomofpress/pressfreedomtracker.us/blob/develop/incident/signals.py#L46

where we purge the cache for every incident that shares a category with the incident currently being edited. I believe we do this because it has the potential to affect the "related incidents" section of the page for all those incidents. @KioHerr reported that she recalls having to reload the page a few times while editing a particular incident because it was being slow. I'm guessing it was this incident, due to the overwhelming number of separate API requests we were sending for it.

I could imagine a couple ways of solving this and am open to discussion:

- [ ] Be less aggressive with what we choose to purge. Maybe we can pinpoint which incidents might actually have this one in their related incident section? Or we could decide it doesn't matter if that section gets out of date (this might cause broken links if the URL of an incident changes or an incident is deleted—otherwise it should only cause an outdated headline or summary in the "Related incidents" section of another incident page)
- [ ] Batch these requests: I believe cloudflare lets you purge a batch of urls with a single API request. If we could batch these, we could reduce the API request count to something reasonable.

chigby:
Fixes #1155 

This pull request alters the "page published" signal for Incident Pages to remove the purging of "related" incidents (as we've defined it) and incidents of the same categories from the cache. When an incident page is published, these other incidents will not be purged from the cache anymore.

### DIFF
--- a/incident/signals.py
+++ b/incident/signals.py
@@ -26,29 +26,6 @@ def purge_incident_from_frontend_cache_for_category(
         )
 
 
-def purge_incident_from_frontend_cache_for_incident(
-    instance=None, sender=None, **kwargs
-):
-    """
-    Busts the cache for any incident that shares a category or is
-    a related incident to the incident just changed.
-    """
-
-    # Purge cache for related incidents
-    for incident in instance.related_incidents.all():
-        purge_page_from_cache(incident)
-        logger.info(
-            f"Purged page IncidentPage with title: {incident.title} and slug: {incident.slug}"
-        )
-
-    # Purge cache for incidents that share a category and may
-    # therefore show up as a related incident
-    for categorization in instance.categories.all():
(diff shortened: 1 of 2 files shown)


SaptakS (approved):
Looks good to me logically. Will need to keep checking in production if this disrupts or causes any unexpected behaviors.

PR #1434, opened 2022-05-24
Adds all category icons in the incident title

### LINKED ISSUE #1421: [Incidents detail page] Incident title showing up with only one category
Even if an incident has multiple categories, they're only showing up with one category symbol in the title on the incident detail page. Here's an example: https://pressfreedomtracker.us/all-incidents/new-york-post-photographer-assaulted-by-man-being-evicted-from-encampment/

image

SaptakS:
Fixes #1421 

Removes the category page symbol classes from the h1 element. Loops over all the categories in the incident and adds a span tag with the category page symbol and also a title attribute to show the category title (and also announce in screen readers)

PS: The category title shown by title attribute won't be visible to keyboard users and touch users so they will just see icons (which might be fine since the text isn't super important).

### DIFF
--- a/incident/templates/incident/incident_page.html
+++ b/incident/templates/incident/incident_page.html
@@ -1,7 +1,12 @@
 {% extends "base.html" %}
 
 {% block main %}
-	<h1 class="category category-{{ main_category.page_symbol }} article-title" id="title">{{ page.title }}</h1>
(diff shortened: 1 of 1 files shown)


chigby (approved):
This looks great! Merging it.

PR #1438, opened 2022-06-06
Increase category visibility and clickability

### LINKED ISSUE #1433: Not all incident detail pages offer clickpath/explanation for category
Example 1:
https://pressfreedomtracker.us/all-incidents/congressional-candidates-ad-touts-crowd-control-munition-use-against-press/

In this example, the meaning of the symbol is not explained in the incident summary - it's only at the bottom of the page under "Related incidents" that I see an explanation that this symbol stands for "Chilling Statement".

Example 2: 
https://pressfreedomtracker.us/all-incidents/reporter-arrested-phone-confiscated-while-covering-nc-homeless-camp-eviction/

In this case the icon is prominently explained, but I can't click to get to other pages in the same category except by way of the links under individual related incidents at the bottom.

I'm flagging these examples for discussion as I've not checked against the Figma design or prior issues if these are intentional tradeoff decisions. It seems to me that, at a minimum, we'll always want to make sure to explain the symbols.

chigby:
Fixes #1433 

This PR fixes the first two items from the ticket. I did not work on the third item because (1) I think it needs a little more fleshing out, and (2) it would not solve the problem for the "Other Incident" category. Let me know, though, if you think something like that is desired.

- Category table "headers" on the database page and incident detail page are now clickable links to the appropriate category page.
- The category headers are shown even if there is no related data, and they're shown after the data-having categories.

Examples:

## Incident detail page

image

## Database page

image

### DIFF
--- a/incident/models/incident_page.py
+++ b/incident/models/incident_page.py
@@ -829,9 +829,15 @@ class IncidentPage(MetadataPageMixin, Page):
     def get_category_details(self, index=None):
         if not index:
             index = self.get_parent()
+
         category_details = {}
+        categories_without_metadata = {}
         for category in self.categories.all():
             category_fields = CATEGORY_FIELD_MAP.get(category.category.slug, [])
+
+            if not category_fields:
+                categories_without_metadata[category.category] = []
+                continue
             category_details[category.category] = []
             for field in category_fields:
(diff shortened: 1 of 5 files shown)


SaptakS (approved):
Looks good to me.

PR #1442, opened 2022-06-13
Use mailchimp API for signup and newsletter block

### LINKED ISSUE #1436: Reinstitute newsletter signup form
We used to have a page that would render a newsletter signup, I believe using the `EmailSignupBlock`. Using this block now throws a 500 (presumably because it has no template). We should update this block to use the Mailchimp API to render an email signup form.

SaptakS:
Fixes #1436 

Moves the entire subscription process to use mailchimp API similar to https://github.com/freedomofpress/freedom.press/pull/1036/. Unlike mailchimp implementation in FPF, uses site settings here instead of page settings.


common/views.py
```diff
@@ -47,6 +61,62 @@ def serve(*args, **kwargs):
     return response
 
 
+class SubscribeForSite(View):
+    def post(self, request):
+        if request.is_ajax():
```
chigby: I wonder if we can replace this with something like [`request.accepts`](https://docs.djangoproject.com/en/3.2/ref/request-response/#django.http.HttpRequest.accepts) since `is_ajax` is, I think, deprecated and kind of old. (I realize that most of this code here is copied over from freedom.press but it's something that stood out to me here as something we could do to improve on that code)

common/utils/mailchimp.py
```diff
@@ -0,0 +1,54 @@
...
+        raise ApiKeyMissingError
+
+    try:
+        client = mailchimp_marketing.Client()
+        client.set_config(
+            {'api_key': settings.MAILCHIMP_API_KEY}
+        )
+        groups_by_audience = defaultdict(list)
+        for group in settings.emails.EmailSettings.mailchimp_groups.all():
+            groups_by_audience[group.audience_id].append(group.group_id)
```
chigby, line 40: As was comparing the differences between this code and the code from freedom.press, I realized I was becoming confused, and then eventually I discovered the reason for that, but it is this: the `MailchimpGroup` class from freedom.press is not the same as the `MailchimpGroup` class that's here. They _used_ to be the same, when the class stored information about how to construct a `<form>` element with data about how to submit that form to Mailchimp and it contained the `category_id` and `group_id` fields. But that class on freedom.press was changed to store information about how to submit an API request to Mailchimp, and now contains the `audience_id` and `group_id` fields -- we will have to make that change here too, or this code won't work.

Sorry this is confusing, it took me a while to figure this out and I wrote both versions of that code! :confused:

common/templates/common/_email_signup.html
```diff
@@ -2,9 +2,17 @@
 	{% for group in email_settings.mailchimp_groups.all %}
```
chigby, line 2: This is related to another comment I've left here, but the `for` statement here only works correctly when the `mailchimp_groups` information is being used to directly create `<form>` elements, which is the legacy functionality. In the new system, we only need to check that mailchimp groups are _present_ before displaying the form. This template should match what we're doing over here: https://github.com/freedomofpress/freedom.press/blob/develop/home/templates/home/_mailchimp_form.html

chigby (changes requested):
Left some comments on additional things that need updating for the API-based system from freedom.press to work, let me know if you need any more elaboration or help on anything

chigby (approved):

chigby:
Changes look good, let's get this coverage situation sorted and then we'll be good!

PR #1445, opened 2022-06-23
Incident card fixes

### LINKED ISSUE #1443: Homepage/Incident card table fixes
image

- [ ] Don't render a location line at all if location is `None`
- [ ] top-align labels and content (e.g., in this screenshot "Targets" should be visually up one line of text to align with the first row of targets)

SaptakS:
Fixes #1443 

- [x] Don't render a location line at all if location is None. Also added individual city and state checks to better modify the display
- [x] top-align labels and content

### DIFF
--- a/client/common/sass/components/_incident-database-card.sass
+++ b/client/common/sass/components/_incident-database-card.sass
@@ -74,7 +74,6 @@
 
 		+tablet-up
 			display: flex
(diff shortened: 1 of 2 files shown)


chigby (approved):
Changes look good to me! Merging.

PR #1446, opened 2022-06-28
Fix broken links to tag filter

harrislapiroff:
Previously the react charts were replacing spaces with '-'s for
generating urls with tags in them, but since the tags are in
querystrings and need to be converted back to spaces for searching
'+' is the correct character to use. The previous approach was
generating filtered pages with no results


chigby (approved):
Looks good to me, this fixed the problem for me locally.

PR #1453, opened 2022-07-18
Incident sidebar filters

chigby:


chigby:
@SaptakS I have added the filter component for type autocomplete. The next thing on my list is the Submit button-- the logic for transforming the current `filterParameters` object into a correctly formatted string is slightly tricky but I've got an okay start on it so far.

chigby:
Just pushed a commit that gets the submit button working

SaptakS:
I have got the visual designs mostly finished and also cleaned up the markup a bit. Added labels, and made each category option markup wise into a checkbox so its both keyboard interactive and makes the intent clear.

This is how it looks:
image

One of the options clicked:
image

SaptakS:
- [x] I am currently working on adding the general filters

SaptakS:
Few things to note that we need to work:

- [x] Fixing the filter responsiveness
- [ ] Adding fieldset labels for the checkbox (and also need to check all input fields for proper labelling for accessibility)
- [ ] Do we want the filters to be applied on click or on clicking "Submit" button? I am in favor of, for now (and maybe even for later?), to use the button submit rather than auto apply (because that means applying a loading overlay screen)

Lastly, I noticed that the number of incidents shown after filtering is more than the number shown for the category in the filter. @chigby do you know why that might be happening?

chigby:
I've checked over the code Saptak added to this PR and it looks great to me.

harrislapiroff (approved):
This looks great! Definitely some future things I want to do, but for now let's get it merged and deployed!

Future items:

- Accordion animation on selecting a category, I think. The current reflow can be disorienting
- Is the apply filters button too much tucked off the bottom of the page when at the top of the page? Not sure the solution here
- Did we agree we wanted category selection to be an OR rather than AND? I can't remember, but OR feels a little unnatural to me with the graph...

PR #1455, opened 2022-07-25
Validate fuzzy date filter inputs

chigby:
This pull request adds validation to the homepage CSV view to check that the dates we receive as input are valid before actually running any DB queries using them. Further explanation for individual changes made in commit messages.

This fixes a problem where a request like this:

```
GET /api/edge/incidents/homepage_csv/?date_lower=INVALID&fields=categories%2Cauthors%2Cdate%2Ccity%2Cstate%2Clatitude%2Clongitude%2Ctags&format=csv
```

Would result in a 500 error. It now results in a 400 error, and that error happens much earlier in the response chain.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1010,9 +1010,9 @@ uritemplate==4.1.1 \
     # via
     #   -r requirements.txt
     #   drf-spectacular
-urllib3[secure,socks]==1.26.10 \
-    --hash=sha256:8298d6d56d39be0e3bc13c1c97d133f9b45d797169a0e11cdd0e0489d786f7ec \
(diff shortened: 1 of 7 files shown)


harrislapiroff (approved):
Confirmed tests fail on develop, succeed on branch. Manually tested the endpoint and it returned a 400 when fed an invalid date and a CSV otherwise.

Question: Is this the only endpoint affected by this issue (maybe specifically because we bypassed DRF for it)?

harrislapiroff:
@chigby confirmed this is not an issue on other endpoints, not because they go through DRF (this one does too) but because they went through our incident filter system. What this PR does is use that filter system in a limited way for the homepage endpoint.

PR #1458, opened 2022-08-01
Fix missing default values for Mailchimp Groups fields

chigby:
Default values are required if there are existing rows for MailchimpGroup objects, otherwise there would be no way to add a column for, for example, audience id as NOT NULL without giving it a default value.


harrislapiroff (approved):
Code looks good.

PR #1460, opened 2022-08-02
Fix mailchimp form names

chigby:
Fixes form names to match expected names, also updates logging to be slightly more informative about this (and some other) errors.


harrislapiroff (approved):
Code looks good! Let’s try it on staging!

SaptakS (approved):

PR #1472, opened 2022-08-12
Fixes asset loading by replacing file-loader with asset modules

SaptakS:
Replaces file-loader with asset modules of webpack since the file-loader module seems to be deprecated: https://github.com/webpack-contrib/css-loader/issues/1358#issuecomment-897542202

The current website doesn't load any of the assets that are loaded through css (e.g, url(), background-image, content: , etc.)Though i don't know why this particular update broke this, as this seems to have been for quite some time in webpack, and its deprecation not support removal, but also we just updated webpack by minor release.

Anyways, this seems to fix loading of assets. Also updated fonts to use the same asset module.


harrislapiroff (approved):
pour one out for `file-loader`

PR #1484, opened 2022-08-29
Arrests category improvements

### LINKED ISSUE #1018: Arrests category improvements
The current structure of charges in the database needs some finessing. In particular, we want to be able to associate charge status with specific charges since a journalist may be charged with multiple things and one may be dropped and the other not. We also want to record a timeline of when charges get dropped, go to trial, undergo appeal, etc.

Let's add the following to incidents:

An inline `IncidentCharge` object with a nested inline `ChargeUpdate` object.

Charge should have a single autocomplete field to charges

ChargeUpdate should have: date and status. It's possible we'll add other fields (such as date of trial) but I need a little time to think about how they should be structured.

We'll also need a migration to move data from the old fields in a logicial way, but we should probably preserve the old fields deprecated for at least a month.

chigby:
Fixes #1018 

This pull request consolidates the `dropped_charges` and `current_charges` incident page fields into a single `charges` field that consists of an `IncidentCharge` model with potentially many `ChargeUpdate` objects related to it. These are as described in the issue except that there are status and date fields on the `IncidentCharge` model itself, which act as "initial" values before any updates are applied. This is because (1) I think the admin UX is nicer here for charges without updates (flat is better than nested), and (2) I was having a hard time getting `min_num` working for the nested inline panel, so this is an easier way to require a status/date for each charge.

As an aside, I also added the `--no-deps` flag to places where we install things from pip, in order to prevent [this kind of bug](https://app.circleci.com/pipelines/github/freedomofpress/pressfreedomtracker.us/2926/workflows/6e006f97-ce5c-4033-8b08-d66470c8adf3/jobs/15135) problem from stopping CI builds from working.

More notes on individual commit messages.


chigby:
Looking at the coverage job results is making me realize there's a little more clean-up to do regarding lingering bits of code surrounding the legacy fields `current_charges` and `dropped_charges`. I can finish that up tomorrow, but the rest of this PR _is_ ready for review now.

SaptakS (approved):
I have reviewed the code. I am happy with the functionality to the extent that it does what it says it does. This PR kind of made me feel that maybe for PRs or feature implementations like this, we should have an additional step of user testing with our stakeholders in staging, given that this feature was requested by them, and they would be the ones who would be better able to say if this suffices, or if they wanted something different. Not sure if that makes sense. cc @harrislapiroff

chigby:
I've incorporated the data from the PFT team's spreadsheet into a data migration, so this PR is once again ready for a final review @SaptakS @harrislapiroff

incident/migrations/0073_convert_to_new_style_charges.py
```diff
@@ -1,20 +1,2739 @@
...
+    "new-hampshire-newspaper-publisher-arrested-charged-with-violating-political-ad-laws": [
+        ChargeInfo(
+            charge_title="publishing: failure to identify political advertising",
+            date=datetime.date(2022, 8, 24),
+            status="CHARGES_PENDING",
+            notes="Hearing scheduled for Oct. 19, 2022.",
+            updates=[],
+        )
+    ],
+}
```
harrislapiroff, line 2700: Bold of you not to use a CSV for this 😅

harrislapiroff (approved):
Gonna give this a quick test on my local machine and then merge. Looks good on quick visual code review—didn't go too deep, but we'll ask PFT folks to check it out on staging.

harrislapiroff:
I think we should clean up the aesthetic here, but I'll put that on my list for a quick mockup later

PR #1487, opened 2022-08-30
Do not install pip dependencies

chigby:
This pull request makes it so that when we install pip packages on our docker containers, we use the `--no-deps` flag. This should address a problem where sometimes the pip install command would fail when new versions of urllib3 were released.

In theory, pip-tools figures out the dependencies for us and puts what is needed into the `*requirements.txt` files.  Adding this flag fixes a bug where sometimes the `--require-hashes` check would fail when dependencies were followed by pip that were not specified in the requirements file, and thus were not hashed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ jobs:
 
       - run:
           name: Install pip dependencies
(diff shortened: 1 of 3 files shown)


SaptakS (approved):
I think this makes sense to me. Merging.

PR #1492, opened 2022-09-12
Don't log additional request_failed event if error is a 404

maeve-fpf:
Bring 404 handling into line with freedom.press. This is copied from @chigby's work there. If you make a 404ing request on each site, you'll see that freedom.press logs a `request_finished` event only, but here we log `request_finished` and `request_failed`.

All other 4xx response that generate an exception internally will still be logged at `error` level.

### DIFF
--- a/common/middleware.py
+++ b/common/middleware.py
@@ -1,6 +1,7 @@
 import time
 import uuid
 
+from django.http import Http404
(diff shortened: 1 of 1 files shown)


chigby (approved):
Looks great!

PR #1493, opened 2022-09-12
Strip ASCII "nul" character (0x00) from text before filtering on it

chigby:
Running a query with this character in it in a `where` clause will cause a

```
ValueError: A string literal cannot contain NUL (0x00) characters.
```

From Django's postgres backend.  We have to strip these characters out manually before the `Filter` class converts this to a DB query. [Example of this on the prod server](https://logs.freedom.press/app/discover#/doc/filebeat-*/filebeat-gke-2022.09.12?id=RfNJM4MBobO3kQ3QaajM).

### DIFF
--- a/incident/tests/test_filtering.py
+++ b/incident/tests/test_filtering.py
@@ -57,6 +57,21 @@ class TestFiltering(TestCase):
 
         self.assertEqual({incident1}, set(incidents))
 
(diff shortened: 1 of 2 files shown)


SaptakS (approved):
This looks like a good initial solution for the fix.

PR #1494, opened 2022-09-13
Log query string information

chigby:
This PR adds the query string to information we are logging about incoming requests.

I'm deciding to log the "raw" string and the "parsed" string separately since Django's parsing does omit information (if a key appears more than once), but I also find the parsed version more legible most of the time.

### DIFF
--- a/common/middleware.py
+++ b/common/middleware.py
@@ -36,6 +36,9 @@ class RequestLogMiddleware(object):
         request_id = str(uuid.uuid4())
 
         request_dict = {key: getattr(request, key, '') for key in self.request_keys}
+        if request.GET:
+            request_dict["query_string_raw"] = request.GET.urlencode()
+            request_dict["query_string_parsed"] = request.GET.dict()
         request_dict["user"] = str(request_dict["user"])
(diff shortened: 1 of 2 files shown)


SaptakS (approved):

maeve-fpf:
Hey, not to be really annoying and bring this up after it's shipped, but, is the parsed version actually useful? I find it kind of cluttered here as we automatically unpack the whole JSON log object into nested fields which in this case probably aren't too useful to index on their own:

Screenshot_2022-09-21_15-01-11

Also, a bad actor who is aware of this could spam that list (probably not a potential DoS, but frustrating).

chigby:
Huh, I had thought that parsing it would make it more readable but this seems like the opposite of that. Maybe parsing it into something that doesn't nest as easily into the overall JSON structure, something like:

```
[('categories', '1,2'), ('date_lower', '2022-01-01')]
```

Which I do find easier to read than what's being recorded for the raw query string item:

image

But maybe that could be improved by converting url-encoded characters before logging it, (and just removing the parsed data from the logs). Either way I'm fine with it.

PR #1496, opened 2022-09-14
Sign up form tests

chigby:
Add selenium-based tests, with JS enabled and disabled, to the Mailchimp signup form in the page footer.


SaptakS (approved):
Looks good to me.

PR #1498, opened 2022-09-20
Fixes export GET parameters for ManyRelationValues

### LINKED ISSUE #1495: Export returning column headers only
On the Incident Database page, selecting "Download your filtered data" appears to link to an endpoint that returns a CSV file with just column headers and no data.

image

Here is the endpoint I was linked to when filtering down to just Leak Cases. Looks like the URL is malformed.
```
https://pressfreedomtracker.us/all-incidents/export/?categories=ManyRelationValue%28pks%3D%5B7%5D%2C+strings%3D%5B%5D%29
```

In general is there a reason we're still using the `/export/` URL? Does its output differ from using an API query like:
```
https://pressfreedomtracker.us/api/edge/incidents/?[FILTER PARAMS]&format=csv&limit=[NUMBER OF INCIDENTS MATCHING FILTERS]
```

SaptakS:
Fixes #1495

### DIFF
--- a/incident/models/incident_index_page.py
+++ b/incident/models/incident_index_page.py
@@ -1,5 +1,6 @@
 import csv
 import json
+import copy
 from typing import TYPE_CHECKING
 from urllib import parse
 
@@ -19,7 +20,7 @@ from common.models.settings import SearchSettings
 from incident.models.export import to_row, is_exportable, to_json
 from incident.models.incident_page import IncidentPage
 from incident.utils.forms import get_filter_forms
-from incident.utils.incident_filter import IncidentFilter, get_serialized_filters
+from incident.utils.incident_filter import IncidentFilter, ManyRelationValue, get_serialized_filters
 from incident.feeds import IncidentIndexPageFeed
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


incident/models/incident_index_page.py
```diff
@@ -200,12 +201,16 @@ def get_context(self, request, *args, **kwargs):
             ).only('category')
 
         # Added: .only('category') to both queryset queries above
-
         if incident_filter.cleaned_data:
+            export_filter_data = copy.deepcopy(incident_filter.cleaned_data)
+            for name, value in export_filter_data.items():
+                if isinstance(value, ManyRelationValue) and value.pks:
```
chigby: I think that technically this does not cover all valid configurations that a ManyRelationValue could have, for instance using the query string `?equipment_broken=photometer` (or anything in the equipment broken field, or other text based relationship) will result in a `ManyRelationValue(pks=[], strings=['photometer'])`. So I think at least we must also handle this case.

chigby: It is also possible, though maybe not so through the user interface, to have both `pks` and `strings`, if you are entering the URL manually. This is maybe less of a concern but technically it is valid will perform the correct search, for example `categories=4,Assault`

chigby (changes requested):

chigby (approved):

PR #1508, opened 2022-10-19
Adds chart to the category pages

### LINKED ISSUE #1479: Category page charts
Let's add the option to add charts to the category page as shown in Accurat's [Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3584%3A55956):

image

For now let's keep the implementation limited:

- [ ] Allow content editor to choose to show either a bar chart or no chart in the category page editing UI
- [ ] Aggregate bar chart by incidents
- [ ] Allow content editor to specify a date range to chart (as with the homepage charts)

But bear in mind these future requirements when implementing to make sure we can adapt to them easily ipn the future (if some of these are quick to do now, feel free):

- We will allow multiple types of charts eventually
- We may want to allow the content editor to choose to aggregate by number of targets or number of charges instead of number of incidents

SaptakS:
Resolves #1479


client/charts/js/components/BarChartCommon.jsx
```diff
@@ -33,17 +33,19 @@ const borders = {
 
 const textPadding = 10
 
-export default function BarChartHomepage({
+export default function BarChartCommon({
```
harrislapiroff: I think it would actually be straightforward to just call this `BarChart`

client/charts/js/components/CategoryPageChart.jsx
```diff
@@ -0,0 +1,57 @@
...
+	databasePath = window.location.pathname,
+	loading = false,
+}) {
+	const chartHeight = width > 480 ? 500 : 480
+
+	const datasetGroupedByYears = groupByYearsSorted(dataset)
+
+	return (
+		<Flashing flashing={loading}>
+			<div className={'cpChartContainer'} style={{ width: width }}>
```
harrislapiroff, line 36: I notice strings in a lot of this code (including some of the stuff I wrote) are javascript expressions. Is there a reason we're doing that that I've forgotten, or can we open an issue to clean that up? It seems weird to me.

client/charts/js/components/BarChartCommon.jsx
```diff
@@ -120,6 +122,7 @@ export default function BarChartHomepage({
 				<svg
 					width={width}
 					height={height}
+					aria-labelledby={id}
```
harrislapiroff: Does this property get ignored by browsers if `id` is an empty string?

client/common/js/categoryPage.jsx
```diff
@@ -0,0 +1,40 @@
...
+		['categories', categoryId]
+	])
+
+	// If start or end date are set, limit the query to those dates
+	if (startDate) params.append('date_lower', startDate)
+	if (endDate) params.append('date_upper', endDate)
+
+	ReactDOM.render((
+		<DataLoader
+			dataUrl={`/api/edge/incidents/?${params.toString()}`}
```
chigby, line 33: I hope the performance on this API call ends up being acceptable. I could see us having to make some similar optimizations to category-level data that we made for the home page data. :/

common/models/pages.py
```diff
@@ -404,6 +447,10 @@ def get_context(self, request, *args, **kwargs):
             )
         ]
 
+        # Hard code the date and month. The state
+        context['viz_data_start'] = f"{self.viz_data_start}-01-01"
+        context['viz_data_end'] = f"{self.viz_data_end}-12-31"
```
chigby, line 450: I notice we're not using `viz_type` here -- how does this field get used if we want to turn off the chart?

common/models/pages.py
```diff
@@ -53,6 +55,8 @@
 
 from .choices import FILTER_CHOICES
 
+BIRTH_YEAR_CHOICES = ['1980', '1981', '1982']
```
chigby: Included by mistake? :sweat_smile:

client/charts/sass/ChartDescription.sass
```diff
@@ -0,0 +1,8 @@
+.chartDescription
+	height: 2rem
+	padding: 0.75rem
+	font-family: var(--font-base)
+	font-weight: 400
+	font-size: 1rem
+	color: #757575
+	margin-bottom: 1rem
```
chigby, line 8: THANK YOU for moving this stuff out of the Javascript :tada: It gets me a little every time I see the static CSS in there

client/charts/js/components/TreeMap.jsx
```diff
@@ -212,6 +213,7 @@ export default function TreeMap({
 				<svg
 					width={width}
 					height={height}
+					aria-labelledby={id}
```
chigby, line 216: I'm a big fan of this

common/models/pages.py
```diff
@@ -313,6 +321,24 @@ class CategoryPage(MetadataPageMixin, Page):
...
+    viz_data_start = models.PositiveIntegerField(
+        'Start Year',
+        blank=True,
+        null=True,
+    )
+    viz_data_end = models.PositiveIntegerField(
+        'End Year',
+        blank=True,
+        null=True,
+    )
```
chigby, line 339: I think this is a fine start for how to structure visualization data on category pages, but I worry a little bit that it's not scalable if we ever wanted to have more than one chart on a page. Though maybe that's not something we're planning.

chigby: Ok I just looked at the ticket and it says "multiple types of chart" which isn't necessarily the same as multiple charts, and which this data schema handles fine.

If we _did_ want to make it easier to add multiple charts in the future, I could actually see replacing these with a StreamField, where the different block types could correspond to the different chart types.

chigby (changes requested):

chigby:
Oh, I forgot to type this in the review box, but I think before this can be merged we need to be able to disable the charts. See the comment about `viz_type` -- if that's set to "none" the chart should not be rendered, right?

client/charts/js/components/CategoryPageChart.jsx
SaptakS: I think we can clean it up.

client/charts/js/components/BarChartCommon.jsx
SaptakS: ignored in the sense that the svg won't have an accessible name, yes.

common/models/pages.py
SaptakS: Yep. 😓

common/models/pages.py
SaptakS: I was supposed to put it in the html itself for checking, and then pass it in the component. Might have missed doing that.

client/common/js/categoryPage.jsx
SaptakS: I did have a similar discussion with @harrislapiroff raising the exact same concerns so..... 🤞🏽

client/charts/js/components/BarChartCommon.jsx
harrislapiroff: Just want to make sure it won't accidentally pair with some other element

common/models/pages.py
SaptakS: I have added a check in the HTML for the same. Also, I am sending the data in JS, and checking the vizType to load the correct chart (in this case, there's only one chart type, but in future we might have multiple types to chose from)

chigby (approved):

chigby:
Changes look good to me. @harrislapiroff do you think we should merge it?

harrislapiroff:
LGTM!

PR #1509, opened 2022-10-25
Updates pytest to remove dependency of py

SaptakS:
We have been getting safety fails because of a [CVE filed in py](https://github.com/pytest-dev/py/issues/287). The pytest maintainers have now made a [new release of pytest](https://github.com/pytest-dev/py/issues/287#issuecomment-1290407715) without py as a dependency. So upgrading pytest should fix the CI.


chigby (approved):
Looks great, though it may in the future be worth it to see if we really even need pytest at all, given what we're using it for.

PR #1517, opened 2022-10-31
Fix handling of charge updates with equal dates the base charge

### LINKED ISSUE #1516: When filtering by charge status, charge status updates from the same day as the charge do not apply
If an `IncidentCharge` has been given a `ChargeUpdate` with the same `date` as the "base" `ChargeUpdate`, then the status of that update should be considered the most recent status for the purposes of filtering.

Steps to reproduce:

1. Create (or update) an incident so it has a charge (for example `rioting`) with a status of "charges pending" with a certain date (any date is fine, say, `2022-10-31`). Then add an update with "charges dropped" occurring on the same date. Then load the filtered incident page with the querystring `charges=rioting&status_of_charges=CHARGES_DROPPED` and the incident will not be shown. It will be shown if the querystring `charges=rioting&status_of_charges=CHARGES_PENDING` is used. This is the opposite of what should happen.

chigby:
Previously, charge updates with dates equal to the date on their base `IncidentCharge` object would not be considered "newer" than the base, but now they will be.

Fixes #1516

### DIFF
--- a/incident/models/incident_page.py
+++ b/incident/models/incident_page.py
@@ -152,7 +152,7 @@ class IncidentQuerySet(PageQuerySet):
         charge's most recent status."""
         newest_update = ChargeUpdate.objects.filter(
             incident_charge=OuterRef('pk'),
-            date__gt=OuterRef('date')
+            date__gte=OuterRef('date')
         ).order_by('-date').values('status')[:1]
 
         latest_status = IncidentCharge.objects.filter(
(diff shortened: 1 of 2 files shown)


SaptakS (approved):
Looks good to me. As discussed during web sync, this doesn't handle the ordering of Charge Updates if there are multiple updates on the same date, and that would require us probably to move to datetime territory. However, there was a consensus on this PR solving the main issue that we have right now.

PR #1518, opened 2022-11-01
Dynamic tag filter controls

### LINKED ISSUE #1482: Improved tag multiple-selection
It seems like we're delaying further work on the database filter interface at least a bit. PFT editors report that being able to filter by multiple tags at once more easily would dramatically improve *their* user experience (in addition to likely improving the experience for site visitors some amount).

Let's add in a better interface as easily as we can bearing in mind that it may well be an interim solution. I leave specifics to the implementer:

- Something similar to the autocomplete interface we had for the previous design would be acceptable.
- If it seems easy enough to implement something similar to what's in the [mockup](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=4330%3A93822), go for that (the mockup seems to show tags as radio buttons though—we will need them as checkboxes)

Please write a comment on this ticket with your proposed architecture before beginning work on it.

chigby:
Fixes #1482

Implements a version of Accurat's multiple-selection filter for choosing tags in the General section of the sidebar on the database page. See the issue for a screenshot of what this original spec looks like.

What appears in this branch is modified from the design document due to our use of AND logic when selecting multiple tags to filter on. This means:
1. The "top 3 tags" concept has been changed from a fixed list (based on your category selection) to a dynamic list based on what boxes you have checked. The list is still sorted by incident count (and shows 3 by default), but checking a box will bring that tag to the top of the list, since, per AND logic, all the resulting incidents from this filter _must_ have that tag. Selecting more than one tag will put them both to the top (with the same count). 
gif 1
2. Tags beyond the top 3 can be added using the autocompleting drop-down field. I created a new component for this, trying to be as generic as was possible. Once the user has chosen something from that list, it will always be shown in the list of checkboxes, even if it's unchecked. This is not an extremely useful feature right now, but I think it's something that will eventually become more valuable once the incident list live-updates based on changes to the filter sidebar-- this would let people more easily explore the different tag combinations.
Peek 2022-11-01 16-58
In the spec, there is a "clear all" button to clear out your picked tags. This is not something I feel is needed here (though maybe I'm wrong?) since the distinction between top tags and picked tags is blurred, unlike in the spec. 
3. The "see complete list" is saved for another ticket. It's nice to have, but not necessary functionality and would require implementing a modal pop-up, which is going to take some more time to get right, I think.

I encourage playing around with this while testing to see how you like it and if any interactions need smoothing out.

I've also made a few changes to the "infrastructure" of the filters, which we had largely kept as-is from Accurate. First, I've moved the code for parsing the querystring into its own file. Ideally this would eventually have the logic for re-encoding the filter parameters object _back_ into a querystring, but I will leave that for another issue. I talk a little about my ideas in the commit message.

Second, I've changed the way we update the filter parameter state to employ the react hook `useReducer` which is a better fit for what we are doing (i.e. changing the state based on the previous value of the state). Also discussed in the commit message. This involved a little bit of `useContext` as well-- something we could possibly expand beyond what I've done here, honestly.

Any and all comments welcome. I want to get this right and make it easier to deploy filter controls based on these ideas.


chigby:
As a side note, I have no idea why the lint check is failing. Any advice is appreciated.

client/common/sass/components/_filters.sass
```diff
@@ -69,3 +77,54 @@
 	margin: 0
 	font-size: 1.25rem
 	font-weight: 400
+
+
```
harrislapiroff: I'm guessing this double line break is the cause of the lint issue and it's just giving the wrong line number?

chigby: Seems like this was it! Thanks!

client/charts/js/components/Autocomplete.jsx
```diff
@@ -0,0 +1,173 @@
...
+					userInput: payload,
+					showSuggestions: true,
+				}
+
+			default:
+				throw new Error(`Unknown action type: ${type}`)
+		}
+	}
+
+	const [state, dispatch] = useReducer(reducer, initialState)
```
harrislapiroff, line 87: TIL react has a built in `useReducer`—donno how that one went by without my notice!

harrislapiroff (approved):
This could use an accessibility review (possibly by @SaptakS) and style cleanup (possibly by me) but I'm eager to get it deployed, so I'll open new tickets for that after merge.

It otherwise looks very good! Nice work.

PR #1521, opened 2022-11-07
Ignore `d3-color` vulnerability

chigby:
This cannot be fixed right now, it's a dependency of `react-animated-dataset` which has not issued an update.


harrislapiroff (approved):
Looks good

PR #1524, opened 2022-11-09
Fixes the statistics grid to fluidly occupy space based on device dimension

SaptakS:
Right now in the category pages, after the charts PR, the statistics table breaks in some dimensions. This is because the grid column widths are fixed for the statistics table, and since only 50% of the whole screen is available, sometimes, the grid row is overflowing the 50% because of fixed widths. Hence there is both overflowing and wrapping of the "See incidents" text. I have fluidified(not sure if that's a word) the first and second column (which is used for spacing) widths using `clamp`. So now, ideally it should still take the fixed widths when available, but it will take lesser width when there is no space. According to me this is an elegant solution, anything else would need us either to increase the space available for the table, or do some other hacky media query. Let me know if anyone disagrees.

image

This is how it look at 1153px now.

### DIFF
--- a/client/common/sass/components/_statistics-table.sass
+++ b/client/common/sass/components/_statistics-table.sass
@@ -20,7 +20,7 @@
 			align-items: center
 
 			+tablet-up
-				grid-template-columns:  436px 100px 30px 1fr
+				grid-template-columns:  clamp(20.00rem, calc(10.66rem + 12.96vw), 27.25rem) clamp(0.00rem, calc(-8.05rem + 11.17vw), 6.25rem) 30px 1fr
 
 	&__link
 		color: $gray-text
--- a/client/common/sass/components/_categorypage.sass
+++ b/client/common/sass/components/_categorypage.sass
@@ -8,6 +8,7 @@
 	+desktop-up
 		display: flex
 		flex-wrap: wrap
+		gap: 14px
 
 		> *
 			flex: 1


chigby:
I'm noticing something a little odd going on when the statistics table is shown at large screen sizes (for me at least), here's a demonstration:

statistics-table2

About around 1940px width (maybe this is wide enough we're not concerned) but it seems to squeeze the link text on the right. If this is something we can fix by adjusting the column parameters, I feel like we might as well handle that while we are here now. Though if that would be more complicated maybe better to leave it for another issue, or decide it's not super-important.

client/common/sass/components/_statistics-table.sass
```diff
@@ -20,7 +20,7 @@
 			align-items: center
 
 			+tablet-up
-				grid-template-columns:  436px 100px 30px 1fr
+				grid-template-columns:  clamp(20.00rem, calc(10.66rem + 12.96vw), 27.25rem) clamp(0.00rem, calc(-8.05rem + 11.17vw), 6.25rem) 30px 1fr
```
chigby, line 23: Strongly endorse the use of `clamp` here, this seems like exactly what it's meant for.

chigby (approved):

SaptakS:
> I'm noticing something a little odd going on when the statistics table is shown at large screen sizes (for me at least), here's a demonstration:
> 
> statistics-table2 [ statistics-table2 ](https://user-images.githubusercontent.com/561931/201132712-37c5f3e7-d045-4e71-a163-3c0845b5e7f0.gif) [ ](https://user-images.githubusercontent.com/561931/201132712-37c5f3e7-d045-4e71-a163-3c0845b5e7f0.gif)
> 
> About around 1940px width (maybe this is wide enough we're not concerned) but it seems to squeeze the link text on the right. If this is something we can fix by adjusting the column parameters, I feel like we might as well handle that while we are here now. Though if that would be more complicated maybe better to leave it for another issue, or decide it's not super-important.

I think we definitely should. Will look into this. I don't have a widescreen right, but will fix it by monday. Might need to adjust the second column width a bit to adjust for this.

harrislapiroff (approved):
LGTM

PR #1530, opened 2022-11-14
Handle case number and case type filters

chigby:
This pull request is a follow up to #1518. It fixes a two bugs in the filter sidebar introduced with that PR, and also adds some information to aid in future debugging.

1. `case_type` and `case_number` filters will now render correctly.
2. "string" type filters now are now correctly encoded into the query string when activating the "Apply Filters" button
3. Enhancements to dev data to more closely match what's in staging and production
4. Add warnings and errors where filter names are not found where they're expected to be found.

### DIFF
--- a/client/charts/js/components/FilterSet.jsx
+++ b/client/charts/js/components/FilterSet.jsx
@@ -241,6 +241,9 @@ export default function FilterSet({ filters, handleFilterChange, filterParameter
 	const components = filters.map((filter, index) => {
 		if (filter.name === 'search' || filter.name === 'tags') {
 			return
+		} else if (!filterParameters.hasOwnProperty(filter.name)) {
+			console.warn(`no filter parameters defined for filter "${filter.name}"`)
+			return
 		} else if (filter.name == 'tags') {
 			return (
 				<details
--- a/client/charts/js/components/FiltersIntegration.jsx
+++ b/client/charts/js/components/FiltersIntegration.jsx
@@ -40,6 +40,9 @@ function filterReducer(state, {type, payload}) {
 			return {...state}
 
 		case SET_PARAMETER:
+			if (!state.hasOwnProperty(payload.filterName)) {
+				throw new Error(`Cannot set parameter for unknown filter name "${payload.filterName}"`)
+			}
 			state[payload.filterName].parameters = payload.value
 			return {...state}
 		default:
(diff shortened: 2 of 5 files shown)


client/charts/js/components/FilterSet.jsx
```diff
@@ -241,6 +241,9 @@ export default function FilterSet({ filters, handleFilterChange, filterParameter
 	const components = filters.map((filter, index) => {
 		if (filter.name === 'search' || filter.name === 'tags') {
 			return
+		} else if (!filterParameters.hasOwnProperty(filter.name)) {
+			console.warn(`no filter parameters defined for filter "${filter.name}"`)
+			return
```
harrislapiroff, line 246: Part of me wonders if we should use `invariant` here, but that wouldn't have fixed our problem since `invariant` disables logging in prod.

client/charts/js/components/FiltersIntegration.jsx
```diff
@@ -40,6 +40,9 @@ function filterReducer(state, {type, payload}) {
 			return {...state}
 
 		case SET_PARAMETER:
+			if (!state.hasOwnProperty(payload.filterName)) {
+				throw new Error(`Cannot set parameter for unknown filter name "${payload.filterName}"`)
+			}
```
harrislapiroff, line 45: Ditto

client/charts/js/components/GeneralFilter.jsx
```diff
@@ -11,9 +15,16 @@ export default function GeneralFilter({
...
 
 	function handleFilterChange(event) {
-		setFilterParameters(event.target.name, event.target.value)
+		updateFilters({
+			type: SET_PARAMETER,
+			payload: {
+				filterName: event.target.name,
+				value: event.target.value,
+			},
+		})
```
harrislapiroff, line 27: Loving this gradual transition to using context and reducers

common/management/commands/createdevdata.py
```diff
@@ -389,6 +390,36 @@ def handle(self, *args, **options):
...
+            ['snails', 'slugs', 'squids', 'octopuses', 'cuttlefish', 'nautiluses', 'ammonites', 'clams', 'oysters'],
+            ['lizards', 'snakes', 'crocodiles', 'geckoes', 'turtles', 'tortoises'],
+            ['ostriches', 'kiwis', 'flamingoes', 'doves', 'hummingbirds', 'storks'],
+            ['kangaroo', 'opossom', 'wombat'],
+            ['shrews', 'rodents', 'rabbits'],
+            ['echidna', 'platypus'],
+            ['plum', 'orchid', 'chrysanthemum', 'bamboo'],
+            ['hyena', 'mongoose', 'civet'],
+            ['tarsier', 'lemur', 'loris', 'baboon'],
+        ]
```
harrislapiroff, line 410: 🦌🦎🐍🐒🐪🌸

harrislapiroff (approved):
LGTM, let's pop it onto staging

PR #1531, opened 2022-11-14
make blog site summary children max-width 50% - desktop up only

### LINKED ISSUE #1388: "site-summary" should be half width
figma:
image

current site:
image

shanejdonnelly:
Fixes #1388 

Screenshot from 2022-11-14 16-52-55

### DIFF
--- a/client/common/sass/components/_site-summary.sass
+++ b/client/common/sass/components/_site-summary.sass
@@ -24,5 +24,9 @@
 	&--blog
 		font-size: 1.125rem
 
+		> *
+			+desktop-up
+				max-width: 50%
+
 	&--error
 		margin-bottom: 104px


SaptakS (commented):
Looks mostly good to me. If there is a raw html with style attribute as part of the site summary (as is there in the createdevdata), then the alignment is a little messed up, but I guess we don't really encourage the editorial team to use raw HTML so should be ok? @harrislapiroff any comments?

image

shanejdonnelly:
@SaptakS Yeah, I figured that if there was raw HTML like that, it was probably best to let it determine it's own positioning, etc. Targeting that with css is certainly doable, but I wasn't sure it was even desirable. (And also, given how this section is currently used on the live site, it seemed like possible overkill.)

harrislapiroff:
I agree—let's let the raw HTML determine its own position

SaptakS (approved):

PR #1537, opened 2022-11-16
Apply reversed coloring to children of aside

### LINKED ISSUE #1377: Aside background should hug text
As seen in figma:

Screen Shot 2022-04-05 at 3 49 22 PM

Current appearance:

image

shanejdonnelly:
This is a pretty simple fix, so perhaps I'm missing some edge cases. I did use `*` to target all children, as opposed to `> *` because that missed some text in development that had no tag. I think `> *` would be preferable, but they both do pretty much the same thing and probably better to protect against stray, random text. 

Screenshot from 2022-11-16 17-05-03

If you all know of places other than the blog to look for this pattern, please let me know! For now, hopefully this works well enough.
_____________________________________________

## Dev: 
(looks weird, but should be correct)  

Screenshot from 2022-11-16 17-11-58

http://localhost:8000/pft-blog/wrong-project-mother-respond-a-simple-successful-18/

_____________________________________________
## Prod: 
(same rule applied in Devtools) 

Screenshot from 2022-11-16 17-05-40
https://pressfreedomtracker.us/blog/celebrating-5-years-since-launch-of-the-us-press-freedom-tracker/

Screenshot from 2022-11-16 17-19-33

### DIFF
--- a/common/templatetags/common_tags.py
+++ b/common/templatetags/common_tags.py
@@ -1,6 +1,8 @@
 import bleach
+import hashlib
 from bs4 import BeautifulSoup
 from django import template
+from django.core.cache import cache
 from django.utils.html import mark_safe
 from wagtail.templatetags.wagtailcore_tags import richtext
 
@@ -31,6 +33,12 @@ def richtext_inline(value):
 def richtext_aside(value):
     "Returns HTML-formatted rich text with span injected in each block level element"
     html_str = richtext(value).__html__()
+
+    # Implicit cache invalidation happens based on the string
+    cache_key = hashlib.md5(html_str.encode('UTF-8')).hexdigest()
+    if cache_key in cache:
+        return cache.get(cache_key)
+
     soup = BeautifulSoup(html_str, 'html.parser')
     block_elems = soup.select('[data-block-key]')
     for elem in block_elems:
@@ -41,7 +49,10 @@ def richtext_aside(value):
             for content in reversed(elem.contents):
                 new_span.insert(0, content.extract())
             elem.append(new_span)
-    return mark_safe(str(soup))
+
+    aside_html = mark_safe(str(soup))
+    cache.set(cache_key, aside_html, 3600)
+    return aside_html
 
 
 @register.simple_tag
--- a/common/templatetags/common_tags.py
+++ b/common/templatetags/common_tags.py
@@ -1,4 +1,5 @@
 import bleach
+from bs4 import BeautifulSoup
 from django import template
 from django.utils.html import mark_safe
 from wagtail.templatetags.wagtailcore_tags import richtext
@@ -24,6 +25,23 @@ def richtext_inline(value):
             'a', 'abbr', 'acronym', 'b', 'code', 'em', 'i', 'strong', 'span'
         }
     ))
+
+
+@register.filter
+def richtext_aside(value):
+    "Returns HTML-formatted rich text with span injected in each block level element"
+    html_str = richtext(value).__html__()
+    soup = BeautifulSoup(html_str, 'html.parser')
+    block_elems = soup.select('[data-block-key]')
+    for elem in block_elems:
+        if elem.string:
+            elem.string.wrap(soup.new_tag('span'))
+        elif len(elem.contents):
+            new_span = soup.new_tag('span')
+            for content in reversed(elem.contents):
+                new_span.insert(0, content.extract())
+            elem.append(new_span)
+    return mark_safe(str(soup))
 
 
 @register.simple_tag
--- a/common/templatetags/common_tags.py
+++ b/common/templatetags/common_tags.py
@@ -35,7 +35,11 @@ def richtext_aside(value):
     html_str = richtext(value).__html__()
 
     # Implicit cache invalidation happens based on the string
-    cache_key = f"aside_html_cache_{hashlib.md5(html_str.encode('UTF-8')).hexdigest()}"  # nosec
+    html_str_hash = hashlib.md5(
+        html_str.encode('UTF-8'),
+        usedforsecurity=False,
+    ).hexdigest()
+    cache_key = f"aside_html_cache_{html_str_hash}"
     if cache_key in cache:
         return cache.get(cache_key)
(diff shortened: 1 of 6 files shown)


SaptakS:
I think it might be a little more complicated than this. So I created an aside block with a heading, followed by a paragraph (with different text formatting), followed by a list. This is how it looks:

image

This is definitely an improvement maybe compared to the one in the prod, but it doesn't seem to address the text "hugging" nature expected in the issue. This is because it adds background color to entire `<ul>` tag instead of individual `<li>` tags. Since we do have a particular set of formats that we allow inside an AsideBlock, maybe a better way would be to address each of those elements individually?

Also, I am wondering how much text hugging are we thinking? Because as we see in the image above, the black background doesn't end at the text end, because all those elements have a `block` display. We would then need to update them to be `inline-block` display. 

image
Though it still won't look fully text "hugging" (see screenshot) as mentioned in the issue, but I would think close enough. @harrislapiroff any suggestions about this? The more complicated solution would be to always wrap all the inner texts inside a span text for all elements (whether h2 or p or li or any other) and then apply the style only to the span text. We will need to update the template for the block for that. Only then I feel it can be exactly like the design.

harrislapiroff:
I do think I want the text hugging. I like your proposed method of injecting those inline spans @SaptakS, so I think the more complicated solution is what we want here.

SaptakS (changes requested):
Let's try https://github.com/freedomofpress/pressfreedomtracker.us/pull/1537#issuecomment-1318842496 for a more accurate implementation of the design.

harrislapiroff:
When this reaches the top of the ready for dev list @SaptakS will take over the code here to bring it to completion

SaptakS:
I did add test, not exactly sure how to get the coverage to be met for the cache hit.

chigby:
Two ideas for your coverage issue:

1. Set the cache in the test. This, to me, is not that different from doing something like adding a page to the DB to prepare a test.
2. If doing that feels too non-isolated for you, I could also see using a mock object for that cache that you can configure to return a mock hit or miss based on your needs. This would also let you make an assertion that `cache.set` was being called correctly on the "cache miss" scenario.

Either one of these, though, should get you the coverage. I'm happy to help more if you want, also. Sometimes the `mock` library can be tricky in practice.

SaptakS:
I was honestly also thinking of going the mock way. I feel like if I called with the same string twice, the test will ideally run through the cache and then the coverage may get passed since it will cover that critical path. Not sure what to test. Or like you said, I can manually set the cache and then maybe see if I get the same output once I pass the same argument to the function.

chigby:
For the test using mocks, the thing that I would assert, given the code we're working with, is that we are calling the right functions on the `cache` object (that we are mocking). So you set up your mock cache to have the correct key in it, then assert that the we called `cache.get`. We are doing some pretty similar things in the [test code for the mail system](https://github.com/freedomofpress/freedom.press/blob/develop/crm/tests/test_mailgun_integration.py#L96), for one example.

common/templatetags/common_tags.py
```diff
@@ -26,6 +29,34 @@ def richtext_inline(value):
     ))
 
 
+@register.filter
+def richtext_aside(value):
+    "Returns HTML-formatted rich text with span injected in each block level element"
+    html_str = richtext(value).__html__()
+
+    # Implicit cache invalidation happens based on the string
+    cache_key = f"aside_html_cache_{hashlib.md5(html_str.encode('UTF-8')).hexdigest()}"  # nosec
```
chigby: This is kind of a minor issue but I do think using `nosec` to ignore _all_ security alerts is possibly a bit too broad. I'd feel more comfortable with either ignoring the specific alert we're encountering with `nosec B324`, or using the hashlib argument `usedforsecurity=False`, as we have done elsewhere: 
https://github.com/freedomofpress/pressfreedomtracker.us/blob/57a8ea99a23f09cbd84ffb5b315064da1390eeb2/common/utils/mailchimp.py#L68-L71

SaptakS: I like the `usedforsecurity=False`. Makes sense to me.

common/tests/test_template_tags.py
```diff
@@ -41,3 +42,32 @@ def test_add_as_string_single_coercion_error_defaults_to_addition(self):
...
+        heading_aside_with_span = '<h3 data-block-key="89agh"><span>This is a heading</span></h3>'
+
+        self.assertEqual(richtext_aside(paragraph_aside), paragraph_aside_with_span)
+        self.assertEqual(richtext_aside(heading_aside), heading_aside_with_span)
+
+    @mock.patch('django.core.cache.cache.get')
+    def test_cache_usage_in_richtext_aside(self, mock_django_cache_get):
+        paragraph_aside = '<p data-block-key="89kgh">This is paragraph with <b>bold text</b></p>'
+        heading_aside = '<h3 data-block-key="89agh">This is a heading</h3>'
+        paragraph_aside_cache_key = f"aside_html_cache_{hashlib.md5(paragraph_aside.encode('UTF-8')).hexdigest()}"  # nosec
```
chigby: See above re: `nosec` comment.

common/tests/test_template_tags.py
```diff
@@ -41,3 +42,32 @@ def test_add_as_string_single_coercion_error_defaults_to_addition(self):
...
+
+        # Testing that calling with the different HTML string doesn't call cache.get
+        richtext_aside(heading_aside)
+        self.assertFalse(mock_django_cache_get.called)
+
+        # Testing that calling with the same HTML string calls the cache.get
+        richtext_aside(paragraph_aside)
+        self.assertTrue(mock_django_cache_get.called)
+        self.assertEqual(mock_django_cache_get.call_count, 1)
+        mock_django_cache_get.assert_called_once_with(paragraph_aside_cache_key)
```
chigby, line 77: This test looks great to me, thanks for adding it.

chigby (approved):

SaptakS (approved):
Approving (since I had originally requested changes) based on review by @chigby

PR #1541, opened 2022-11-22
Adds CSP rules explicitly for directives not covered by default-src

SaptakS:
Refs #https://github.com/freedomofpress/infrastructure/issues/3372


chigby (approved):
Looks great

PR #1544, opened 2022-11-28
Adds optional chaining for eventlistener when incidents are loaded

### LINKED ISSUE #1543: "Load more posts" button in the blog page is not working
On clicking the "Load more posts" button in https://pressfreedomtracker.us/blog/, it get's frozen in the loading state. Seems like there is some JS error that is happening in the console panel. 

image

SaptakS:
Fixes #1543

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,7 +20,8 @@ module.exports = {
 	"parserOptions": {
 		"ecmaFeatures": {
 			"experimentalObjectRestSpread": true
-		}
+		},
+		"ecmaVersion": 2020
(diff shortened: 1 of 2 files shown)


chigby (approved):
1. This fixes the problem with the blog index page (and the incident index page still works as before, so as far as I know the `ArticleLoader` is now functioning in both of the places we're using it), so that's great.
2. Adding a note here that a definitely support using the `?.` operator as it makes working with JS objects a lot less cumbersome. Has uses in almost any part of the front-end.

PR #1549, opened 2022-12-05
Fixes date formatting in filter sidebar

### LINKED ISSUE #1546: Filtering by date range causes error
If you enter a start date and an end date for an incident, the filter controls gray out and you get a JavaScript error:

Screenshot from 2022-11-30 08-46-24

SaptakS:
When applying both the max and min date filters, the one which is set before gets converted to the ISOString, causing errors when trying to convert to string again. This commit ensures that the min and max values in the DateFilter is always finally set to a Date format.

Fixes #1546

### DIFF
--- a/client/charts/js/components/FilterSet.jsx
+++ b/client/charts/js/components/FilterSet.jsx
@@ -42,7 +42,10 @@ function DateFilter({
 							type: SET_PARAMETER,
 							payload: {
 								filterName: name,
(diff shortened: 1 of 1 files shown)


chigby (approved):
This change looks good, and it also makes me think we ought to probably have some kind of filter sidebar test suite similar to the previous site design. I think we could probably get a basic setup going in React Testing Library (similar to what we have on freedom.press) without too much trouble.

PR #1550, opened 2022-12-05
Removes filtering from treemap clicks in homepage

### LINKED ISSUE #1547: Homepage treemap links incorrect
image

The tracker editors noticed that when you click on a category on the homepage treemap that it takes you to the category page filtered to only incidents from the last six months. This happens regardless of what date range you have selected.

I also notice the category page doesn't have any way of indicating active filters. I think it might make most sense to do away with this functionality altogether—simply have the treemap link to category pages with no applied filters and possibly even strip the code from category pages that handles filtering.

SaptakS:
Fixes #1547 

As mentioned in the issue, the treemaps always added 6 month filtering to the category page links. However the category page has no way of displaying active filters and it always will show only the 8 latest incident with the filter applied. So removing the filters from the treemap in the homepage.

### DIFF
--- a/client/charts/js/lib/utilities.js
+++ b/client/charts/js/lib/utilities.js
@@ -75,7 +75,7 @@ export function getFilteredUrl(databasePath, filtersApplied, currentDate) {
 		parameters.push(`state=${filtersApplied.state.replace(' ', '+')}`)
 	}
 
-	if (filtersApplied.year !== null && filtersApplied.monthName === undefined) {
+	if (filtersApplied.year !== null && filtersApplied.year !== undefined && filtersApplied.monthName === undefined) {
 		parameters.push(`date_lower=${filtersApplied.year}-01-01&date_upper=${filtersApplied.year}-12-31`)
 	}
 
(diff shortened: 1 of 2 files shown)


client/charts/js/lib/utilities.js
```diff
@@ -75,7 +75,7 @@ export function getFilteredUrl(databasePath, filtersApplied, currentDate) {
 		parameters.push(`state=${filtersApplied.state.replace(' ', '+')}`)
 	}
 
-	if (filtersApplied.year !== null && filtersApplied.monthName === undefined) {
+	if (filtersApplied.year !== null && filtersApplied.year !== undefined && filtersApplied.monthName === undefined) {
```
chigby, line 78: I think this is fine, but I do wonder if this whole function `getFilteredUrl` could do with some cleanup for all this null/undefined checking that's going on (something like iterating over fields that _do_ exist or having the `filtersApplied` object know how to convert itself to a URL and have the code live there... just some ideas for future enhancements).

chigby (approved):
This works for me! Merging.

client/charts/js/lib/utilities.js
SaptakS: Yes. I agree

PR #1559, opened 2022-12-20
Add attribution to aligned captioned image block template

### LINKED ISSUE #1558: Add image attribution to template for image blocks
Looks like we're not rendering image attribution on template blocks. Sometimes tracker editors just embed the attribution directly in the caption ([see](https://pressfreedomtracker.us/blog/charges-and-punishments-for-j6-rioters-who-hurt-journalists-damaged-news-equipment/)):

Screenshot 2022-12-20 at 11 56 25 AM
Screenshot 2022-12-20 at 11 56 30 AM

However, we are displaying attribution correctly for header images (on the same page):

image

Let's copy that behavior into the image block template.

chigby:
This PR adds some logic for showing the attribution of an image belonging to the Aligned Captioned Image Block. Minor changes made to CSS here to display that caption inline with the attribution with the correct line height.

Also adds the wagtail inventory package. **Note**: we will need to run `manage.py block_inventory` after this is deployed.

Fixes #1558 

### Screenshots

Top to bottom: Long caption with attribution, short caption with attribution, no caption.

image

image

### DIFF
--- a/common/templates/common/blocks/aligned_captioned_image.html
+++ b/common/templates/common/blocks/aligned_captioned_image.html
@@ -1,6 +1,14 @@
 <figure class="inline-media {{ value.alignment }}">
     {% include 'common/blocks/image_block.html' with value=value.image %}
-    {% if value.caption %}
-        <figcaption class="inline-media__caption">{{ value.caption }}</figcaption>
+    {% if value.caption or value.image.attribution %}
+        <figcaption class="inline-media__caption">
+			{% if value.caption %}{{ value.caption }}{% endif %}
+			{% if value.image.attribution %}
+				<span
+					class="media-attribution"
+					aria-label="Attributed to: {{ value.image.attribution }}"> — {{ value.image.attribution }}
+				</span>
+			{% endif %}
+		</figcaption>
     {% endif %}
 </figure>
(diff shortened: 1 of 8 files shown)


common/templates/common/blocks/aligned_captioned_image.html
```diff
@@ -1,6 +1,14 @@
...
     {% include 'common/blocks/image_block.html' with value=value.image %}
-    {% if value.caption %}
-        <figcaption class="inline-media__caption">{{ value.caption }}</figcaption>
+    {% if value.caption or value.image.attribution %}
+        <figcaption class="inline-media__caption">
+			{% if value.caption %}{{ value.caption }}{% endif %}
+			{% if value.image.attribution %}
+				<span
+					class="media-attribution"
+					aria-label="Attributed to: {{ value.image.attribution }}"> — {{ value.image.attribution }}
```
SaptakS: I don't think an `aria-label` makes sense in a span, since it is not interactive and many screen readers ignore it because it has no interactive role. I feel the figcaption giving the accessible name to the image will also ignore the aria-label part. If we really want to mention "Attributed to: " then we can place that text inside another span with `.sr-only`

SaptakS (commented):
I left one comment related to the aria-label. I also saw that in the issue, @harrislapiroff mentioned that the editors often put the attribution in the captions itself. In that case, is the attribution supposed be shown twice? Not sure what the expected behavious for that is.

harrislapiroff:
Someone will manually go through and remove in from the caption in any instances where it’s been done that way, unless someone has a cleverer idea. (Hence the inclusion of Wagtail inventory.)

SaptakS (approved):

PR #1569, opened 2023-01-11
Remove public autocomplete URLs

### LINKED ISSUE #1568: Remove public autocomplete URL routes
These were used in the pre-redesign version of the site when the public-facing search UI relied on these autocomplete widgets and the corresponding backend. The new filter design does not use this code at all, and it should be safe to remove.

Inspired by getting [error messages on the URL](https://logs.freedom.press/app/kibana#/discover?_g=(time:(from:now-7d,to:now))&_a=(index:'filebeat-*',query:(language:kuery,query:'_id:IZTDnoUB__PuiT3nt1Qr'))) due to malformed data.

chigby:
Fixes #1568

This functionality is part of the old site design and no longer needs to be exposed to the public.


SaptakS (approved):
looks pretty straightforward and good to me

PR #1578, opened 2023-02-13
Updates ipython

SaptakS:
Versions prior to 8.1.0 are subject to a command injection vulnerability with very specific prerequisites. This vulnerability requires that the function 'IPython.utils.terminal.set_term_title' be called on Windows in a Python environment where ctypes is not available. The dependency on 'ctypes' in 'IPython.utils._process_win32' prevents the vulnerable code from ever being reached in the ipython binary. However, as a library that could be used by another tool 'set_term_title' could be called and hence introduce a vulnerability. Should an attacker get untrusted input to an instance of this function they would be able to inject shell commands as current process and limited to the scope of the current process. Users of ipython as a library are advised to upgrade. Users unable to upgrade should ensure that any calls to the 'IPython.utils.terminal.set_term_title' function are done with trusted or filtered input. (id: 53269)

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -469,9 +469,9 @@ inflection==0.5.1 \
 ipdb==0.13.9 \
     --hash=sha256:951bd9a64731c444fd907a5ce268543020086a697f6be08f7cc2c9a752a278c5
     # via -r dev-requirements.in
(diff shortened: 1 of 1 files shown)


chigby (approved):
Update looks good!

PR #1583, opened 2023-03-02
Updates json5 vulnerability

SaptakS:

PR #1584, opened 2023-03-07
Update NPM dependencies

### LINKED ISSUE #1469: Update smaller npm dependencies
Update all the minor JS dependencies (e.g., not React and react related packages) to their latest version. One can check the outdated npm dependencies using: npm outdated. If it's a minor release, one can just update using npm updata <package-name>. If it's a major release update, then one needs to update the package.json and then run npm install (or npm i --package-lock-only if you don't want to install).

### LINKED ISSUE #1470: Remove unused npm dependencies
We are probably not using this packages, especially after the redesign, so might be a good time to get rid of them.

- [x] Decimal.js
- [x] hamburgers
- [x] js-cookie : ensure it's not being used anywhere to read CSRF tokens
- [x] normalize-css and normalize.css

chigby:
Fixes #1469 
Fixes #1470

This PR removes a number of unused (or barely used) npm packages, and updates everything else *except* the `stylelint` related packages because the next version of stylelint [deprecates a number of rules that we are actually using](https://github.com/stylelint/stylelint/blob/main/docs/migration-guide/to-15.md#deprecated-stylistic-rules) and I'm not exactly sure what to do about that. We've had some problems with stylelint in the past being not-quite compatible with the Sass indented syntax and I think tinkering a lot of with this is outside the scope of these issues. 

Depends on #1583 and should only be merged after it.


client/common/js/ArticleLoader.js
```diff
@@ -138,20 +137,23 @@ class ArticleLoader {
 		this.isLoading = true
 		this.nextLinkElm.innerHTML = '<div class="loader">Loading…</div>'
 
-		axios.get(this.nextLinkElm.href)
+		fetch(this.nextLinkElm.href)
```
harrislapiroff, line 140: So pleased to finally give axios the axe

docker-compose.yaml
```diff
@@ -67,6 +67,7 @@ services:
       DEPLOY_ENV: dev
       DJANGO_PROFILE: "${DJANGO_PROFILE:-no}"
       SELENIUM_HOST: selenium
+      DJANGO_TEST_DB_NAME:
```
harrislapiroff: 👀

harrislapiroff (approved):
LGTM, though I'm not sure I understand what the deal is with the `DJANGO_TEST_DB_NAME` var?

PR #1595, opened 2023-03-20
Templates for newsletter and special blog posts

### LINKED ISSUE #1480: Blog post templates
Let's add the different templates for different blog post categories as documented in Accurat's Figma:

- [ ] [Newsletter](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3698%3A62440)
- [ ] [Special Section](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3698%3A69119)

I leave it up to the implementer whether these should be implemented as separate page types or as a selection on the existing page type. Please review the mockups and write an architecture proposal as a comment on this ticket before beginning implementation.

A note:

- Newsletters should support rendering an interactive treemap with filters selected by the content editor as the lead graphic.
- However, the chart block type shown mid-article in the figma should be saved for another ticket.

SaptakS:
Fixes #1480 

The things addressed in this PR:
- Separates the blog templates into 3 different includes: `_default_blog.html`, `_newsletter.html`, `_special_blog.html`
- Have a dropdown in admin to select the type of blog: Default, Newsletter, Special. This dropdown is in the settings panel. If someone feels it should be somewhere else, let me know.
- Depending on the type selected by the editor, we chose in blog_page.html which of the 3 templates to include.
- Adds SiteSettings for donation page and newsletter subscription page (Used in newsletter templates)
- Adds ButtonBlock in the body of the blog model so that "Related incidents" buttons can be added to the body of the newsletters as shown in the design provided
- Adds filtering by type in the blog index page. This is currently used only in the individual newsletter pages, for the button "Go to archived editions", which filters by `type=newsletter`. But all the other types (`default`, `special` can also be used by updating the GET parameters in the URL
- Adds absolute positioned hero image for the special blog templates (everything else was same as default template). Here, I did take few decisions, so let me know if all these makes sense:
  - I have made the lead image of the blog post as the absolute positioned (full width background image) image
  - while doing the above, I noticed that the `.header` and all our buttons have a white background. This didn't work for this template, so I made them `transparent`, instead of `white`. I checked a bit and didn't seem to break anything, so if anyone knows a reason why this might be problematic, let me know.
  - I added a gradient on top of the image which is darker below, and quite white above. This is my desperate attempt at hoping that in most scenarios, the text on top of the image is accessible, but it will still depend on the editors and the image they select. So might be something to draw attention to the editor. Also, in case someone feels the gradient needs to be adjusted, let me know (attn @harrislapiroff )
  - The author and publication dates are shown on top of this background, and the attribution for the image as well. Since this is a background image, I am not sure what to do with caption. I do want to add a caption, but also, a screen reader user will never encounter the image (just an empty div). I was thinking of adding an `.sr-only`  caption to indicate we do have an image and this is what the image shows. Not sure if there is some other way we should be using the caption.
- The only thing I have skipped in this PR, that I initially mentioned in the issue about implementing is the "In the Database" section. The default template and the special blog template have a "In the Database" section at the bottom of the page in the design. But I was not very clear what incidents we really want to show here. I think this can be implemented as a separate PR as well.
- The special blog template design has some dated sections in the body, I am not really sure if they are supposed to be updates or something else. Right now, the body is just like default blog body. (attn @harrislapiroff )
 
Screenshot of an example newsletter:
image

Screenshot of an example special blog post:
image

@chigby I would be interested to know what kind of tests you think I should add for these, since most of them are visual changes. I am however thinking of updating createdevdata to add some newsletter and some special blog templates.

PS: This doesn't add or change how the visualization works in newsletter. It's still treated an image shown exactly how it is in default template.

Another PS: This PR doesn't change the blog index design. Accurat did provide us a fancier design for the index page which changes the layout for special blog listing. I think can be done as a separate PR.

Lastly, the migration of the blog posts to different type either needs to be manual, I think. Though deploying this won't break anything, just all posts will continue to be shown in default template.


chigby:
Typically I would say visual changes and styles don't really need automated tests. There are some cases where if a django template has a lot of logic, conditionals, or behavior that varies a lot based on the underlying data, then maybe there should be some tests for that. Something like set up data, request page using the test client, then assert that the rendered HTML has certain features. But glancing at this PR it doesn't really seem like that's the case here for most of these templates.

And of course I always support adding representative content to the createdevdata command :slightly_smiling_face:

harrislapiroff (changes requested):
This looks good. We can continue to refine as we see it used.

Thanks for noting the button bg color change. I think maybe we should add a special case for the transparent buttons. Making all buttons transparent does slightly change the design, e.g., on the homepage where some buttons are on a gray background. (It will also be relevant for buttons on the gray Incidents Database sidebar, though we don't have any there right now that are in Accurat's mockups.)

I'm fine with a `.sr-only` element for image caption. I also would take using an `` element for the header image (that might have some other advantages, like offloading cropping to wagtail, which has knowledge of focus aread) but making it a background image is fine too.

Before merge:

- [x] Update button color back to while
- [x] Make special transparent button for use in special post header
- [x] Add .sr-only caption for images or make header image img elements

Let's also open some tickets after the fact for changes to the blog index page:

- [x] Add a special preview for special sections
- [x] Add a UI for looking at posts in a particular category

SaptakS:
>  Add a special preview for special sections

Not sure what you mean by this. Do you mean the preview in the blog index page as mentioned by Accurat?

> Add a UI for looking at posts in a particular category

Right now, the filter does work, just the UI is same for all categories. Do we have designs for listing of blog posts in separate category?

harrislapiroff:
> Not sure what you mean by this. Do you mean the preview in the blog index page as mentioned by Accurat?

I think we're talking about the same thing? "Teaser" might have been a better word for me to use here 😅. I'm referring to this:

image

> Right now, the filter does work, just the UI is same for all categories. Do we have designs for listing of blog posts in separate category?

Yes, Accurat designed this tab set for swapping between them

image

[Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3689-54926&t=q92LFBtIBwAa4rUz-4)

SaptakS:
Got it. Thanks for clarifying.

SaptakS:
Updated the PR to include your suggestions. Also created [separate issue](https://github.com/freedomofpress/pressfreedomtracker.us/issues/1617) for the index page related updates.

blog/templates/blog/blog_page.html
```diff
@@ -1,5 +1,11 @@
 {% extends "base.html" %}
 
+{% if page.blog_type == 'special' %}
+	{% block menu %}
+		{% include "common/_menu.html" with btn_class="transparent" %}
+	{% endblock %}
+{% endif %}
+
```
harrislapiroff: This doesn't actually work, as per the note at the [bottom of the template inheritance docs](https://docs.djangoproject.com/en/4.2/ref/templates/language/#template-inheritance):

> [`{% block %}`](https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#std-templatetag-block) tags are evaluated first. That’s why the content of a block is always overridden, regardless of the truthiness of surrounding tags. For example, this template will always override the content of the title block:
>
> ```
> {% if change_title %}
>    {% block title %}Hello!{% endblock title %}
> {% endif %}
> ```

It doesn't cause any visible problems (it's actually fine having transparent buttons on other blog posts as well) but maybe for legibility we should clean it up one way or another.

harrislapiroff (approved):
Looks generally good. One small fix and then we're good to go!

PR #1600, opened 2023-03-21
Add Troubleshooting Doc

dkaoster:
Current working PR as I note the things I run into while getting started.

I noticed that many docs in this repo are `.rst` files instead of `.md`, let me know if I should also making this troubleshooting doc a `.rst`.


chigby (approved):
This is a great idea, and I don't think .md is a problem. It's possible our readme is in RST for historical reasons, I feel like I've heard rumblings about migrating it to MD sometime.

PR #1601, opened 2023-03-21
Change Color of "Assault" Category

### LINKED ISSUE #1598: Change homepage chart colors
One source of confusion that we've observed with users is around the color on the homepage charts:

image

In the treemap on the left, orange represents the assaults category. But on the center and right it represents all incidents. Sometimes users get confused and think it represents only assaults across all three charts.

This ticket is to develop and implement a way of handling colors in these charts that avoids this confusion (e.g., by making the map and bar charts B&W or by removing orange from the treemap).

You can review the charts color palette in this [Figma document](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=1893-9644&t=eKBR6PpiYEMvkInA-4) provided to us by the designers. This color palette is a guideline, not a rule—if your design solution introduces a new color, that should be fine, as long as it looks good.

dkaoster:
Closes #1598

### DIFF
--- a/client/charts/js/components/BarChart.jsx
+++ b/client/charts/js/components/BarChart.jsx
@@ -44,7 +44,6 @@ export default function BarChart({
 	id = '',
 	numberOfTicks = 4,
 	openSearchPage,
-	highlightColor = '#E6E6E6',
 }) {
 	if (!data.length) return null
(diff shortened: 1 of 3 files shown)


SaptakS (approved):
Looks good to me.

PR #1604, opened 2023-03-23
Update Incident Cards Design

### LINKED ISSUE #1387: Homepage incident cards should have more compact table as shown in Figma
Homepage incident cards when seen in mobile and tab screens don't match the design with the one in figma.

dkaoster:
closes #1387 

~Fully html + css solution, no js needed.~

Progressive enhancement such that without js, the card is in the "expanded" state without a toggle button, and with js the card initializes to "closed" with a button to "show more".


SaptakS (changes requested):
I used to be a huge supporter of the checkbox hack and CSS-only widgets (and have made multiple PRs using that in past), but over the years I have changed that opinion. The checkbox hack for collapsible items has lot of accessibility issues. This article demonstrates some of them: https://adrianroselli.com/2023/03/css-only-widgets-are-inaccessible.html

Instead of using the checkbox hack for a CSS-only implementation, I would propose a progressive enhancement approach. Let's implement the card design such that the "Show More +" link is not shown for non-js users, and the card is expanded for the non-js users always. In the JS code, make the "Show More +" link visible and hide the expanded content, and then implement the toggle using JS. This way, for non-js users, they just always see the entire card, instead of a collapsible card. For users with JS enabled, they get the collapsible experience. Does that make sense?

dkaoster:
@SaptakS That makes sense, never knew that the checkbox method was bad for accessibility, thanks for the help.

client/common/sass/components/_incident-card.sass
```diff
@@ -3,19 +3,76 @@
...
+				padding: 0
+
+			&__toggle-icon
+				width: 0.8rem
+				height: 0.8rem
+				background-size: contain
+				background-color: transparent
+				background-image: url(../icons/plus.svg)
+				background-repeat: no-repeat
+				margin-left: 1rem
```
SaptakS: Based on the figma design, I think this should be 0.5rem.

client/common/js/collapsibleCard.js
```diff
@@ -0,0 +1,30 @@
+function collapsibleCard(card) {
+	function toggleCard() {
+		if (card.classList.contains('closed')) {
+			card.classList.remove('closed')
+			card.classList.add('open')
+			card.querySelectorAll('.incident-card-collapsible__toggle-button-text')
+				// eslint-disable-next-line no-param-reassign
+				.forEach((buttonText) => { buttonText.innerText = 'Show Less' })
```
SaptakS: I think we also need to update the toggle  icon from a plus to a cross and vice-versa in the other scenario. Also, do we really need to `querySelectorAll`? I don't think there will ever be a case where there will be more than one "Show more" button, and that way we can avoid the `forEach`?

SaptakS (changes requested):
This PR works great! Tested both JS disabled state and also checked that it doesn't break the other incident card designs such as the database cards. I requested 2 minor changes. Once they are addressed, I think this PR is ready to be merged.

SaptakS (approved):

PR #1605, opened 2023-03-25
Improve carousel behavior

### LINKED ISSUE #1415: Restructure caruousel code
We use these article grids that turn into carousels on mobile in a couple places on the site. The current code behavior for them isn't as consistent as I'd like and doesn't preserve normal scrolling behavior. It's important that the javascript here also be a "progressive enhancement"—that with javascript disabled you simply get native horizontal scrolling.

image 

The new carousel should:

- [ ] Be scrollable using touch-drag/scroll wheel
- [ ] Use ⬅️ ➡️ buttons as a secondary method of scrolling, e.g., "scroll to next article"
- [ ] Deactivate and activate buttons automatically according to scrollability

See comments here: https://github.com/freedomofpress/pressfreedomtracker.us/pull/1406#issuecomment-1101565958

dkaoster:
closes #1415

### DIFF
--- a/client/common/js/articleScroller.js
+++ b/client/common/js/articleScroller.js
@@ -37,7 +37,7 @@ class ArticleScroller {
 	}
 
 	shiftArticles(shift) {
-		this.articleParent.style.transform = 'translateX(' + shift + ')'
+		this.articleParent.scrollLeft = shift
 	}
 
 	prevExists() {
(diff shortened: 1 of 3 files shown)


SaptakS (approved):
Looks great! The cards scroll to the proper one while using keyboard navigation as well. So I am happy about that. Merging.

PR #1606, opened 2023-03-28
Track Matomo Events on Filter

### LINKED ISSUE #1599: Analytics on filters sidebar and search bar
I want to have better data on how users are navigating the site. Using [Matomo](https://matomo.org/) I can see requests like:

`/all-incidents/?categories=10&date_lower=2022-01-01&date_upper=2022-12-31`

which appear to show me that a user has used the Categories filter and data filters. But in actuality, they probably arrived at that URL by clicking the 2022 bar of the bar chart on the Assaults page:

image

I want to be able to better distinguish between that sort of interaction and a user actively using the sidebar filters:

image

This ticket includes developing and implementing a system to do so. Please do some quick research to figure out the best approach. I suspect [Matomo event tracking](https://matomo.org/guide/reports/event-tracking/) will be useful here, but if you see a better approach go for it. (Don't overthink it, the priority is to spend minimal time on this so we can get to looking at the metrics.)

Add a comment to this PR describing your planned approach and get my approval before implementing.

dkaoster:
closes #1599 

Adds specific filter tracking to matomo by [sending events](https://developer.matomo.org/guides/tracking-javascript-guide#manually-trigger-events).

## Testing
- Set `ANALYTICS_ENABLED` to `True` in [base.py](https://github.com/freedomofpress/pressfreedomtracker.us/blob/develop/tracker/settings/base.py#L219) and run the app
- Open the [incidents database locally](http://localhost:8000/all-incidents/) and set a few filters
- The network tab should show the tracking requests


client/charts/js/components/TagFilter.jsx
```diff
@@ -87,6 +88,8 @@ export default function TagFilter({
 									item: tag,
 								}
 							})
+							// Fire matomo event
+							trackMatomoEvent(['Filter', 'Tags', 'Toggle', tag.name])
```
chigby: I think the `tag` variable here is a simple string (rather than an object), so we can just track that value directly. At least, when I test this out by checking and unchecking the tag boxes I'm seeing this function called with `undefined` in that last position.

client/charts/js/components/TagFilter.jsx
```diff
@@ -110,6 +113,8 @@ export default function TagFilter({
 							item: tag,
 						}
 					})
+					// Fire matomo event
+					trackMatomoEvent(['Filter', 'Tags', 'Select', tag.name])
```
chigby: See above regarding `tag` vs. `tag.name`

chigby (changes requested):
This is looking really good, though I did mark a couple places where a variable needs a small change. 

Also, I think adding a call to the `trackMotomoEvent` function in one other place would catch a lot more activity. Specifically any changes to controls happening _within_ the expanded category filter groups. Here's that place: https://github.com/freedomofpress/pressfreedomtracker.us/blob/sidebar-analytics/client/charts/js/components/CategoryFilter.jsx#L126-L132 -- let me know if I should provide more information about this.

SaptakS:
I just wanted to bring to attention that these changes may need us to update our privacy policy before it is deployed so that users are informed that they are actions are getting tracked. Also, usually a privacy policy change should be informed to the users on their first visit after the change. Right now, I think our privacy policy is more about a web page requests and not related to what we track for users interaction inside a page. Not sure if we have any plans. cc @harrislapiroff

harrislapiroff:
Checked in with @eloquence on updating our privacy policy—will wait to hear back.

eloquence:
I agree with Saptak that we should a sentence to the privacy policy prior to deploying additional tracking (I would think in the "Analytics" section). With recommended language, we can ask Harlo and Trevor to make a separate determination on how to communicate the change.

Right now the policy states:

> FPF's internal analytical logging involves logging, for up to 30 days, two bytes of the IP address, as well as the referrer page, time stamp, page requested, user agent, language header, website visited, and a hash of all of this information. 

What specifically should be added to this list, and what is the intended retention of this data?

eloquence:
Discussed a bit at Web Sync today, a couple of additional things to note:
- We have two copies, one at https://freedom.press/privacy-policy/ and one at https://pressfreedomtracker.us/privacy-policy/ (securedrop.org links back to freedom.press, and dangerzone.rocks currently does not link to any privacy policy)
- The policy has explicit language regarding notice requirements:

> FPF's Privacy Policy may change from time to time. However, any revised privacy policy will be consistent with FPF's mission. If we make any substantive changes to our policies, we will place notice in our newsletter and post notice of changes on this page.

I would argue in favor of interpreting this as a substantive change, in part because it'll help us go through the proper process at least once.
- Saptak and others are welcome to add further thoughts or proposed language here; I defer to Harris on final language to propose to Harlo/Trevor.

harrislapiroff:
Whoops I missed that the next step here is for me to come up with proposed language for this change. I'll get that this week

harrislapiroff:
Okay, at long last here is my proposal. Sorry for the long delay on this—I think I was overthinking it:

> FPF's internal analytical logging involves logging, for up to 30 days, two bytes of the IP address, as well as the referrer page, time stamp, page requested, user agent, language header, website visited, **actions taken while on the page (such as selecting particular search filters or clicking particular buttons),** and a hash of all of this information.

If we wanted to be extra diligent we _could_ keep a list of the events that we record in the policy. Do we think that would be beneficial?

One thing we could do with the privacy policy is just keep a running changelog at the bottom. I think that would fulfill the notice in the policy element and would also be wonderfully transparent.

I _believe_ these events are governed by the same data retention settings as the rest of our matomo data, but I can double check.

harrislapiroff:
I've converted this back to draf to indicate that we're not ready to merge it until we iron out the privacy policy elements. @eloquence the language above is ready for your review and for kicking to Harlo and Trevor when you think it's ready.

eloquence:
I think the language you suggested is good. I don't think we need to go into more detail than that, but will discuss w/ Harlo. 

I agree that a changelog in the policy should meet our notice requirement for the policy page itself. However, we also mention the newsletter, so we should (to adhere to that commitment) draft some language for that as well.

I'll make an ad-hoc channel on Slack to discuss this asynchronously.

harrislapiroff:
We have updated the Privacy Policy across our sites to allow this sort of analytics collection. Unfortunately this PR has gotten somewhat out of date in the meantime, so kicking back to @dkaoster to get it back up to date for merge (Sorry! Thank you!)

dkaoster:
Added some testing steps, this should be ready for review!

chigby:
I think this is probably good to go, though I have one question: while doing one final round of testing this out I noticed that on `TextFilter` components, the track event function is called for every time the text input field changes, so if you're typing something you get an update for each keystroke. If that's something we're okay with for launching this feature, that's fine, but it might be nice to put some timer or debouncing logic on that to make the events that are ultimately sent to Matomo less noisy. 

On the other hand, we don't have all that many text filters on the site right now, so it probably won't be a huge problem to go live with that. Let me know what you all think @dkaoster @harrislapiroff

dkaoster:
@chigby I think debouncing makes a lot of sense, I can go ahead and add that in.

chigby (approved):

chigby:
The debouncing seems to work great, and I think the failing prod check will be resolved in #1735.

PR #1610, opened 2023-04-01
Use hashed css and js build files to prevent cache during local development

dkaoster:
I noticed that I was once again getting cached when doing local development, so I decided to dig a little deeper and found that when the static files built by webpack are being served by django, the static files do get cached in the browser. [See this ticket on stackoverflow for an explanation](https://stackoverflow.com/questions/7013735/turn-off-caching-of-static-files-in-django-development-server).

This is what I was seeing locally:

Screenshot 2023-03-31 at 4 18 40 PM

Now this is probably the wrong way to turn off the static file caching in django as it will probably affect production environments, but I wanted to just put this solution out there to see if folks had any suggestions for a better way to do this.

**Edit:** Decided to go with a solution of using webpack to also generated a hashed `common.js` and `common.css` files in development mode so that we still have client side caching in local development environments.


SaptakS:
I am personally always mindful of doing a hard-refresh (CTRL+SHIFT+R) whenever making a static file change. And I mean not just for django application, but always. I really consider client side caching as a feature, not a bug. But I can see how developing might be little frustrating unless one is just used to hard-refreshing everything 😅 .  So I am bit torn on turning off client side caching.

harrislapiroff:
Maybe an alternative approach would be to get webpack to add a hash to the bundle name? That way new bundles would reload every time they were regenerated

SaptakS:
That's kind of what happens in the prod side, i.e., the bundled js files and css files created have a hash. That ensures that after a deploy, the users get the new design I think. So I am fine with that approach.

dkaoster:
Unfortunately even with hard refreshes I've noticed things still get cached, and the only way I've (so far) been able to consistently get it to not cache is to turn this setting off in django. I see a couple options here:

1. I can try adding a hash as @harrislapiroff is suggesting, so that `common.js` and `common.css` are something like `common.xxxxxx.js` and `common.yyyyyy.css`? I do fear that that might become a rabbit hole of hunting down how these files are linked up and making sure that they update all the way through the resource request chain.

2. I'm wondering if turning off client side caching in django will even matter in production environments. If production builds are already creating filename hashes and/or have a CDN to manage the cache rules, turning it off in django should only affect development environments, which is why I think this method should be the best option. I just wanted to make sure that I wasn't misunderstanding anything about the production environments here.

dkaoster:
@harrislapiroff @SaptakS @chigby Just tried the hashing method and the good news is that it is really straightforward and seems to work pretty well!

chigby (approved):
Glad that this is working for you! :sparkles:

PR #1612, opened 2023-04-03
Fix Chart Page

dkaoster:
While starting work on charts, I noticed that the `/charts` page was broken, this adds some checks to prevent things from breaking on load.


styleguide/templates/styleguide/index.html
```diff
@@ -338,6 +338,7 @@ <h4 class="heading-regular">Key-Value Details</h4>
 			If a new details table is needed for a specific page, create a new include
 			template file, following above structure
 		</p>
+	</details>
```
dkaoster, line 360: separate from this, but noticed that the styleguide page was also missing a closing `</details>` tag.

chigby (approved):
Everything seems to look good here, let's merge it.

PR #1614, opened 2023-04-05
Vertical Bar Chart

### LINKED ISSUE #1608: Embed in blog posts a vertical bar chart showing counts of incidents over time
Add a vertical bar chart stream block.

- [x] Allow content editors to select which data to use for the bar chart (OK to limit to category, tag, and date range filters for now)
- [x] Render a vertical bar chart in the blog post (the x-axis should be date, the y-axis should be incident count)
- [x] **Stretch goal:** A screenreader should produce reasonable audio of the chart. Ideally this would be a readout of each bar label and value. If it's not practical to complete this with the current version of `react-animated-dataset` let's open a separate ticket to track it in the future.

[Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=2031-5227&t=i2IiB2EU1RUwVS5u-4)

dkaoster:
closes #1608


common/tests/utils.py
```diff
@@ -337,4 +354,4 @@ def streamfield(self, fields=None):
             else:
                 raise Exception('unknown field: {}'.format(field))
 
-        return json.dumps(streamfield_data)
+        return streamfield_data  # json.dumps(streamfield_data)
```
harrislapiroff: If we definitely don't need to convert this to JSON, let's remove this comment

SaptakS:
@chigby @dkaoster there seems to be a merge conflict. Can either of you look into that?

chigby:
@SaptakS Should be fixed now

common/blocks.py
```diff
@@ -434,3 +435,45 @@ class Meta:
...
+    )
+    upper_date = blocks.DateBlock(
+        label='Filter by Date, upper bound',
+        required=False,
+        help_text='If set, no incidents after this date will be included.',
+    )
+
+
+class VerticalBarChart(blocks.StructBlock):
+    title = blocks.CharBlock(required=True)
```
harrislapiroff: I think title should be optional and just not be rendered if blank.

common/blocks.py
```diff
@@ -434,3 +435,45 @@ class Meta:
...
+    incident_set = SimpleIncidentSet()
+    description = blocks.TextBlock(
+        required=True,
+        help_text='Summary of the chart for people using assistive technology, '
+        'such as screen readers.',
```
harrislapiroff: Let's add a little additional guidance here:

> Description for assistive technology users. If the chart is demonstrating a specific trend, try to include that, e.g., "Bar chart showing a decreasing number of assaults over the course of 2023."

(This is just a general idea of the type of guidance I'd like to provide—if you have better phrasing feel free to use it.)

harrislapiroff (changes requested):
A couple changes in the comments to request and one more here:

image

Can we get some smarter tick mark formatting? I think I'd like to see logic something like this (feel free to suggest alternatives):

1. If all bars are from a single year, use just a month abbreviation, e.g., "Jan" (as on the homepage bar chart)
2. If bars represent months from multiple years, use a month abbreviation, except for replace the first tick and Jan with the year: e.g., for an x-axis spanning oct 22 to sep 23 "2022, Nov, Dec, 2023, Feb, Mar, Apr, May, Jun, Jul, Aug, Sep"
3. I couldn't tell from my testing, but does it switch to breaking down by year instead of month at some point? If not I think we should—maybe when the date range is more than 2 years? And then tick marks will be just the full year, e.g., "2021, 2022, 2023"

Otherwise this looks really good and I'm very excited to present it to our editors! They're gonna love it.

chigby:
@harrislapiroff @dkaoster Just pushed some changes to this PR that address all of Harris's changes except for the tick marks.

harrislapiroff (approved):
The updates here look great! Thanks, all! Let's get it in! 😁

harrislapiroff (changes requested):
Shoot, sorry, kicking the tires a little more, I noticed the chart isn't sufficiently responsive. Can we get some work on the chart for mobile before we merge @dkaoster? If it could have the slider selector the homepage chart has too, that would be cool, but is not a necessity—we can leave that for a future PR.

dkaoster:
@harrislapiroff that should do it!

harrislapiroff (approved):
Amazing! ✨

PR #1616, opened 2023-04-10
Adds Access-Control-Allow-Origin headers to all the API endpoints

### LINKED ISSUE #1597: Access-Control-Allow-Origin not set on non-incident endpoints
Currently we have `Access-Control-Allow-Origin: *` on the [Incidents endpoint](https://pressfreedomtracker.us/api/edge/incidents/) but not others, e.g., the [journalists endpoint](https://pressfreedomtracker.us/api/edge/journalists/). We should set this across the board to allow people to query our API directly in their web applications if they so choose.

SaptakS:
Resolves #1597 

This PR adds `Access-Control-Allow-Origin: *` and `Access-Control-Allow-Methods: GET,OPTIONS,HEAD` to all the API endpoints. I have done it in the same way that it's implemented in incidents API endpoint.

### DIFF
--- a/incident/api/views.py
+++ b/incident/api/views.py
@@ -197,42 +197,123 @@ class JournalistViewSet(viewsets.ReadOnlyModelViewSet):
     queryset = models.Journalist.objects.all()
     serializer_class = ItemSerializer
 
+    def dispatch(self, *args, **kwargs) -> 'HttpResponse':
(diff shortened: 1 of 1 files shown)


chigby (approved):
This all seems like exactly what it's supposed to be.

PR #1618, opened 2023-04-12
Remove unnecessary computation from Incidents Database page

chigby:
This PR 

1. Removes a call computing the incidents summary from the `IncidentIndexPage.get_context()` method. We aren't using this data anywhere (as far as I can tell) on that page, and this is just wasting time.
2. Updates `django-silk` to the latest version
3. Adds `pyinstrument` (https://pypi.org/project/pyinstrument/) for additional profiling help. It's enabled separately from silk because I didn't want them interfering with each other (not sure if this is a problem, but I'm trying to be careful here).

### DIFF
--- a/incident/models/incident_index_page.py
+++ b/incident/models/incident_index_page.py
@@ -228,7 +228,6 @@ class IncidentIndexPage(RoutablePageMixin, MetadataPageMixin, Page):
 
         context['entries_page'] = entries
         context['paginator'] = paginator
-        context['summary_table'] = incident_filter.get_summary()
 
         get_data = request.GET.copy()
         context['sort_choices'] = []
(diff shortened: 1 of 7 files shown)


incident/models/incident_index_page.py
```diff
@@ -228,7 +228,6 @@ def get_context(self, request, *args, **kwargs):
 
         context['entries_page'] = entries
         context['paginator'] = paginator
-        context['summary_table'] = incident_filter.get_summary()
```
SaptakS, line 231: I think this can be removed from `CategoryPage` as well? Though doesn't affect the performance of the page much I guess.

chigby: Yes, I think it can be removed, I'll do it

SaptakS (approved):

PR #1620, opened 2023-04-13
Update Categories API

### LINKED ISSUE #1337: [API] Category endpoint improvements
- [x] Add URL to category page
- [ ] Change symbol ID to a URL to the SVG instead
- [ ] Add colors (as specified in the homepage charts—maybe this will require adding back a category color setting in wagtail)

Consider:

- [ ] Add aggregate incident counts
- [ ] Add methodology items

### LINKED ISSUE #1338: Update categories API; Update javascript to use API
Currently category data is hardcoded into a couple places in the charts javascript. We should generate this on the server side and feed it the javascript. We could do this in a django template or via API after #1337 

https://github.com/freedomofpress/pressfreedomtracker.us/blob/redesign/client/charts/js/components/HomepageMainCharts.jsx#L18

https://github.com/freedomofpress/pressfreedomtracker.us/blob/redesign/client/charts/js/lib/utilities.js#L186

dkaoster:
closes #1337, closes #1338


client/charts/js/components/HomepageSelection.jsx
```diff
@@ -39,7 +39,9 @@ export default function HomepageSelection({
 }) {
 	const years = d3
 		.groups(
-			originalDataset.map((d) => ({ year: d.date.getUTCFullYear() })),
+			originalDataset.map((d) => ({
+				year: (typeof d.date === "string" ? d3.timeParse("%Y-%m-%d")(d.date) : d.date).getUTCFullYear()
```
dkaoster: Ran into this a few times while developing: Sometimes date comes in as a string formatted like `2023-01-02`, and sometimes it is an actual date object. There probably is a better way to do this instead of trying to detect a string and parsing it each time? Or should I just make this logic a function?

client/charts/js/components/DataLoader.jsx
```diff
@@ -8,39 +8,87 @@ function Loader() {
...
 	useEffect(() => {
-		(async () => {
+		(() => {
 			setLoading(true)
-			const response = await fetch(dataUrl)
-			const content = await response.text()
-			const parsed = dataParser(content)
-			setData(parsed)
-			setLoading(false)
+
```
dkaoster, line 19: This chunk is the biggest change in terms of fetching logic, ended up allowing Dataloader to take arrays for `dataUrl`, `dataParser`, and `dataKey` so that it can fetch from multiple endpoints and cache responses as needed.

client/charts/js/components/DataLoader.jsx
```diff
@@ -8,39 +8,87 @@ function Loader() {
...
+				setData(data)
+				setLoading(false)
+			})
 		})()
-	}, [dataUrl, dataParser])
+	}, [dataUrl, dataParser, dataKey])
 
 	// If data hasn't loaded yet, return the loading component instead
-	if (loading) return loadingComponent
+	if (loading && loadingComponent) return loadingComponent
```
dkaoster, line 76: The `loadingComponent` pattern that we were doing here felt a little redundant and less than ideal, but instead of reworking the entire thing I simply made it so that it could be disabled with a falsy attribute. Open to ideas here if folks have thoughts.

incident/api/serializers.py
```diff
@@ -89,6 +94,8 @@ class CategorySerializer(serializers.Serializer):
     methodology = serializers.CharField()
     page_symbol = serializers.CharField()
     plural_name = serializers.CharField()
+    slug = serializers.CharField()
```
dkaoster, line 91: Added `slug` and `url`, the ticket has tasks to rename `page_symbol` but do we know what we want to rename it to?

Also there's a point about "Change symbol ID to a URL to the SVG instead" and while I can do that, I'm not clear on how that field can be used on the client-side. Looks like most of the svg icons are specified in CSS and generated with django? Am I missing something?

client/charts/js/components/HomepageSelection.jsx
chigby: Huh! I'm not seeing that locally (they're all strings for me) -- maybe something's different between our local databases?

chigby: Oh, actually, it seems like they're supposed to be objects. Also weird!

incident/api/tests/test_api.py
```diff
@@ -343,6 +343,8 @@ def test_result_attributes(self):
         self.assertEqual(data, {
             'id': self.category.pk,
             'title': self.category.title,
+            'slug': self.category.slug,
+            'url': "http://testserver" + self.category.url,
```
chigby: It's interesting that you had to include "testserver" here, which feels like an artifact of how Django's test client/server is configured. This caused me to look into a few things and it seems like what you are using to determine the full url, `build_absolute_uri`, is sidestepping Wagtail's Sites settings (which is, admittedly, a system I don't love). It might make more sense here for the category `url` field to be wired up similarly to how the incident serializer does it, as a `SerializedMethodField` hooked up to this method:

https://github.com/freedomofpress/pressfreedomtracker.us/blob/060333c917636a905bedb0633ceb9378f328ef96/incident/api/serializers.py#L170-L175

Which is to say, using Wagtail to come up with the URL. I think this is slightly "safer" because it respects wagtail's settings and is consistent with the rest of the API, even if in practice it shouldn't be different from the URL constructed from the incoming request's data. 

Let me know what you think about that, though. If that change is made, then we can update the test to look at `self.category.full_url` (though this is not IMO the main reason to make this change).

incident/api/serializers.py
chigby: @harrislapiroff  Do you have thoughts on `page_symbol` here? Did Accurat have a term more specific/meaningful they used when coming up with this concept?

chigby: It looks like the symbols are being used on the database page, i.e. http://localhost:8000/all-incidents/ -- in the category graph and controls for selecting one in the filter sidebar. The code for that is centered around this, it seems like:

https://github.com/freedomofpress/pressfreedomtracker.us/blob/060333c917636a905bedb0633ceb9378f328ef96/client/charts/js/components/CategoryFilter.jsx#L32-L46

https://github.com/freedomofpress/pressfreedomtracker.us/blob/060333c917636a905bedb0633ceb9378f328ef96/client/charts/js/components/CategoryFilter.jsx#L210

It's not totally clear from the issue to me, but maybe the idea is to replace these SVG definitions in JS with ones directly on the category object that are drawn from the API?

chigby (changes requested):
Added some comments. Let me know what you think. I'm not necessarily "Requesting Changes" but (1) making a potential suggestion about the `url` field, and (2) not sure exactly what to do with the SVG URL field, but it might turn into some more work if we get additional info

incident/api/serializers.py
harrislapiroff: I think if we can provide a URL to the SVG associated with the category, that would be ideal, but it may be pretty tricky given that SVGs seem to get bundled by webpack with hash urls, [e.g.](https://pressfreedomtracker.us/static/bundles/15618b161cd6278fab09.svg)

If that's not easily possible I think we can remove the field? It looks like the bit that @chigby cited all references to the SVGs are encoded in the javascript, not pulled from the API, so it's not clear to me that we're using the data from the API anywhere and it's definitely not a useful field for a third-party consumer as is.

client/charts/js/components/HomepageSelection.jsx
dkaoster: exactly, they're supposed to be objects but somehow are getting passed in as strings. Let me do some more digging to see if it makes more sense to do this parsing somewhere else.

incident/api/tests/test_api.py
dkaoster: That makes sense to me, let me give it a try

incident/api/serializers.py
harrislapiroff: One other possibility, since these SVGs are probably fairly small, would be to actually send the raw SVG code. That should be retrievable from the filesystem, I think.

incident/api/tests/test_api.py
dkaoster: looks like it worked, thanks!

incident/api/serializers.py
dkaoster: I thought of another way, which is simply to use CSS to add the symbols the same way that it is being done for all the other category symbols used around the app. That way, it doesn't need to be hardcoded and it also doesn't need to be sent from the API. Thoughts?

dkaoster:
@chigby looks like it's failing test coverage, I'm not sure how to test the `get_url` portion, any examples I can reference?

chigby:
@dkaoster I had thought that the other place we used this pattern would've been covered, but I guess that was before we instituted the coverage check on CI. So there's no place to "borrow" the corresponding test from. Sorry about that! 

Here's what I would do, though. The API tests folder should probably have a `test_serializers.py` or something similar, and here's a minimal test case I came up with that ought to cover the part of the new function that's being flagged:

```python
from django.test import TestCase
from wagtail.models import Site

from incident.api.serializers import CategorySerializer
from common.tests.factories import CategoryPageFactory


class CategorySerializerTest(TestCase):
    @classmethod
    def setUpTestData(cls):
        site = Site.objects.get(is_default_site=True)
        cls.category = CategoryPageFactory(parent=site.root_page)

    def test_successfully_serializes_category_page_urls(self):
        result = CategorySerializer(self.category)
        self.assertEqual(
            self.category.get_full_url(),
            result.data['url']
        )
```

Let me know if this works, or if, when you see this, you think of a different way of doing it.

incident/api/serializers.py
harrislapiroff: I think for the purposes of this PR let's leave it hard-coded and just remove the field from the API. We can revisit in the future, but I don't want to bikeshed too hard on this. (I'm concerned trying to add the symbols with CSS into an SVG will get dicey, so the right solution here doesn't seem immediately obvious to me.)

dkaoster: Sure, happy to remove it and leave it hard-coded. Although I'm curious what you mean by "get dicey" in this case? I've used this pattern before a couple times and although it looks odd at first, the only scenario that I've seen issues with this is when the svg is transferred into some environment that isn't the DOM, such as downloading the svg or trying to render it to a canvas. If those are things we plan to do then I agree we should revisit this in the future.

harrislapiroff: @dkaoster Okay, you've convinced me. Those definitely aren't things I expect us to be doing with this chart in particular. Go ahead and implement it the way you've proposed and then remove the SVG field from the API :)

client/charts/js/components/DataLoader.jsx
harrislapiroff: The reason we have the loading component here is to avoid a page reflow when a visualization loads by rendering a component that is the same visual size as the charts will be once the data loads. If you have a way to restructure that more elegantly, I'd be happy to see a PR for that!

harrislapiroff (approved):
This looks great to me!

chigby (approved):

PR #1621, opened 2023-04-14
Add snapshot frontend testing

### LINKED ISSUE #1548: Add frontend testing for react components
We already have some basic react components in the filter sidebar and charts. So I think we can start adding tests to catch some of the errors by JS unittesting. We might need to add further integration tests (or end-to-end tests). I think we can use jest and [react-testing-library](https://github.com/testing-library/react-testing-library) to write most of the unittest (it technically does integration tests for react components since it actually checks the dom generated, but not an E2E test with the backend). For an E2E testing further we can consider selenium or cypress maybe.

I think the deliverable for this issue can be to test all the react components used in the sidebar to prevent issues like https://github.com/freedomofpress/pressfreedomtracker.us/issues/1546

dkaoster:
connects #1548 

One easy, effective, and imperfect way to start doing some frontend testing is to add snapshot tests, which basically render the react component with mock data and save those to the repo, so that it notifies you whenever something changes.

Pros:
- Easy to create, and they provide a baseline test for unintentional regressions
- Maintain a static, rendered version of components in the repo for easy reference

Cons:
- Can generate noise in PRs
- Often imperfect because they aren't very good at testing logic or state, which still needs to be written separately


dkaoster:
@harrislapiroff @chigby @SaptakS I'm going to go ahead and say this is ready for review. There's a lot more test cases that I could write to get frontend testing more complete, but I'm feeling like this is a good breakpoint for the first iteration of snapshot testing. If y'all have a preference for getting even higher code coverage before merging I'm happy to keep chipping away at this too.

chigby:
When I run the tests locally, with 

```
docker compose exec node npm test
```

it does seem to make a change to one of the snapshot files (which I've truncated here, but it's more of the same):

```
diff --git a/client/charts/js/components/__tests__/__snapshots__/TimeMonthsFilter.test.js.snap b/client/charts/js/components/__tests__/__snapshots__/TimeMonthsFilter.test.js.snap
index 5a344f42..21967a89 100644
--- a/client/charts/js/components/__tests__/__snapshots__/TimeMonthsFilter.test.js.snap
+++ b/client/charts/js/components/__tests__/__snapshots__/TimeMonthsFilter.test.js.snap
@@ -87,99 +87,99 @@ exports[`renders TimeMonthsFilter with mocked data 1`] = `
         [
           {
             "count": 0,
-            "date": "2020-01-01T07:00:00.000Z",
+            "date": "2020-01-01T00:00:00.000Z",
           },
           {
             "count": 0,
-            "date": "2020-02-01T07:00:00.000Z",
+            "date": "2020-02-01T00:00:00.000Z",
           },
           {
             "count": 0,
-            "date": "2020-03-01T07:00:00.000Z",
+            "date": "2020-03-01T00:00:00.000Z",
           },
           {
             "count": 0,
-            "date": "2020-04-01T07:00:00.000Z",
+            "date": "2020-04-01T00:00:00.000Z",
           },
           {
             "count": 0,
-            "date": "2020-05-01T07:00:00.000Z",
+            "date": "2020-05-01T00:00:00.000Z",
           },
           {
             "count": 1,
-            "date": "2020-06-01T07:00:00.000Z",
+            "date": "2020-06-01T00:00:00.000Z",
           },
```

Is this expected, or maybe I'm not quite doing it right?

dkaoster:
@chigby ah yes that behavior is not quite right, let me fix this

chigby (approved):

Issue #1573, opened 2023-01-18
Full-text search is slow


harrislapiroff:
We observed a number of timeouts over this past weekend. Trawling through the logs, Cameron noticed around those timeouts some slow requests for pages that included a text search term. I am able to reproduce that performing a text search on the tracker website takes several seconds to respond: https://pressfreedomtracker.us/all-incidents/?search=antifa

Let's find ways to improve performance here. I'll set the bar for closing this ticket as consistently achieving <1s responses for filtering queries involving text searches on the database page.

Some ideas that came up in discussion:

* Can we rely on Wagtail's internal page search index?
* Can we get better performance by reordering how filters are applied (e.g., if we do text-search first can that search be run on an indexed dataset instead of dynamically generating the index)?
* Something something materialized view (I didn't fully understand this, feel free to add your actual idea @chigby)

chigby:
Saw some more worker timeouts today, and here are the surrounding log entries:

[Around 02:00](https://logs.freedom.press/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2023-03-30T05:57:00.000Z',to:'2023-03-30T06:03:00.000Z'))&_a=(columns:!(django.event,django.response.status_code,duration_ms,django.request.meta.QUERY_STRING),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'filebeat-*',key:django.request.path_info,negate:!t,params:(query:%2Fhealth%2Fok%2F),type:phrase),query:(match_phrase:(django.request.path_info:%2Fhealth%2Fok%2F))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'filebeat-*',key:resource.labels.cluster_name,negate:!f,params:(query:www-production),type:phrase),query:(match_phrase:(resource.labels.cluster_name:www-production))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'filebeat-*',key:resource.labels.namespace_name,negate:!f,params:(query:pressfreedomtracker),type:phrase),query:(match_phrase:(resource.labels.namespace_name:pressfreedomtracker)))),index:'filebeat-*',interval:auto,query:(language:kuery,query:''),sort:!(!('@timestamp',desc))))

[...]

PR #1626, opened 2023-04-19
Incident search filter improvements

chigby:
This pull request:

1. Updates the incident `SearchFilter` to employ the `IndexEntry` model's pre-computed textsearch vectors for querying against. This improves performance since we are not having to calculate these ourselves. I did observe faster load times locally, but staging is the place to test this. I will note that the performance gains were dependent on the search term used, with the greatest gains coming when it was a term that appeared in many incidents.
2. Updates the IncidentPage model to tell wagtail to index more fields into its overall searchable data. These fields are:
  - Body blocks
    - Rich text  
    - Image captions
    - Image attributions
    - Text contained embedded tweetes
    - Blockquotes
    - Blockquote sources
    - Pull quotes
    - Video captions
    - Video attributions 
  - City
  - State name
  - Introduction
  - Teaser
  - Primary/teaser image caption
  - Primary/teaser image attribution
  - Updates
    - Title
    - Body body blocks (same as above, where applicable)

I think this tends to make the search more useful.

Refs #1573

### DIFF
--- a/incident/models/incident_page.py
+++ b/incident/models/incident_page.py
@@ -31,7 +31,7 @@ from wagtail import blocks
 from wagtail.fields import StreamField, RichTextField
 from wagtail.models import Page, Orderable, PageManager, PageQuerySet
 from wagtail.images.blocks import ImageChooserBlock
-
+from wagtail.search import index
 from wagtailautocomplete.edit_handlers import AutocompletePanel
 from common.blocks import (
     RichTextBlockQuoteBlock,
@@ -769,6 +769,14 @@ class IncidentPage(MetadataPageMixin, Page):
     ]
 
     parent_page_types = ['incident.IncidentIndexPage']
+
+    search_fields = Page.search_fields + [
+        index.SearchField('body'),
+        index.SearchField('city'),
+        index.RelatedFields('state', [
+            index.SearchField('name'),
+        ])
+    ]
 
     def get_context(self, request, *args, **kwargs):
         context = super(IncidentPage, self).get_context(request, *args, **kwargs)
--- a/incident/models/incident_page.py
+++ b/incident/models/incident_page.py
@@ -775,7 +775,13 @@ class IncidentPage(MetadataPageMixin, Page):
         index.SearchField('city'),
         index.RelatedFields('state', [
             index.SearchField('name'),
-        ])
+        ]),
+        index.SearchField('introduction'),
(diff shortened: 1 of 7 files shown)


SaptakS:
@chigby I can review the code, but like you mentioned, it mostly needs to be tested at a staging environment, so should we wait for it to be merged till the experimental staging gets available again?

chigby:
@SaptakS That sounds reasonable to me

incident/models/incident_page.py
```diff
@@ -770,6 +770,24 @@ class IncidentPage(MetadataPageMixin, Page):
...
+        index.SearchField('introduction'),
+        index.SearchField('teaser'),
+        index.RelatedFields('teaser_image', [
+            index.SearchField('attribution'),
+        ]),
+        index.SearchField('image_caption'),
+        index.RelatedFields('updates', [
+            index.SearchField('title'),
+            index.SearchField('body'),
+        ]),
```
SaptakS, line 788: I see that we are using only the [title and body indexes](https://github.com/freedomofpress/pressfreedomtracker.us/pull/1626/files#diff-a3770fdca8d1cfb14cd8af7251251982012e063185ec18ea1b4573ba2985c9f7R524-R529) in the search. Adding these fields is to future proof?

SaptakS (approved):
Codewise looks good to me. We definitely need to test it in staging (probably experimental?) to get a better idea of whether it improves the speed. I am approving the PR but I think should be merged once further testing has been done. Though I don't think it necessarily breaks anything.

incident/models/incident_page.py
chigby: So how Wagtail's `IndexEntry` model works is not super well documented, but the title and body fields on the from your link are its built-in fields that contain searchable data (there is also a third one called `autocomplete` that allows for partial matches -- intended to be used for find-as-you-type varieties of searches). From how they are named, it might seem like there is a 1-to-1 correspondence between those fields and the fields of the model being indexed, but that's only true for `title`. The `body` field (on `IndexEntry`) includes all the search data that's not part of the title -- so all the fields that are listed in the `search_fields` attribute of `IncidentPage` will have their searchable text pulled in and indexed. You can actually test this out by performing a search based on, say, text in the `introduction` field and see that you will get results.

SaptakS:
We can test this further in staging. Merging.

PR #1629, opened 2023-04-20
Systematically convert Filter values to querystring values

### LINKED ISSUE #1627: Filtered download link does not properly handle date fields
### Steps to reproduce:

1. Visit the database page with a date filter, e.g. http://localhost:8000/all-incidents/?date_lower=2022-01-01

### Expected behavior:

The link at Download Dataset > Download your filtered dataset is: 

http://localhost:8000/all-incidents/export/?date_lower=2022-01-01

And this link downloads dates correctly filtered by that value.

### Actual behavior 
The link at Download Dataset > Download your filtered dataset is: 

http://localhost:8000/all-incidents/export/?date=%28datetime.date%282022%2C+1%2C+1%29%2C+None%29

which is not a valid format for the incident filter code.

chigby:
This change implements a filter-type-aware way of converting _valid_ filter values back into URL parameter values for further manipulation in the template (or elsewhere).

Fixes #1627

### DIFF
--- a/incident/models/incident_index_page.py
+++ b/incident/models/incident_index_page.py
@@ -2,7 +2,6 @@ import csv
 import json
 import copy
 from typing import TYPE_CHECKING
-from urllib import parse
 
 from django.db import models
 from django.http import StreamingHttpResponse, HttpResponse, JsonResponse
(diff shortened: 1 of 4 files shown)


SaptakS (approved):
Tested locally, works for me.

PR #1635, opened 2023-04-25
Updates postcss-loader and stylelint

SaptakS:
Check https://github.com/freedomofpress/securedrop.org/pull/1019#issue-1683075664

In this PR, I also had to update stylelint and related packages, since they all use cosmiconfig.


chigby (approved):
Looks great, and happy to not be required to order sass properties alphabetically :)

PR #1638, opened 2023-04-26
Fix Vertical Bar Chart

dkaoster:
Fixes tags being null


chigby (approved):
Works for me! Let's get it merged.

Issue #1636, opened 2023-04-25
Further Data Visualizations


harrislapiroff:
Now that we have a working [vertical bar chart](https://github.com/freedomofpress/pressfreedomtracker.us/pull/1614), let's implement some other visualization stream blocks. Here's what we should have along with some completion criteria for each. 

Editors being able to select which incidents to include in the dataset is completion criteria for each and should be accomplished in the same way as with the vertical bar chart. I imagine you'll want to create a reusable class for some of these fields.

Please complete each chart in a separate PR, queuing if needed. Please also include screenshots of the completed work in each PR description so review of the visuals is easy :)

- [ ] Treemap ([Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=2462-12651&t=YQgBwECdekhCewgM-4))

  - [ ] Editors should be able to select which incidents to include in the dataset

  - [ ] Design: Should rotate to vertical when viewport is narrow (responsive)

  - [ ] Design: Let's try to achieve an accessibility tree that looks something like this (if this is too complicated to achieve with react-animated-dataset, leave it incomplete for now and we'll discuss later):

        list "Categories"
           listitem "Arrests and Criminal Charges: 10 incidents"
           listitem "Subpoenas and Legal Orders: 7 incidents"
           listitem "Equipment Search and Seizures: 3 incidents"
           ...

      One useful thing here might be to recall that the categories API has a `plural_name` field for each category.

  - [ ] It might be nice to keep the code flexible enough that in the future we can give editors the option to choose what field to segment by (say, `arresting_authority` instead of `categories`) but don't spend too much work on this until we know it's a feature editors will use. I can imagine this being deceptively complex since different fields have different formats and tallying methods.

- [ ] Bubble Map ([Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=2803-18928&t=YQgBwECdekhCewgM-4))

  - [ ] Editors should be able to select which incidents to include in the dataset

  - [ ] Editors should be able to select whether to aggregate by `city, state` or just `state`

  - [ ] Design: Bubble tooltips as documented in the Figma

  - [ ] Design: Let's try to achieve an accessibility tree that looks something like this (if this is too complicated to achieve with react-animated-dataset, leave it incomplete for now and we'll discuss later):

        list "U.S. Map"
[...]

PR #1642, opened 2023-04-28
Treemap chart

dkaoster:
Implements TreeMap Chart in #1636

## Default State
Screenshot 2023-04-28 at 3 17 21 PM

## Hover State
Screenshot 2023-04-28 at 3 17 29 PM

## Multiple Selected State
Screenshot 2023-04-28 at 3 17 40 PM

@SaptakS Would be great to get some eyes on accessibility in this version of the chart.


chigby:
It's possible I don't quite understand exactly how a content editor is supposed to use or create charts of this type. Specifically, I think I'm having trouble with what the dataset options are and how that corresponds to the chart. Is it supposed to be the case that each "area" (not sure if this is the right word for the sections of the chart you can highlight) always corresponds to a category? Meaning, as an editor what you do is choose some categories and a date range to define a set of incidents, then this chart displays the total count of those incidents by category? This feels slightly limiting to me in terms of the spectrum of charts this feature can produce, but I also admit I'm not fully clued-in here. I just want to make sure I understand it before I go too deep into code review.

harrislapiroff:
@chigby I haven't looked at the PR, but I think it's possible you have understood the chart because it is a fairly limited form! It's roughly the equivalent of a category pie chart for a selection of incidents. I recommend glancing at the [newsletter figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?type=design&node-id=3698-62440&t=nPzREmEmD5mRJYxP-0) to see if it matches what's there.

I did mention in the Issue that we may want to someday expand its functionality to allow segmenting on something other than category, but I worried that would be too much to approach now without getting it as a specific request from editors.

chigby:
@harrislapiroff Gotcha. The newsletter figma does have a chart with "Camera Equipment" and "Other" but if that's out of the scope of this issue, that's fine.

harrislapiroff:
@chigby Oh, good catch. But, still, in the spirit of small changes lets queue that up next.

chigby:
Though I guess one thing I'm noticing is there's potentially a kind of mismatch between being able to select a set of incidents by categories, but those categories you select are not necessarily the only ones that will appear on the chart, because an incident can belong to more than one category.

harrislapiroff:
I think that's fine. I think it will be somewhat uncommon for editors to limit categories for this chart, but that is the behavior I would expect when doing so

common/templates/common/blocks/tree_map_chart.html
```diff
@@ -0,0 +1,26 @@
...
+	<div
+		class="chart-tree-map"
+		{% if value.incident_set.category %}
+			{% for category in value.incident_set.category %}
+				data-category-{{ forloop.counter }}="{{ category.title }}"
+			{% endfor %}
+		{% endif %}
+		{% if value.incident_set.tag %}
+			data-tag="{{ value.incident_set.tag }}"
+		{% endif %}
```
chigby, line 12: If we want to be able to have a tag value here, we ought to include it in the `PluralIncidentSet` block, above.

common/blocks.py
```diff
@@ -462,6 +462,25 @@ class SimpleIncidentSet(blocks.StructBlock):
...
+    lower_date = blocks.DateBlock(
+        label='Filter by Date, lower bound',
+        required=False,
+        help_text='If set, no incidents before this date will be included.',
+    )
+    upper_date = blocks.DateBlock(
+        label='Filter by Date, upper bound',
+        required=False,
+        help_text='If set, no incidents after this date will be included.',
+    )
```
chigby: Forgive the slightly meandering comment here. But I guess I want to communicate that I think we should be a little careful here with expanding the number of blocks that generally describe types of incident sets. I don't want us to end up in a confusing situation, either for developers or for end-users of this system. 

Because it seems like what we have here is a struct block that's _slightly_ different from `SimpleIncidentSet` except for (1) the ability to choose more than one category, and (2) lack of a tag chooser (though this may be an oversight, see comment below in `tree_chart_map.html`). And if we do feel that this type of chart needs/wants datasets containing a more complex category specification, I think it's worth a little time to ponder about if it's either possible to make this slight adjustment to our existing simple incident set block, or if we don't want that, if the capability for multi-category choices could be built as part of the basis for more complex dataset specifications in general (which I think we all want to have eventually).

My first inclination, honestly, is that I don't think we need this block at all. The issue (#1636) says choosing the dataset "should be accomplished in the same way as with the vertical bar chart," which I'm interpreting to mean we don't really need multiple categories here (though as I've said, my understanding of the requirements here is limited, so @harrislapiroff can correct me if I'm wrong). I really do think this is the simplest thing in terms of not having unnecessarily proliferating block types.

The second-simplest thing is probably to keep `PluralIncidentDataset` with some changes: 
  
1. update the help text on the category field to explain what choosing more than one category does, i.e. are incidents in _any_ selected category included or is it only ones in _all_ of them? 
2. Add the tag field from `SimpleIncidentSet` (if necessary) 

Overall, I guess my point is I'm afraid of us having a different type of data-set block for every chart type. If we can avoid that going forward, I'll be happy.

common/tests/utils.py
```diff
@@ -179,6 +179,23 @@ def generate_vertical_bar_chart():
...
+    }
+    return generate_field(
+        'tree_map_chart',
+        {
+            'title': make_words().capitalize(),
+            'incident_set': incident_set,
+            'description': fake.text(max_nb_chars=200),
+        }
+    )
+
```
chigby, line 197: Love this, this is great!

chigby (commented):
Leaving these thoughts here. The actual chart worked very well for me and I'm not sure if I have any commentary on that.

common/blocks.py
harrislapiroff: Hm, I'm not sure I see a reason for us to have two different ways of specifying an incident selection at all here. I'm guessing the multiple category select was added here because it felt wrong to only be able to select one category for a treemap where the purpose is to break down by category?

I do also agree we should allow tag selection here.

I think we should unify this behavior either by allowing multiple category selection on both chart blocks or switching to single category selection on both and I don't have a strong feeling which way we go. As mentioned previously, I don't expect the category selection to be used much with this block, but I'd like the way editors select groups of incidents to chart to be standardized except where absolutely necessary.

dkaoster: I think that's fair, I'm inclined to just allow multiple category selection on both chart blocks / any future visualization that we may want to do this for.

harrislapiroff: That sounds good to me!

common/blocks.py
```diff
@@ -438,12 +438,12 @@ class Meta:
 
 
 class SimpleIncidentSet(blocks.StructBlock):
-    category = blocks.PageChooserBlock(
+    category = blocks.ListBlock(blocks.PageChooserBlock(
```
chigby: It's looking to me like if we change this to a list block, pages that have a single value here are throwing an error in the wagtail admin (for me it's `TypeError: object of type 'NoneType' has no len()`). Though can you verify this? My local DB might be in a weird state. It's happening to me on pages with a vertical bar chart from before this branch was created.

If it is the case that it's causing an error, I could see two options: 1. we write a data migration to update the existing pages to have a list rather than a single value for the any blocks that use Simple Incident Set. I don't think it's a trivial problem but it probably can be done. Option 2: add a new field `categories` alongside this one that exposes the new behavior, migrate the old charts to the new format, then remove `category` once we're sure that's happened. This one has more steps but is probably safer.

dkaoster: Yeah I was seeing this issue too, and I had to do a complete delete / rebuild of the DB instance in docker fixed it. I'm guessing this means that we'd have to manually create a data migration to do this (option 1)?

I'm happy to do whichever method you think makes more sense for this ticket.

chigby: Yeah, we'd have to create a migration ourselves manually to, like, loop through all the blog pages and their blocks and if any of them has a simple incident set, update the category to be either `[]` or `[value]` if there's a value. Conceptually it's straightforward but, the last time I looked, anyway, the API for manipulating `StreamField` values, especially nested ones like we have here, is _not_ the most straightforward or user-friendly. I guess [here is the documentation](https://docs.wagtail.org/en/v4.1.5/topics/streamfield.html#modifying-streamfield-data). 

But the fact that we probably don't have that many pages using these charts yet combined with my general overall wariness about automated modifications to pages (which in this case would potentially affect every blog page) is making me lean towards the two-phase approach. Plus this gives us a chance to make the name plural, and also add some help text about what effect choosing multiple categories has (i.e. does it filter for _all_ choices or _any_ choice)). So maybe let's go with that one.

blog/migrations/0031_alter_blogpage_body.py
```diff
@@ -0,0 +1,35 @@
...
+
+def migrate_simple_incidents(apps, schema_editor):
+    blog_pages = apps.get_model('blog', 'BlogPage')
+    for blog_page in blog_pages.objects.all():
+        for block in blog_page.body:
+            if block.block_type == 'vertical_bar_chart':
+                incident_set = block.value.get('incident_set')
+                if incident_set.get('category') is not None:
+                    incident_set['categories'] = [incident_set.get('category')]
+                    blog_page.save()
```
dkaoster: @chigby took a quick first pass here, it seems like it works locally but some things feel a little strange to me.

1. Do I need to explicitly set `incident_set['categories']` to `[]` if the prior category is `None`?
2. is `blog_page.save()` the right call here?

client/charts/js/lib/utilities.js
```diff
@@ -128,6 +128,44 @@ export function filterDatasetByFiltersApplied(originalDataset, filtersApplied, c
...
+	return dataset
+		.filter(({ categories, tags, date }) => {
+			const incidentCategories = categories ? categories.split(',').map(d => d.trim()) : []
+			const incidentTags = tags ? tags.split(',').map(d => d.trim()) : []
+
+			const isExcludedCategory = filterCategories && !incidentCategories.find(c => filterCategoryMap[c])
+			const isExcludedTag = filterTags && !incidentTags.find(c => filterTagsMap[c])
+
+			const [startDate, endDate] = dateRange;
+			if (startDate) startDate.setHours(0)
```
dkaoster: TODO: this is going to need to change to use UTC dates too

blog/migrations/0031_alter_blogpage_body.py
chigby: 1. I'm realizing that I possibly misspoke (mis-typed?) when I said this earlier, as I think I was imagining the alternative solution of migrating the `category` field from a single value to a list in-place, rather that what we're actually doing here, which is having two fields side-by-side. So sorry about that, and no, I think in this case we don't need to set the categories to `[]` -- it already will be that (if I understand correctly).
2. Yes, `save()` is correct.

blog/migrations/0031_alter_blogpage_body.py
```diff
@@ -0,0 +1,36 @@
...
+        ),
+        migrations.RunPython(migrate_simple_incidents),
```
chigby: One thing I think would be nice here is to add a "reverse" function here that does not change the database, but having one allows the migration to be undone (meaning I can run `docker compose exec django python manage.py migrate blog 0030` to revert the DB state to before this migration -- it's definitely useful for testing the code, when you might have to run and re-run the migration a few times, but I think it's a good practice in general "just in case" we might want to do it in another environment).

[Here's some documentation](https://docs.djangoproject.com/en/3.2/ref/migration-operations/#runpython) but essentially what we want is something like:

```python
migrations.RunPython(migrate_simple_incidents, reverse_code=migrations.RunPython.noop),
```

common/blocks.py
```diff
@@ -444,6 +444,12 @@ class SimpleIncidentSet(blocks.StructBlock):
         page_type='common.CategoryPage',
         help_text='If selected, only incidents in the chosen category will be included.',
     )
+    categories = blocks.ListBlock(blocks.PageChooserBlock(
+        label='Filter by Category',
+        required=False,
+        page_type='common.CategoryPage',
+        help_text='If selected, only incidents in the chosen category will be included.',
```
chigby: Sorry to keep mentioning this, but I do think it would be helpful for users of this feature to explicitly spell out in the help text what happens if two or more categories are chosen: for an incident to be included in the chart does it have to be in _all_ the chosen categories, or is any single one of them sufficient?

blog/migrations/0031_alter_blogpage_body.py
chigby: To be clear, I think this migration works perfectly. It converted the charts I had locally so the correct category was in the new field, which is exactly what it's supposed to, from where I sit.

chigby (changes requested):

common/blocks.py
dkaoster: oh sorry this is on my for forgetting, thank you for the reminder!

chigby (approved):

PR #1646, opened 2023-05-08
Bubble Map

dkaoster:
Part 2 of #1636 

Implements Bubble Map


chigby:
Curious to know if you have a comment on the "accessibility tree" aspect of #1636 -- it's definitely not my area of expertise but using the Firefox Accessibility inspector on the chart does not seem to show a set of `listitem`. Maybe I'm looking in the wrong place? Also definitely possible that lack of control of these properties is a limitation of the `react-animated-dataset` package. But either way, let me know your thoughts on it.

dkaoster:
@chigby good catch, I must have missed this when putting this PR together. I've added it in now!

client/charts/js/components/IncidentsBubbleMap.jsx
```diff
@@ -0,0 +1,56 @@
...
+	countIncidentsOutsideUS
+} from '../lib/utilities'
+import { ParentSize } from '@visx/responsive'
+import ChartDownloader from './ChartDownloader'
+import USMap from './USMap'
+
+export default ({
+	dataset,
+	title,
+	description,
```
chigby, line 15: Speaking of accessibility: is this `description` field being used anywhere? Is this the kind of thing that we should be displaying as alt text or an ARIA attribute somewhere? Like I said, this is kind of out of my expertise.

chigby (commented):
I've looked over the code and I've tested out the various forms of this chart map locally and I think everything looks great. I put one question in code about the description field, and once that's answered/resolved (I might just not be looking in the right place for things) I think this should be good to go.

client/charts/js/components/IncidentsBubbleMap.jsx
dkaoster: also my bad, thanks for catching!

chigby (approved):
With the addition of the `description` into the chart, I think this is looking great.

PR #1649, opened 2023-05-16
Show Production URL when Downloading Charts

### LINKED ISSUE #1641: Hide URL in Chart When Downloading on Preview
Because we don't want downloaded charts to reference preview links, we should make sure that downloaded images only contain production urls.

dkaoster:
closes #1641


chigby (commented):
So I'm having a little trouble verifying that this is working, but it's possible I don't fully understand the scenario these changes are intended to affect. Here's what I'm doing, let me know if this is right:

1. Create a new blog page
2. Add the required title & date fields
3. Add a vertical bar chart to the body field with some arbitrary parameters
4. Preview the page, and open the preview in its own window
5. Download the image of the chart as a PNG

If I do that, the image does still seem to have the URL embedded in it. I think this may partially be related to how it seems like `page.full_url` is still populated with what the URL would be, even if it's an unpublished page being previewed. But also there might be something else going on?

dkaoster:
@chigby that's right, after thinking about this a little bit more, I figured that based on the problems we ran into last time, the issue wasn't that the URL was present in the graphic, rather it was that the URL was wrong because it linked to a preview URL. It feels like if we remove the URL entirely on preview environments, it may seem a little bit strange why sometimes the URL will show up and sometimes it won't. 

Maybe the crux of the issue is whether editors will make a blog page, use the vertical bar chart with only the intention of downloading the graphic without ever intending to publish it? If so, maybe not showing it when downloading from preview is better. Otherwise, I think we can just always have the graphic download with the correct URL (even if it is not published yet)

Thoughts @harrislapiroff?

harrislapiroff:
Let's get a little more information from @KAMcCudden about how of the team might make a page *just* to get a chart. If the answer is "never" or "infrequently" I think it's fine to display the "eventual" URL. If the answer is frequently, possibly we should just display the root URL during preview.

A couple other notes I'm thinking of:

- If this is something editors are doing often we should probably provide a dedicated interface for building and downloading charts.
- Maybe it would be nice for the download button to be a dropdown with options for either PNG or SVG (SVG would provide easier manual editing capabilities)
- In wagtail templates you can use `{% if request.is_preview %}` to accomplish different behaviors on preview vs publish—if that's helpful.

harrislapiroff:
Okay, chatting with Kirstin really simplified our requirements. Let's standardize on just showing "pressfreedomtracker.us" on all charts—no need to link back to the specific blog post.

harrislapiroff:
But also, for posterity, the team currently does not make charts independent of blog posts.

chigby (approved):
Looks great! Thanks for following this one though its various iterations.

PR #1650, opened 2023-05-16
CSV API Performance

chigby:
This pull request tries to improve the performance of the CSV-formatted API requests. It was inspired by the fact that we are using this API for our data visualizations, and these can be quite slow in production. For example, here's a page with one: https://pressfreedomtracker.us/blog/celebrating-5-years-since-launch-of-the-us-press-freedom-tracker/

It makes a request to the endpoint: https://pressfreedomtracker.us/api/edge/incidents/?format=csv&limit=99999&fields=date,title,url,categories

Of note:
* CSV formatted (I have learned the CSV format is _preferred_ over the JSON format for data visualizations, which was new info for me).
* `limit=99999` implies no pagination, though this is the default behavior for CSV requests
* Only 4 specific fields mentioned: date, title, url, categories
* No filtering

With this in mind, in this PR I have modified how our incident API view works so that (1) the existing functionality is preserved, but (2) in cases where it's possible, we give a faster response time. The insight I'm applying here is that it is less Django Rest Framework (DRF)'s serializer that's slowing us down (we are already using its most performant type of serializer), but Django's serializing of database rows into model instances. So we can improve performance by never doing that, using SQL expressions to generate the data we need, and then calling [`.values()` on our QuerySet](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#values) to avoid Django's models altogether. 

I've done this by creating a parallel set of code that takes effect for a user's requests to the incident API if it meets all of these requirements:
1. The user has asked for CSV format
2. The user has asked for a subset of fields I've enabled, which are mostly either ones that are "simple" values that don't require any type of processing, or ones I've written SQL expressions for that processing. 
3. The user has not requested any filters

These limitations are mostly there in the interest of expediency: I want to get this out there and I don't want to break anything. For each data field, I have to write some code deriving it as a database expression. Requirement 1 allows me to focus on one output format only. Requirement 2 limits how much code has to be written right now, especially when we have a lot of fields and I don't know which ones to prioritize (ideally the ones used in data visualizations across the site, or regularly requested by other API users). Requirement 3 is because I'm not exactly sure how the filters will interact with this method. I want to enable filters eventually, but I'm imagining it will initially have to be only a subset of ones supported (kind of like the home page CSV endpoint has support for date filters only).

All this is to say:

Under this PR, compared to develop, I'm seeing much faster times for requests like:

```
$ time curl -s "http://localhost:8000/api/edge/incidents/?format=csv&limit=9999&fields=title,date,url,categories"
```

(where I have ~1800 incidents in my local db).

And if there are any specific fields or filters we'd like to add before merging this PR, please let me know.

### DIFF
--- a/incident/api/views.py
+++ b/incident/api/views.py
@@ -178,6 +178,7 @@ class IncidentViewSet(viewsets.ReadOnlyModelViewSet):
             annotated_fields = {
                 'categories': 'category_summary',
                 'tags': 'tag_summary',
+                'links': 'link_summary',
                 'arresting_authority': 'arresting_authority_title',
                 'status_of_seized_equipment': 'status_of_seized_equipment_display',
                 'arrest_status': 'arrest_status_display',
--- a/incident/api/views.py
+++ b/incident/api/views.py
@@ -190,6 +190,7 @@ class IncidentViewSet(viewsets.ReadOnlyModelViewSet):
                 'third_party_business': 'third_party_business_display',
                 'status_of_prior_restraint': 'status_of_prior_restraint_display',
                 'url': 'url',
+                'state': 'state_abbreviation',
             }
             result_fields = []
             annotations = []
--- a/incident/api/views.py
+++ b/incident/api/views.py
@@ -134,8 +134,7 @@ class IncidentViewSet(viewsets.ReadOnlyModelViewSet):
     def requested_fields(self):
         if fields := self.request.GET.get('fields'):
             return set(map(str.strip, fields.split(',')))
-        else:
-            return set()
+        return set()
 
     def get_serializer_context(self):
         context = super().get_serializer_context()
@@ -146,8 +145,7 @@ class IncidentViewSet(viewsets.ReadOnlyModelViewSet):
     def can_apply_csv_serializer(self):
         if self.requested_fields:
             return self.requested_fields <= CSVIncidentSerializer().fields.keys()
-        else:
-            return False
+        return False
 
     def get_renderer_context(self):
         context = super().get_renderer_context()
@@ -166,8 +164,7 @@ class IncidentViewSet(viewsets.ReadOnlyModelViewSet):
             # that.
             if self.can_apply_csv_serializer:
                 return CSVIncidentSerializer
-            else:
-                return FlatIncidentSerializer
+            return FlatIncidentSerializer
         return super().get_serializer_class()
 
     def paginate_queryset(self, queryset):
--- a/incident/api/views.py
+++ b/incident/api/views.py
@@ -179,6 +179,8 @@ class IncidentViewSet(viewsets.ReadOnlyModelViewSet):
                 'categories': 'category_summary',
                 'tags': 'tag_summary',
                 'links': 'link_summary',
+                'equipment_broken': 'equipment_broken_summary',
+                'equipment_seized': 'equipment_seized_summary',
                 'arresting_authority': 'arresting_authority_title',
                 'status_of_seized_equipment': 'status_of_seized_equipment_display',
                 'arrest_status': 'arrest_status_display',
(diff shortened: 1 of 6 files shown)


harrislapiroff:
Nice insights! 

Not sure if this is the PR for it, but the mention of `limit=9999` makes me wonder if we should have an explicit flag for "all incidents" like `limit=0` or `limit=none`.

Looking forward to reviewing the performance gains here.

chigby:
Pushed a change to allow filters when using the "new" API logic. They didn't have much effect on performance.

harrislapiroff:
Hm, this is tough, because I could definitely imagine any number of fields being requested for data visualizations. For now let's just draw from the fields we use in existing visualizations. This means we should add:

* city
* state
* longitude
* latitude
* tags

One question: is this the sort of thing where we want to keep the list of supported fields pretty short for performance or is it the sort of thing where we can make it as long as we want it just means writing additional code for each field?

chigby:
To answer your question: I don't think having more _supported_ fields will affect performance much. I want to get as many supported as possible, but I wanted to start with the highest priority ones.

incident/api/views.py
```diff
@@ -123,6 +125,30 @@ def dispatch(self, *args, **kwargs) -> 'HttpResponse':
...
+    def csv_format_is_requested(self):
+        return getattr(self.request, 'accepted_renderer', None) and \
+            self.request.accepted_renderer.format == 'csv'
+
+    @cached_property
+    def requested_fields(self):
+        if fields := self.request.GET.get('fields'):
+            return set(map(str.strip, fields.split(',')))
+        else:
+            return set()
```
SaptakS: Small knitpick but I think the `else` is not needed since the `if` block has a return statement.
```suggestion
        if fields := self.request.GET.get('fields'):
            return set(map(str.strip, fields.split(',')))
        return set()
```

incident/api/views.py
```diff
@@ -135,8 +161,13 @@ def get_renderer_context(self):
...
     def get_serializer_class(self):
-        if getattr(self.request, 'accepted_renderer', None) and self.request.accepted_renderer.format == 'csv':
-            return FlatIncidentSerializer
+        if self.csv_format_is_requested:
+            # If it's possible to use the faster CSV serializer, use
+            # that.
+            if self.can_apply_csv_serializer:
+                return CSVIncidentSerializer
+            else:
+                return FlatIncidentSerializer
```
SaptakS:
```suggestion
            if self.can_apply_csv_serializer:
                return CSVIncidentSerializer
            return FlatIncidentSerializer
```

incident/api/views.py
```diff
@@ -123,6 +125,30 @@ def dispatch(self, *args, **kwargs) -> 'HttpResponse':
...
+        context = super().get_serializer_context()
+        context['requested_fields'] = self.requested_fields
+        return context
+
+    @cached_property
+    def can_apply_csv_serializer(self):
+        if self.requested_fields:
+            return self.requested_fields <= CSVIncidentSerializer().fields.keys()
+        else:
+            return False
```
SaptakS:
```suggestion
        if self.requested_fields:
            return self.requested_fields <= CSVIncidentSerializer().fields.keys()
        return False
```

SaptakS (changes requested):
Code looks mostly good to me. Left some comments. I also wonder if we can cache the annotations and enabled_filter_set_categories. I don't know how performance intensive operations they are though.

incident/api/views.py
chigby: Good point, I do like this more

SaptakS (approved):
I did a code review and looks good to me. I am not sure if there are more fields, etc. that needs to be added, so not merging it. Feel free to merge if you think it's complete.

PR #1652, opened 2023-05-18
Improve Vertical Bar Chart, Fix UTC Bug

### LINKED ISSUE #1640: Improve Vertical Bar Chart
After deploying #1614 there are some improvements that we would like to make:

- [ ] The x-axis can sometimes get too crowded if there are too many months (or entries) being drawn. (see below)
- [ ] We should add the ability to select whether to chart by month or by year in Wagtail.

image (2)

dkaoster:
closes #1640, also fixes a bug where the dates aren't parsing correctly due to not using UTC. (my bad)


SaptakS (approved):
Looks good to me.

PR #1661, opened 2023-05-25
Rework category chooser for incident sets as checklist

### LINKED ISSUE #1664: Improve category selector for incident datasets on blog page charts
Carving this off of #1661:

> I also think the treemap category chooser could be improved UX-wise. I know they're pages but I wonder if we could use a multi-autocomplete field or checkboxes for it instead of a page chooser? Feel free to split into a separate ticket if it's too hard to do both at once.

> If we can make checkboxes work, that would be great. I think the significant thing to an editor is the name of the categories, not where they’re found in the page tree, so I think that just adds extra steps. If it turns out to be complicated to implement, we can skip it though, once the bug is fixed. I do expect the most common behaviors will be to select one or zero categories, so this widget is still easy enough for those cases.

> Those checkboxes look much better to my eye. I think the help text could use a bit of work. I'd propose: "If selected, incidents belonging to any of the selected categories will be included," but feel free to modify

chigby:
Fixes #1664

This PR updates the `SimpleIncidentSet` block to store `categories` values as a list of strings whose values are the titles of category pages. These can be manipulated via a checkbox list.

It also includes a data migration to convert charts with the old style of categories to the new style. I have tested this, but please also check this using your local system. The migration in this PR should convert pages with something like this:

image

to something like this:

image


common/choices.py
```diff
@@ -1,3 +1,20 @@
+from django.db.models import TextChoices
+
+
+class IncidentCategoryChoices(TextChoices):
```
dkaoster: Is there any way to query these dynamically or do they have to be hardcoded?

chigby: Actually, now that you mention it, I think there is. Let me try to get that going.

chigby: In fact, we're already using that technique on the tags field! :headslap:

dkaoster (approved):
Works nicely for me!

SaptakS (approved):

PR #1662, opened 2023-05-25
Fix Treemap without Categories

### LINKED ISSUE #1656: Treemap won't render without a category
I can't seem to get a treemap to render if I don't choose some categories for it. We should make it render with a dataset including all categories if no categories are selected.

I also think the treemap category chooser could be improved UX-wise. I know they're pages but I wonder if we could use a multi-autocomplete field or checkboxes for it instead of a page chooser? Feel free to split into a separate ticket if it's too hard to do both at once.

image

dkaoster:
closes #1656 

Fixes the bug where having no categories causes the treemap to not show up. Generally updates the `filterCategories` prop to only be an array to prevent bugs where it could be either.


chigby (approved):
Fixes the problems I'm seeing locally, let's get it merged!

PR #1665, opened 2023-05-31
Update database page incident card design

### LINKED ISSUE #1659: Wrong incident preview design on database page
At some point the incident design on the database page got off sync from what it's supposed to look like on desktop, probably while modifying incidents for some other viewport or location on site?

Refer to [Figma](https://www.figma.com/file/7x2YldIvHjSbHZtpNX2lUq/PFT-%E2%80%93-Hand-off-Guide-(Copy)?type=design&node-id=4410-107575&t=k5nqAjLIA1iJYNLt-4) for designs and let me know if you have any questions.

## Desktop

Here's what it looks like now:
image

Here's what it should look like:
image

In particular:

- [x] Each field should be displayed compactly on the same line
- [x] No borders between fields
- [x] I'm also noticing tags should be gray not yellow. This one I think we never had, but let's update it to match the Figma. You can darken the text to pass WCAG color contrast standards (consult with @SaptakS for specifics)

## Mobile

Here's what it looks like now:
image

Here's what it should look like:

image

This is a lot closer, but let's:

- [x] make the spacing a little more compact
- [x] lighten up the lines
- [x] make the general fields single line
- [x] reduce horizontal margins around the category fields that aren't aligned

## General guidance

In general I've noticed we have trouble keeping these styles under control. There's some complicated components here that get reused in several places and we'll often update in one place not realizing it's also changing in another place. It might be worth applying some thought to how best share those styles while reducing the likelihood of error like that.

dkaoster:
closes #1659 

Updates the incidents card on the database page to match the [figma designs](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?type=design&node-id=3555-37026&t=VpJgTRKijO6Hca3j-0)

Desktop screenshot:
Screenshot 2023-05-30 at 8 37 03 PM

Mobile screenshot:
Screenshot 2023-05-30 at 8 37 33 PM

## To test
Open the incidents database on both desktop and mobile viewports, and ensure that it matches the mocks. Also double check other areas of the application to ensure there are no accidental regressions.


SaptakS:
@dkaoster the linting is still failing. I think let's update the stylelint config to use `color-hex-length: "long"` since I think we do use the long format everywhere. @harrislapiroff @chigby any objections to that?

harrislapiroff:
I’m on the fence. I think I prefer the short format, but not strongly, so I leave it to your and @dkaoster’s discretion

SaptakS:
I don't have any strong feelings about this either. 😅

dkaoster:
I don't have any strong feelings either, I'll just go ahead and use the short format for now!

client/common/sass/components/_buttons.sass
```diff
@@ -125,10 +125,13 @@
 		font-weight: 400
 		line-height: 1.23
 		padding: 0.25rem 0.5rem
-		background-color: $accent-background-tags
+		background-color: $tag-inactive-bg
+		text-transform: capitalize
+		color: $gray-text
```
SaptakS: This color doesn't pass the contrast ratio, I would atleast make it `#6b6b6b`. Maybe add a separate variable called gray-dark.

client/common/sass/components/_buttons.sass
```diff
@@ -125,10 +125,13 @@
 		font-weight: 400
 		line-height: 1.23
 		padding: 0.25rem 0.5rem
-		background-color: $accent-background-tags
+		background-color: $tag-inactive-bg
```
SaptakS: This changes the background color for tags in all the pages. Right now in the design, the database cards have grey and on hover becomes a bright yellow, while other pages have a lighter yellow color which becomes brighter on hover. I do like the idea of consistency, but @harrislapiroff do you want the grey background everywhere for tags, instead of yellow?

SaptakS: Honestly, the grey background with grey text kind of gives me a disabled vibe.

client/common/sass/components/_incident-database-card.sass
```diff
@@ -81,14 +77,28 @@
 
 		&--inline
 			display: block
-			padding-left: 19px
 			line-height: 0.5
 
+			+tablet-up
+				padding: 0.1rem 0
+				border: none
```
SaptakS: I think let's keep the same style for mobile and tablet given the closed state design is same and we don't have a specific design for tablets.

client/common/sass/components/_incident-database-card.sass
```diff
@@ -131,10 +141,11 @@
 .heading-table
 	margin: 0
 	border-bottom: 1px solid
-	border-color: $black
+	border-color: $gray-border
 	padding: 0.25rem 0 0
 	+tablet-down
-		padding: 0.5rem 0
+		padding: 0.3rem 0
+		margin-top: 1rem
```
SaptakS: This adds margin top for the category details heading also. When there are multiple categories, in the database page, there is no spacing between the category details tables.

This is the design:
image


This is how it looks after this PR:
image

SaptakS (dismissed):
I have left a few comments in the code. One other thing I noticed is that the legal order now shows weirdly in mobile. I think the "legal orders" key should be top aligned with the value.
image

client/common/sass/components/_buttons.sass
harrislapiroff: Hmmm, I see Accurat also was inconsistent about this, seemingly deliberately. Possibly the idea here was to make the database page less distracting? Or possibly just a quirk of the design.

image

I do like the idea of increasing the contrast which I think will help with the disabled vibe, so let's do that. But lets leave it inconsistent for now (yellow on the homepage, gray on the database) and see how it is to live with it.

SaptakS: I would probably just recommend using `$black` for text instead of finding a shade of gray. Even with an AA contrast level color (e.g. #6b6b6b), it looks pretty inactive to me.

SaptakS: And in database page, yellow is also used as the color for selected filters, so could be that was the reason for not using yellow background for tag?

dkaoster: Personally I am in favor of keeping it yellow the way it originally was for the whole site, because I do agree the gray makes it feel kind of like a disabled button. But curious to hear what you think @harrislapiroff

harrislapiroff (approved):
This looks great to me! @dkaoster I think you said you had changed the tags to yellow on the database page, but I'm still seeing them in gray (and yellow on hover) with darker gray text. Frankly, either is fine with me—I see good reasons either way. I think it looks great as is; I'll let you make the call, @dkaoster. Either merge as-is or change the tags to yellow and then merge. :)

dkaoster:
oh that's right, I did them and forgot exactly what I did to the colors! I think let's keep it this way for now to match the designs. We can revisit in the future if needed.

PR #1666, opened 2023-06-01
Add command to update blog page types by title

### LINKED ISSUE #1637: Devise a one-off command to convert newsletter posts to the newsletter template
In #1595 we added templates for newsletters and special sections. We have a lot of newsletter blog posts in the archive that are currently using the default template. We should update these to use the newsletter template.

I think the easiest approach here will be, in coordination with @maeve-fpf:

1. Devise a Python shell command that will identify blog posts that are newsletters (possibly by checking for the word "newsletter" in the blog post content?) and bulk update them to the newsletter template
2. Run this command on PFT staging first to check it
3. Run this command on PFT prod

chigby:
Fixes #1637

Adds a command `convert_blog_type` that takes two arguments: a regular expression and a blog page type. The command finds all blog pages whose titles match the regular expression and prints them, along with their current page type. If the command is run with `--commit`, then it updates all those pages to the given type. This, I hope, is explained with the `--help` flag. 

This command is a little more flexible than asked for by the issue, but I thought it might be helpful to have multiple options for converting pages, in case there's not a single, obvious way to select all blog pages (or if we need to bulk-convert some pages back).

### DIFF
--- a/blog/management/commands/convert_blog_type.py
+++ b/blog/management/commands/convert_blog_type.py
@@ -1,0 +1,50 @@
+from django.core.management.base import BaseCommand
+
+from blog.models import BlogPage
+
+
+class Command(BaseCommand):
+    help = 'Bulk update blog pages to use another template'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--commit',
+            action='store_true',
+            help='Commit changes to the database',
+        )
+
+        parser.add_argument(
+            'select',
+            metavar='REGEX',
+            help='Regular expression to select page titles to convert'
+        )
+
+        parser.add_argument(
+            'template_type',
+            choices=[choice[0] for choice in BlogPage.BLOG_TEMPLATE_CHOICES],
+            help='Template type to convert pages to',
+        )
+
+    def handle(self, *args, **options):
+        select_regex = options['select']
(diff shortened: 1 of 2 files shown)


harrislapiroff:
@chigby Just to clarify, part of the ticket is to figure out what the actual regex (or whatever approach) should be :)

chigby:
@harrislapiroff Looking at the production data, I think the command will work if run like this:

```
manage.py convert_blog_type "Newsletter$" newsletter
```

blog/management/commands/convert_blog_type.py
```diff
@@ -0,0 +1,50 @@
...
+
+        if selected_count < 1:
+            return
+        longest_title = max(len(page.title) for page in pages)
+        top_row = f'{"id":^4} | {"title":^{longest_title}} | current type'
+
+        self.stdout.write(top_row)
+        self.stdout.write('-' * len(top_row))
+        for i, page in enumerate(pages):
+            self.stdout.write(f'{page.pk:>4} | {page.title:<{longest_title}} | {page.blog_type}')
```
SaptakS, line 44: ❤️

SaptakS (approved):

PR #1669, opened 2023-06-02
Use chart or visualization as primary image

### LINKED ISSUE #1658: Allow charts to be the "primary image" for blog posts
(I've rethought this since I first drafted this, so apologies if you saw the substantially different previous version)

Accurat's [Figma files](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?type=design&node-id=3689-55821&t=PEyeZsQ2CW6xdy34-4) allow blog posts to have charts as their primary image:

Blog page:
image

Blog index page:
image

Here's how I'd like this to be implemented:

- [ ] Let's make the "lead image" field into a streamblock that can accept a single image block or any of our chart types
- [ ] Image or chart block gets rendered above the intro text on blog posts, as pictured in the mockup—full site width
- [ ] If the lead image is an image, we render thumbnails on the blog index page same as now
- [ ] If the lead image is a chart we render a simplified version of the chart, as pictured in the mockup above
   - [ ] **Stretch goal:** Simplified chart should be interactive and show a tooltip on hover, much like the full chart. We should probably include the chart title in the tooltip since the context of the page won't be around it
   - [ ] **Stretch goal:** We should render the chart server side to be the open-graph image (I'm guessing this won't actually be feasible without introducing a node rendering process into our server side setup, but adding this item in case I'm mistaken)
   - [ ] Vertical bar chart should be rendered vertically even though only horizontal bar chart thumbnails are shown in the mockup
   - [ ] There's no mockup for a simplified bubble map, so feel free to invent one, or render it as is

dkaoster:
closes #1658

Allows `IncidentsTimeBarChart`, `IncidentsTreeMap`, or `IncidentsBubbleMap` to be used as a primary image on blog posts, showing up at the top of the post. The chart in this position is static and cannot be interacted with.

## Screenshots of charts as primary image on blog posts

Screenshot 2023-06-13 at 6 43 12 PM
Screenshot 2023-06-13 at 6 43 45 PM
Screenshot 2023-06-06 at 3 59 46 PM
Screenshot 2023-06-13 at 6 42 46 PM

## Screenshot of charts as teaser image on blog posts listing pages

Screenshot 2023-06-14 at 5 38 57 PM

## To test and verify
- In wagtail, set the `lead_graphic` to a chart and it should show up in the primary image position of the blog post.
- In wagtail, set the `teaser_graphic` to a chart and it should show up as the teaser image position on the blog post listing pages.


blog/migrations/0035_alter_blogpage_lead_image.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('blog', '0034_update_and_migrate_incident_set_categories'),
+    ]
+
+    operations = [
+        migrations.RemoveField(model_name='blogpage', name='lead_image'),
```
dkaoster: TODO: I don't think this is quite right because it will delete the old image when instead we should be moving it into the new schema.

harrislapiroff: Feel free to ping @chigby about this today if you need any help with the necessary data migration.

blog/migrations/0035_auto_20230615_2139.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def update_lead_graphic(apps, schema_editor):
+    BlogPage = apps.get_model("blog", "BlogPage")
+
+    for page in BlogPage.objects.all():
+        stream_data = []
+
+        if page.lead_image:
+            stream_data.append(page.lead_image)
```
dkaoster: This doesn't quite work and I'm having trouble figuring it out. I need to convert `CustomImage` to something that works in the streamfield and I'm not sure exactly how to do this here. @chigby do you know how to do this easily?

blog/models.py
```diff
@@ -215,21 +215,19 @@ class BlogPage(MetadataPageMixin, MediaPageMixin, Page):
...
+    ], use_json_field=True, blank=True, default=[], max_num=1)
 
-    teaser_image = models.ForeignKey(
-        'common.CustomImage',
-        null=True,
-        blank=True,
-        on_delete=models.SET_NULL,
-        related_name='+',
-    )
+    teaser_graphic = StreamField([
```
dkaoster, line 225: TODO: double check all of the areas that use `teaser_image` to make sure they all get migrated over and don't break.

dkaoster:
Still need to fix up some tests that are failing, but this should be ready for a look!

harrislapiroff:
@dkaoster If you need any help on the tests go ahead and ping someone in #www-projects :)

dkaoster:
@chigby helped me get this figured out! 

Thank you Cameron for finding my typos 😬

common/templates/common/_video_or_image.html
```diff
@@ -4,8 +4,18 @@
...
+						<div class="primary-banner-media__inner">
+							  {% image image.0.value width-910 alt=image.attribution|default:image.0.value.title %}
+            </div>
+        {% elif image.0.block_type == "vertical_bar_chart" %}
+            {% include 'common/blocks/vertical_bar_chart.html' with value=image.0 static_graphic=true %}
+        {% elif image.0.block_type == "tree_map_chart" %}
+			      {% include 'common/blocks/tree_map_chart.html' with value=image.0 static_graphic=true %}
+        {% elif image.0.block_type == "bubble_map_chart" %}
+						<div class="primary-banner-media__inner">
+                {% include 'common/blocks/bubble_map_chart.html' with value=image.0 static_graphic=true %}
```
chigby: It's possible these lines need to be updated, but want to check first to make sure you're seeing the same tihng I am. 

When I create a graphic, the chart does render on the page, but it doesn't seem like any of the additional properties are being applied, such as the title or category/date/tag filters. Are you seeing this as well? If not, it might be something weird with the state of my system.

It's possible that these lines are responsible. My suspicion is that `image.0` is the struct block, while `image.0.value` is the `StructValue`, which is the object that has the properties we need to put on the chart element. So I think that replacing `image.0` with `image.0.value` might fix the problem. But I'm not 100% sure it's a universal problem.

common/templates/common/blocks/bubble_map_chart.html
```diff
@@ -1,5 +1,5 @@
 {% if value.title %}<h2>{{ value.title }}</h2>{% endif %}
-<div className="chartContainer">
+<div class="chartContainer">
```
chigby, line 2: Good catch there

home/templates/home/_recent_incidents_section.html
```diff
@@ -6,7 +6,7 @@ <h2 class="homepage-section-header">{{ label|default:"Recent Incidents" }}</h2>
 	<ul class="article-carousel__items">
 		{% for incident in incidents %}
 			<li class="article-carousel__item article-carousel__item--latest">
-				{% include "incident/_incident_card.html" with size="collapsible" details="small" incident=incident photo=incident.teaser_image only %}
+				{% include "incident/_incident_card.html" with size="collapsible" details="small" incident=incident photo=incident.teaser_graphic.0 only %}
```
chigby: Did we change _incident_ teaser images into graphics (with chart support) as well? I'm not seeing that update to `IncidentPage` elsewhere in this code.

blog/migrations/0035_auto_20230615_2139.py
```diff
@@ -0,0 +1,53 @@
...
+        ),
+        migrations.RunPython(update_lead_graphic, reverse_code=migrations.RunPython.noop, elidable=True),
+        migrations.RunPython(update_teaser_graphic, reverse_code=migrations.RunPython.noop, elidable=True),
```
chigby, line 52: These worked great for me :)

blog/models.py
```diff
@@ -215,6 +215,22 @@ class BlogPage(MetadataPageMixin, MediaPageMixin, Page):
...
+
+    teaser_graphic = StreamField([
+        ('image', ImageChooserBlock()),
+        ('vertical_bar_chart', VerticalBarChart()),
+        ('tree_map_chart', TreeMapChart()),
+        ('bubble_map_chart', BubbleMapChart()),
+    ], use_json_field=True, blank=True, default=[], max_num=1)
+
+    # Note: The following fields are deprecated in favor for lead_graphic and teaser_graphic,
+    # and will be removed in a future version
```
chigby, line 233: Thanks for adding this note!

chigby (changes requested):

common/templates/common/_video_or_image.html
dkaoster: oh yes very good catch!!

chigby (approved):

PR #1675, opened 2023-06-13
Blog index special post cards

### LINKED ISSUE #1617: Blog index page design for newsletter and special blog templates
We need to add a special teaser format for the special blog templates in the blog index page as shown in the design by accurat below.

image

Also, we need to add a tab UI to swap between the different kind of blog posts (articles, newsletter, special section)

image

[Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?node-id=3689-54926&t=q92LFBtIBwAa4rUz-4)

_Originally posted by @harrislapiroff in https://github.com/freedomofpress/pressfreedomtracker.us/issues/1595#issuecomment-1501879718_

Deliverables:
- [ ] Update index page layout to include the special blog teaser
- [ ] Add a tab to swap between different blog types

chigby:
Fixes #1617 

This pull request:

1. Adds type filter buttons to the top of the blog index page (when not also filtering by author or organization), for newsletters, articles, and special sections.
2. Creates a new template for showing full-width, background image blog cards. I've tried to make this match the design, but it's possible some of the width or height values will need to be adjusted to make potential images visually appealing across mobile/tablet/desktop widths. As part of testing, please experiment here, I'm sure there are things I didn't try.

### DIFF
--- a/blog/templates/blog/_blog_list.html
+++ b/blog/templates/blog/_blog_list.html
@@ -1,30 +1,32 @@
 {% load wagtailcore_tags common_tags %}
-
 <section class="blog-list {% if type %}blog-list--{{ type }}{% endif %}">
+	{% if not post_filters.author and not post_filters.organization %}
+		<div class="button-row">
+			<a class="btn btn-secondary button-row--item{% if not post_filters.blog_type %} button-row--active{% endif %}" href="?{% query_transform request type=None %}"><span>All posts</span></a>
+			{% for value, label in type_choices %}
+				<a class="btn btn-secondary button-row--item{% if value == post_filters.blog_type %} button-row--active{% endif %}" href="?{% query_transform request type=value %}"><span>{{ label }}</span></a>
+			{% endfor %}
+		</div>
+	{% endif %}
+
 	<h2 class="heading-regular">{{ heading | safe }}</h2>
 
 	<ol class="blog-list__posts {% if type %}blog-list__posts--{{ type }}{% endif %} {% if ajax_load and blogs.has_other_pages %}js-article-loading-parent{% endif %}">
 		{% for blog in blogs %}
 			<li
 				class="blog-list__item
-					{% if type %}blog-list__item--{{ type }}{% endif %}
+					{% if blog.blog_type == 'special' %}blog-list__item--special{% endif %}
 					{% if ajax_load %}js-article-loading-item{% endif %}"
 			>
-				{% include 'blog/_blog_card.html' with post=blog size=card_size|default:'small' photo=blog.teaser_graphic only %}
+				{% if blog.blog_type == 'special' %}
+					{% include 'blog/_special_blog_card.html' with post=blog only %}
+				{% else %}
+					{% include 'blog/_blog_card.html' with post=blog size=card_size|default:'small' photo=blog.teaser_graphic only %}
+				{% endif %}
+
 			</li>
 		{% endfor %}
 	</ol>
-
-	{% if type == 'featured' %}
-		<div class="blog-list__featured-scroller">
-			<button class="blog-list__featured-scroller--button blog-list__featured-scroller--button-prev">
-				{% include 'common/arrow-prev.svg' %}
-			</button>
-			<button class="blog-list__featured-scroller--button blog-list__featured-scroller--button-next">
-				{% include 'common/arrow-next.svg' %}
-			</button>
-		</div>
-	{% endif %}
 
 	{% if request and blogs.has_other_pages %}
 	<div class="blog-list__pagination">
--- a/client/common/sass/components/_blog-card-special.sass
+++ b/client/common/sass/components/_blog-card-special.sass
@@ -57,3 +57,13 @@
 
 		p
 			margin: 0
+
+	&::before
+		content: ' '
+		background: linear-gradient(0deg, #000 0%, rgba(0, 0, 0, 0.00) 100%)
+		z-index: 1
+		position: absolute
+		height: 600px
+		/* 1440px + 32px (margin left) + 32px (margin right) */
+		width: 1504px
+		max-width: 100%
--- a/client/common/sass/components/_blog-card-special.sass
+++ b/client/common/sass/components/_blog-card-special.sass
@@ -29,6 +29,7 @@
 		margin: 0
 		line-height: 1.2
 		font-size: $font-size-h2
+		font-weight: 400
 		color: $white
 		text-decoration: var(--title-text-decoration)
 		+desktop-up
--- a/client/common/sass/components/_blog-card-special.sass
+++ b/client/common/sass/components/_blog-card-special.sass
@@ -6,6 +6,8 @@
 	border-color: $black
 	height: 600px
 	cursor: pointer
+	margin-left: -32px
+	margin-right: -32px
 
 	--scale: 1
 	--title-text-decoration: none
--- a/client/common/sass/components/_blog-card-special.sass
+++ b/client/common/sass/components/_blog-card-special.sass
@@ -56,6 +56,9 @@
 		p
 			margin: 0
 
+		.text-link--card
+			text-decoration: none
+
 	&::before
 		content: ' '
 		background: linear-gradient(0deg, #000 0%, rgba(0, 0, 0, 0.00) 100%)
--- a/client/common/sass/components/_blog-card-special.sass
+++ b/client/common/sass/components/_blog-card-special.sass
@@ -1,0 +1,57 @@
+.special-blog-card
+	display: grid
+	grid-template: "container"
+	overflow: hidden
+	border-top: 1px solid
+	border-color: $black
+	height: 600px
+	cursor: pointer
+
+	--scale: 1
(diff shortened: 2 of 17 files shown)


blog/templates/blog/_blog_list.html
```diff
@@ -1,31 +1,33 @@
...
+	{% if not post_filters.author and not post_filters.organization %}
+		<div class="button-row">
+			<a class="btn btn-secondary button-row--item{% if not post_filters.blog_type %} button-row--active{% endif %}" href="?{% query_transform request type=None %}"><span>All posts</span></a>
+			{% for value, label in type_choices %}
+				<a class="btn btn-secondary button-row--item{% if value == post_filters.blog_type %} button-row--active{% endif %}" href="?{% query_transform request type=value %}"><span>{{ label }}</span></a>
+			{% endfor %}
+		</div>
+	{% endif %}
+
+	<h2 class="heading-regular">{{ heading | safe }}?</h2>
```
SaptakS: I think this question mark is left here by mistak?

client/common/sass/components/_button-row.sass
```diff
@@ -0,0 +1,28 @@
...
+
+	&--active
+		color: $white
+		background-color: $black
+
+	&--item
+		list-style-type: none
+		display: inline-block
+		text-indent: 0
+		padding: 1rem
```
SaptakS: The padding according to the figma should be `0.5rem 0.75rem`

client/common/sass/components/_button-row.sass
```diff
@@ -0,0 +1,28 @@
...
+	padding: 0
+
+	&--active
+		color: $white
+		background-color: $black
+
+	&--item
+		list-style-type: none
+		display: inline-block
+		text-indent: 0
```
SaptakS: What was the need for `text-indent` here? I am a bit confused.

client/common/sass/components/_blog-card-special.sass
```diff
@@ -0,0 +1,57 @@
+.special-blog-card
+	display: grid
+	grid-template: "container"
+	overflow: hidden
+	border-top: 1px solid
+	border-color: $black
+	height: 600px
+	cursor: pointer
```
SaptakS, line 6: I think it needs a `margin-left: -32px; margin-right: -32px` to cancel out the padding added by the `constrain-to-desktop` mixin so that in laptop screen width it can take the full width.

client/common/sass/components/_blog-card-special.sass
```diff
@@ -0,0 +1,57 @@
...
+	> *
+		grid-area: container
+
+	img
+		width: 100%
+		height: 100%
+		object-fit: cover
+		place-self: center
+		transition: transform 0.15s ease-in-out
+		transform: scale(var(--scale))
```
SaptakS, line 26: I think we should add a gradient filter on the image. Adding the following ::before to `.special-blog-card` should do the trick I think 
```css
	&::before
		content: ' '
		background: linear-gradient(0deg, #000 0%, rgba(0, 0, 0, 0.00) 100%)
		z-index: 1
		position: absolute
		height: 600px
		
		 /* 1440px + 32px (margin left) + 32px (margin right) */
		width: 1504px 
                max-width: 100%
```

Basically to darken the background image a bit so that the text contrast is better? It will still need to be reviewed by editor when chosing the image, but I feel maybe adding the gradient will help. It's also there in the figma designs.

client/common/sass/components/_blog-card-special.sass
```diff
@@ -0,0 +1,57 @@
+.special-blog-card
+	display: grid
+	grid-template: "container"
+	overflow: hidden
+	border-top: 1px solid
+	border-color: $black
```
SaptakS: I actually don't see a `border-top` in the figma design.

client/common/sass/components/_blog-card-special.sass
```diff
@@ -0,0 +1,57 @@
...
+		place-self: center
+		transition: transform 0.15s ease-in-out
+		transform: scale(var(--scale))
+
+	&__title
+		margin: 0
+		line-height: 1.2
+		font-size: $font-size-h2
+		color: $white
+		text-decoration: var(--title-text-decoration)
```
SaptakS, line 34: Since you are doing a `text-decoration` here, I think the text-decoration on hover of `text-link--card` needs to be removed for the special blog cards. If you hover particularly on the link text, you can see a black underline otherwise.

PS: super happy to see variable usage. 🥳

SaptakS (changes requested):
I think it's okay to use `grid-auto-flow: dense` as discussed for one of the solutions. But one of the things I notice (don't know how big of a deal it is) is that in the figma design the reason it looks very nice is because of the 8 normal card and then 1 special card taking full width and then 8 normal card again. The uniformity makes it look nice.

image

But in our dev environment for example, there are 12 normal cards, followed by 2 special cards, and then 2 normal cards. I am a little torn about this and probably @harrislapiroff can weigh in more:
- The current approach with `dense` is nice because it maintains chronology (mostly) but the tab ordering is different than the ordering of the HTML elements and also the layout (according to me) doesn't look great without the uniformity
- The other solution that I had proposed before that would maintain the tab order and also the design uniformity is if we create a list of blog post such that it will always be 8 normal blog and then the latest special blog, and then 8 normal blog and then the second latest special blog. The issue with this is the chronology won't be maintained, and if a special blog is really old, it will still appear as the 9th element in the listing maybe. But th UI will be what we want. So basically this approach is like treating 2 separate chronology for the normal blogs and special blogs and then merged in UI. Does that make sense?

I have also requested some other changes in the PR irrespective of the above.

harrislapiroff:
Some comments without testing the code myself:

I think I'm fine with the `dense` approach used here. I don't think we need to divide blog posts up into groups of eight—four is fine—at least for the moment.

As long as the tab ordering is still reasonable, I don't feel super strongly about it strictly matching the layout. In fact it might be preferable if the tab ordering was strict chronological order. Is it? Or is it something else?

client/common/sass/components/_blog-card-special.sass
harrislapiroff: Oh, great! Glad you caught this.

client/common/sass/components/_blog-card-special.sass
```diff
@@ -0,0 +1,57 @@
...
+		+desktop-up
+			font-size: $font-size-h3-desktop
```
harrislapiroff, line 36: It looks to me in the Figma like the heading is larger than this and not bold weight

SaptakS:
> In fact it might be preferable if the tab ordering was strict chronological order. Is it? Or is it something else?

It is.

blog/templates/blog/_blog_list.html
chigby: Yes, it definitely was :sweat_smile:

client/common/sass/components/_button-row.sass
chigby: I'm actually not sure about this! Maybe I was originally using a different element that needed this? Either way, I can remove.

client/common/sass/components/_blog-card-special.sass
chigby: That's a good point. Sometimes it's hard for me to tell by looking at the Figma what's supposed to be full-width or not.

client/common/sass/components/_blog-card-special.sass
chigby: Thanks for coming up with the CSS for this! Again, it's not super obvious from the Figma that this is there (or maybe it can be seen by drilling down into the Figma element information? Not sure), as visually I wasn't sure if they just carefully chose an image that was darker on the bottom or if a gradient like this was applied.

client/common/sass/components/_blog-card-special.sass
chigby: Nice job seeing this!

client/common/sass/components/_blog-card-special.sass
chigby: I see the lack of bold, now, but when I look at the font size in Figma (by selecting the text element in the design) it says 60px (see far right column):

image

Which, it seems like this `$font-size-h3-desktop` gives us?

https://github.com/freedomofpress/pressfreedomtracker.us/blob/d16414ffff77eec57d74cca5b9e9f928bad62195/client/common/sass/_typography.sass#L20

That was my reasoning, anyway. Happy to change it to something else but I'm not sure what.

harrislapiroff: probably just a perceptual artifact of the screenshot size then—go ahead and keep it

client/common/sass/components/_blog-card-special.sass
SaptakS: Yes. Usually, you need to select the elements in the Figma design and they have the information about the gradients and other stuff. I do think that editors will need to still chose the image carefully.

SaptakS (approved):

PR #1676, opened 2023-06-15
Replace Docker Compose v1 with v2 on CircleCI

chigby:
I've replaced `docker-compose` with `docker compose` everywhere I could find in the CircleCI config and the Makefile (which is used by the CircleCI config). This is the command (I think?) we are using locally for development, so it should match CI. What this does, basically, is start using docker compose version 2:

https://www.docker.com/blog/new-docker-compose-v2-and-v1-deprecation/

I've also updated what version of Ubuntu VMs CircleCI is using, so that we have access to that version of docker/docker compose. 

Also made a minor fix to the `compile-pip-dependencies` command to also update the ci requirements.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,8 @@
 ---
+refs:
+  setup_remote_docker: &setup_remote_docker
+    setup_remote_docker:
+      version: 20.10.18
 version: 2.1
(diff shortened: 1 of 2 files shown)


harrislapiroff (approved):
lgtm

PR #1681, opened 2023-06-23
Fix graphics without links

dkaoster:
Fixes the bug we are currently seeing with the bubble map not rendering.

Screenshot 2023-06-23 at 10 55 59 AM

Essentially, the map was expecting the dots to be able to link out to another page, but on blog posts we don't link out to the database page. Fixing this uncovered the fact that the `<a>` will still get rendered when we don't have a url to link to, which will render fine but is bad for accessibility. To fix this, I've made these changes that check whether we have a url to link to, and only renders the wrapping `<a>` tag if we do.

### DIFF
--- a/client/charts/js/components/BarChart.jsx
+++ b/client/charts/js/components/BarChart.jsx
@@ -49,7 +49,7 @@ export default function BarChart({
 	id = '',
 	numberOfTicks = 4,
 	description,
-	searchPageURL = () => {},
+	searchPageURL,
 	// function prop received from ChartDownloader that binds the svg element to allow
 	// it to be downloaded
 	setSvgEl = () => {},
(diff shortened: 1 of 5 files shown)


harrislapiroff (dismissed):
Thanks for the quick fix! This mostly looks good, but a couple things:

- [x] Looks like you'll need to update snapshot tests and make the linter happy
- [x] Since we're OK to wait till Monday, I wonder if we could DRY this up somehow. Two ideas come to mind:
  1. Make a quick component for each internal component inside the link or
  2. Dynamically swap the outer component between link `<a>` or fragment `<>` depending on whether the URL exists. (I'm not 100% sure this is possible, but it might be by importing the `a` and `Fragment` components directly and then doing like a `const wrapper = url ? a : Fragment`?)

chigby (approved):
I think the DynamicWrapper here is a pretty clever solution to this problem and gets around the mentioned DRY-ness problems nicely.

PR #1683, opened 2023-06-27
Add capability to segment treemaps by other fields

### LINKED ISSUE #1657: Provide options for fields to segment treemap by
Currently treemaps are always segmented by category:

image

Let's provide a few additional options. We don't have a super good user story right now for which fields editors will find most useful, but I think the best way to prompt some feedback will be to offer a few and see what fields editors request for us to add.

I suspect this issue will have some hidden complexity (hence adding fields one by one instead a generic process) as tallying different fields will require different approaches. So please be verbose here and in slack when you run into conceptual issues or want to propose an architecture that you're unsure of.

Let's start with allowing these fields:

* Categories
* Status of Charges (will probably want to count "by charge" rather than "by incident" for this one)
* Assailant (this is a simple single choice field so can count "by incident")

chigby:
Fixes #1657 

This PR adds the ability for treemap charts to be segmented by "Assailant" and "Status of Charts" (in addition to "Categories" which was previously the only way to configure this type of chart). I'm calling these segments "branches" in the code (see, it's like a tree with branches, right?) The door is definitely open to add more branch kinds later, this is essentially a proof of concept PR.

I got into some more detail about the changes in the commit messages, so please read those. The short version is there's now a field on the tree map block that lets you choose what you want to group by, and that selection passes the names of the branches and where they're found in the dataset to the JS component, which basically works exactly as it did before.

I had to add "status of charges" to the API as a field you could query for. It was _filterable_ before, but you couldn't ask for a comma-separated list of all the different charge statuses that applied to an incident.

I also updated the tree map component to use the "normal" API to get its data, as opposed to the "homepage CSV" action of the API. With [recent improvements](https://github.com/freedomofpress/pressfreedomtracker.us/pull/1650), the performance should be on par (but let me know if you see otherwise).

### Some screenshots

image

image

Note: the colors are the same for all charts.


dkaoster:
One thing I'm noticing while testing is that if there are no items in the "category", the button still shows up in the legend but doesn't actually have any values in the treemap. See screenshot:

Screenshot 2023-06-28 at 10 34 40 PM

Seems like the easiest way to solve this is to simply filter out categories that have numberOfIncidents = 0.

chigby:
Oh, interesting! I guess my personal test data had all the items for all the groups. Let me see if I can address it like you say.

chigby:
@dkaoster Here's my attempt at fixing this: https://github.com/freedomofpress/pressfreedomtracker.us/pull/1683/commits/bce2d5d82119f8e558ccfae0116cdef8b5f295e7 -- what do you think?

dkaoster:
Looks like that worked! Let me know if you need any help resolving conflicts now that my PR was merged.

chigby:
Merge conflicts should be resolved (it wasn't too bad!)

dkaoster (approved):
From my somewhat limited understanding of everything it seems to look and work well!

harrislapiroff (approved):
Approving on the basis of Daniel's review since he doesn't have codeowner privileges, whoops.

PR #1684, opened 2023-06-29
Fix searchPageURL references

dkaoster:
Fixes a `searchPageURL` reference that I missed in a previous round of work. This fixes the bar chart link in the category pages.


harrislapiroff (approved):
Tested this locally and it worked well!

PR #1687, opened 2023-06-30
Fix charts not showing up on home page, and hide chart titles

### LINKED ISSUE #1686: Blog post teaser charts not displaying on homepage for featured blog posts
When featured on the homepage, blog posts with a teaser graphic chart set don't get the chart rendered:

Image

I suspect we need to add the rendering javascript to the homepage. It should render the same as it does on a blog index page:

Image

Though in both places I think we should also not render the chart title (it makes the graphic too tall and is unnecessary with the article title there).

dkaoster:
closes #1686 1686

Fixes teaser image on posts not showing up in the home page, and fixes titles showing up on these post lists.

### DIFF
--- a/client/charts/js/bubble-map-chart.js
+++ b/client/charts/js/bubble-map-chart.js
@@ -46,3 +46,17 @@ function engageCharts() {
 }
 
 engageCharts();
+
+// Add listener for query change to rerun
+let previousUrl = '';
+
+const observer = new MutationObserver(function(mutations) {
+	if (window.location.href !== previousUrl) {
+		previousUrl = window.location.href;
+		engageCharts();
+	}
+});
+const config = {subtree: true, childList: true};
+
+// start listening to changes
+observer.observe(document, config);
--- a/common/models/pages.py
+++ b/common/models/pages.py
@@ -163,6 +163,14 @@ class MediaPageMixin:
         except Exception:
             pass
 
+        # If the page is a home page, loop through all the panels and posts to see if
+        # there is a chart as the primary image
+        try:
+            for blog_post in self.featured_blog_posts.select_related('page'):
+                block_cls_names = block_cls_names + get_page_blocks(blog_post.page)
+        except Exception:
+            pass
+
         for block_cls_name in block_cls_names:
             block_cls = import_string(block_cls_name)
             if hasattr(block_cls, 'Media') and hasattr(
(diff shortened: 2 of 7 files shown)


client/charts/js/bubble-map-chart.js
```diff
@@ -46,3 +46,17 @@ function engageCharts() {
 }
 
 engageCharts();
+
```
dkaoster, line 49: Realized that I forgot this in the previous PR, adding this in here

common/models/pages.py
```diff
@@ -163,6 +163,14 @@ def streamfield_media(self, media_type):
         except Exception:
             pass
 
+        # If the page is a home page, loop through all the panels and posts to see if
```
dkaoster, line 166: This makes sure the js is included on the home page.

common/templates/common/blocks/bubble_map_chart.html
```diff
@@ -1,4 +1,4 @@
-{% if value.title %}<h2>{{ value.title }}</h2>{% endif %}
+{% if static_graphic is None and value.title %}<h2>{{ value.title }}</h2>{% endif %}
```
dkaoster, line 1: This hides the title if the graphic is a "static_graphic" when used in blog listing page or as primary image on the top of a page.

SaptakS:
When I tried to use bar chart as teaser image, it shows up like this 
image

Is this how it is supposed to show? I feel like there should be black borders somewhere. Though maybe this is a separate issue requiring a separate PR.

SaptakS:
One other thing I noticed is that graphic takes some time to load for obvious reason. Can we have a placeholder till the vizualization is loaded to avoid any jump in the layout?

harrislapiroff:
Looks like in [the figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?type=design&node-id=3689-55821&mode=design&t=jM0YYpGBK8qOE8U8-4) they add a subtle darker background. I don't know how big of a difference that makes, but we can open a separate ticket for this and for the loading time, but I'd like to get this fix merged if it's otherwise good. Did it look good to you otherwise @SaptakS?

image

harrislapiroff (approved):
Saptak confirmed in slack that this PR looks good otherwise, so I'm going to merge and open new tickets for the other issues.

PR #1688, opened 2023-07-03
Updates django to 3.2.20

SaptakS:
You can read about the vuln here https://www.djangoproject.com/weblog/2023/jul/03/security-releases/

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -196,9 +196,9 @@ decorator==5.1.1 \
     # via
     #   ipdb
     #   ipython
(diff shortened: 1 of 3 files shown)


harrislapiroff (approved):
LGTM

harrislapiroff:
Safety is newly failing. This clearly isn’t a regression and I’m already a skeptic of ReDOS vulns (which is what it's reporting) and its not proposing any reasonable fixes, so I’m gonna override merge and we can deal with it later.

PR #1690, opened 2023-07-05
Adds back teaser images for incident pages

SaptakS:
As reported by Kirstin in slack, the incident pages were not showing the teaser images. This is because while adding of charts as primary image, the logic for image in blog was changed, and that code is shared with incidents template as well.

I have put the code back for adding image in incident pages to that template. Also, ensured that both visualization and normal images work in blog pages.

### DIFF
--- a/common/templates/common/_video_or_image.html
+++ b/common/templates/common/_video_or_image.html
@@ -1,24 +1,26 @@
 {% load wagtailcore_tags wagtailembeds_tags wagtailimages_tags %}
 
 {% if video or image %}
-    <figure class="primary-banner-media">
-        {% if video %}
-            {% embed video %}
-        {% elif image.0.block_type == "image" %}
-						<div class="primary-banner-media__inner">
-							  {% image image.0.value width-910 alt=image.attribution|default:image.0.value.title %}
-            </div>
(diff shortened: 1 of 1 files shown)


common/templates/common/_video_or_image.html
```diff
@@ -5,17 +5,19 @@
...
         {% elif image.0.block_type == "tree_map_chart" %}
 			      {% include 'common/blocks/tree_map_chart.html' with value=image.0.value static_graphic=true %}
         {% elif image.0.block_type == "bubble_map_chart" %}
-						<div class="primary-banner-media__inner">
-                {% include 'common/blocks/bubble_map_chart.html' with value=image.0.value static_graphic=true %}
-						</div>
+			<div class="primary-banner-media__inner">
+				{% include 'common/blocks/bubble_map_chart.html' with value=image.0.value static_graphic=true %}
+			</div>
+		{% else %}
```
dkaoster: Doesn't this need to be `{% elif image %}`? Otherwise what happens if you don't include an image at all?

dkaoster (approved):

harrislapiroff:
Going to force merge on the basis of @dkaoster's approval. Safety failure is unfixable, possible in error?

PR #1693, opened 2023-07-07
Service for Pregenerating Charts and Maps

### LINKED ISSUE #1668: Prerender and pregenerate charts for "primary image"
Related to #1658, we'd like to explore methods for rendering charts to html and static images.

## The Problem(s)
The reason for doing this is two-fold. As can be seen in #1658, we would it to be possible to set the chart as the primary image of a blog post. While there is currently code to generate the interactive version of these charts, they are all written in client-side React, which means that they only get loaded after the page loads. (and they don't load at all for folks with javascript disabled)

image

The second problem lies in the "meta-image", or the image that gets shared to social media sites and search engines as the promotional image for a blog post. If we were to use this chart as a primary image, the chart should also be the promotional image for the blog post, as the mocks show below.

image

Unfortunately, this would require for the image to be hosted on a URL somewhere for the image to be properly linked to.

## The Proposal(s)

The proposal is to spin up a lambda service running node.js that can generate both the SVG chart to put in the large primary image slot on blog pages, as well as render the SVG chart as a png / jpg. Both of these could then be stored on an S3 bucket and then linked to directly in the html. Then, whenever a new blog post is published in wagtail, wagtail could simply spin up this lambda service to generate these assets and store them in S3. This way, the service only runs when it is needed, and isn't in the line of requests blocking a user's page load of a blog post.

dkaoster:
Fixes #1668

This pull request adds a service called `node-chart-pregenerator` to our docker compose setup.  This serivce exposes an HTTP API (documented further below) and returns SVG or PNG images of the Javascript charts that appear on the site.  To interact with this serivce, this pull request also adds capabilities to the Django app to request and save these images.  Finally, blog pages have been updated so that if the `teaser_graphic` field is configured as a chart, then the image generated by the service will appear in the social media metadata for that page (on the page itself and in the RSS feed).

## Chart Pregeneration Service

This work allows for charts and maps to be generated on the fly and included as images on social media platform open graph images.

To run, check out this branch, and simply run the app as normal:

```
docker compose up
```

Then, make sure that the main application is running, and you can see the charts and maps as follows:

| Chart or Map     | URL                                                     |
| ----------------- | --------------------------------------  |
| Bar Chart SVG | http://localhost:3000/bar-chart.svg |
| Bar Chart PNG | http://localhost:3000/bar-chart.png |
| Treemap Chart SVG | http://localhost:3000/treemap-chart.svg |
| Treemap Chart PNG | http://localhost:3000/treemap-chart.png |
| Bubble Map SVG | http://localhost:3000/bubble-map.svg |
| Bubble Map PNG | http://localhost:3000/bubble-map.png |

Further query parameter options and specifications [can be found here](https://github.com/freedomofpress/pressfreedomtracker.us/blob/cc44f84da388ea25f5326b3b03a597a753e3d60d/chart_pregenerator/README.md).

### Technical details
This is a node application that essentially renders charts server-side. To accomplish this, there is some code to handle running the service, but it simply copies over the relevant chart react components into a subdirectory named `client/` to access the code there. This approach is the most performant and reliable, as it cuts through a separate build process or convoluted `npm link` logistics.

## Django capabilities

1. New settings. Environment variables `DJANGO_CHART_HOST` and `DJANGO_CHART_PORT` must be set or the integration will not function correctly.  These correspond to the hostname and port the node chart pregenerator service is running on, respectively.  

2. An API for requesting snapshots.  The Django code refers to images generated by the pregenerator service as "snapshots" because they represent the chart as they existed at a particular moment in time ("live" charts as they appear on the website will change as new data is added to the incidents database).  In `common.utils.chart_pregenerator` there is an `api` module that exposes a `request_snapshot` function.  There is documentation for this function but the gist is that you pass in what type of image you want, what type of chart you want, and any particular options you want for that chart type, and this function will make the request to the pregeneration service and give back either a string of SVG text, or an instance of Django's `ImageFile`.  The intent here is that basically all Python code that talks to the pregenerator API should go through this function.

3. A Django model for saving snapshots.  Once a snapshot has been generated, we want to save that image for some length of time.  The `ChartSnapshot` model is added to `common.models.charts` that follows the general architecture laid out by Harris in the discussion on this PR.  The main thing you want to do with this model is call `ChartSnapshot.get_or_generate` with the same three parameters that the API would accept: image type, chart type, and additional options.  The model checks if it has an non-stale image meeting that criteria in the DB.  If so, it returns that image, otherwise it passes the parameters along to the API and saves the resulting image.  **Note**: PNG-format images are saved as _Wagtail_ images (i.e. the `ChartSnapshot` has a foreign-key to `get_image_model()`).  This is so they can be used by `wagtail-metadata`.

4. Updates to `BlogPage.get_meta_image` and `BlogIndexPageFeed._get_teaser_image`.  These functions have been modified to check `BlogPage.teaser_graphic` and if a chart block is specified, look the corresponding snapshot up with `ChartSnapshot`.

5. A way for chart block values to talk to the pregenerator API.  Users of the wagtail admin define charts by adding a `VerticalBarChart`, `BubbleMapChart`, or `TreeMapChart` block to a stream field.  These are `StructBlock` classes that need to be able to, essentially, get a snapshot of themselves.  Their data is stored in the DB as JSON in a particular format, and we need it in a slightly different format to communicate with the API.  In `common.utils.charts` I've defined a `ChartValue` that each of these blocks can use as their `Meta.value_class` to facilitate this communication and transformation.  `value_class` is Wagtail's way of letting us add functionality to blocks: when you access a `StreamField`'s blocks, what you get back is an instance of the value class.  In this case: there is a subclass of `ChartValue` for each kind of chart

The way `ChartValue` works is it has methods available for each type of chart we might want to generate (SVG, PNG, mini PNG, etc.), as well as an `options_schema` for each one that points to a `marshmallow` schema class (e.g. `VerticalBarChartOptionsSchema`) that defines how to map the block data to the snapshot data.  This may sound more complicated than it is, but the marshmallow schema allows us to write the transformation declaratively and reuse code more easily between the different chart types.

6. Chart health endpoint: http://localhost:8000/admin/check_chart_health/ will give information about the node pregeneration service's status.

## Miscellaneous extras

1. More efficently use `curl` on CircleCI when testing the app for readiness.
2. Minor update to `createdevdata` command's image assignment logic.

## Testing and acceptance

1. Does the node service work locally for you?  Does it have the expected results for the different chart types, image types, and various options?

2. Does the chain of events around pregenerating blog teaser images as social media metadata work?  Can you add a chart to one or more blog posts and they're all properly created, shown in the right place, saved & loaded correctly, and display how we want them to?


dkaoster:
This is still a little bit rough for now, but wanted to get a draft PR up in order to start having some discussion around this.

Namely, here are things that still have yet to be resolved:
- [x] Integrate this with the `docker compose` script so that it doesn't live on it's own
- [x] Find a way to provide display options to the graphic, likely with query parameters
- [x] Find a way to render the big charts (react-animated-dataset compatibility is a challenge)
- [x] Write some tests for the service
- [ ] Figure out how to deploy this onto our production infrastructure

dkaoster:
Further conversation and consideration points:
- Do we want this service to be publicly accessible? If so, what is the best way of securing it so that only our services can generate images? If not, how will we trigger from django?

harrislapiroff:
Here's a quick sketch of how I see this looking on the Python side. Please feel free to rearrange any of the logic, add any needed models, question everything, etc. I haven't tested any of this so I don't know if it works or makes sense—do as I meant, not as I did ;)

## models/chartrendition.py

```python
class ChartRendition(models.Model):
    """
    This model represents one instance of a chart saved as either a PNG or SVG.
    
    A couple implementation notes:

    * Currently the query parameters are specified by the `chartType` and `query` fields,
      but you could also imagine using a stream block field instead and share the data
      format with our chart stream blocks.
    * Each instance only encodes either an image or an SVG (as specified by the
      `renditionType` field) but we could consider always generating both and using
      the same rendition for either type. However, the node service as currently written
      generates one or the other, so I figured I'd stick with that model.
    """
    last_generated = models.DateTimeField(auto_now=True)

    # These are the parameters that are used to generate the chart
    chart_type = models.CharField(max_length=255, choices=CHART_TYPES)  # Will need to create a list of chart types
    query = models.JSONField()
    rendition_type = models.CharField(
        max_length=3,
        choices=[
            ('PNG', 'Raster (PNG)'),
            ('SVG', 'SVG')
        ]
    )

    # These are the "cached" fields--things generated by the node service
    # that we then save
    chartImage = models.ImageField(null=True)
    chartSvg = models.TextField(null=True)

    class meta:
        unique_together = ['chart_type', 'query', 'rendition_type']

    def is_stale(self):
        """
        Returns true if the chart is stale and needs to be regenerated.
        """
        pass  # logic here to determine if chart needs to be regenerated

    def generate(self):
        """
        Generates an image or SVG code given the instance's parameters.
        """
        # This is where we call out to the node service
        pass 

    @classmethod
    def get_or_generate(cls, chart_type, query, rendition_type):
        """
        Generates a chart rendition given the parameters.
        """
        try:
            chart = cls.objects.get(chart_type=chart_type, query=query, rendition_type=rendition_type)
        except cls.DoesNotExist:
            chart = cls.objects.create(
                chart_type=chart_type,
                query=query,
                rendition_type=rendition_type
            )
            chart.generate()

        if chart.is_stale():
            chart.generate()

        return chart
```

## templatetags/chart.py

```python
@register.simple_tag
def chart(chart_type, query, rendition_type='PNG'):
    rendition = ChartRendition.get_or_generate(chart_type, query, rendition_type)
    # It's possible we'll want to add an option to return the full rendition data
    # or the image field instead of just the URL or the SVG code. Something to think
    # about.
    if rendition_type == 'PNG':
        return rendition.chartImage.url
    return rendition.chartSvg
    
```

You could then use the tag in templates like so:

```html+django
{% load chart %}

{% if page.lead_graphic.0.block_type == 'some_chart_type' %}
    <meta property="og:image" content="{% chart 'some_chart_type' page.teaser_graphic.0.value.query 'PNG' %}" />
{% else %}
    <meta property="og:image" content="{% image page.lead_graphic.0.value fit-1500x1500 %}"
{% endif %}
```

Or in python like so:

```python
class BlogPage(MetadataPageMixin, MediaPageMixin, Page):
    # ...
    def get_meta_image(self):
        if self.lead_graphic and self.teaser_graphic[0].block_type == 'some_chart_type':
            return ChartRendition.get_or_generate('some_chart_type', self.teaser_graphic[0].value.query, 'PNG')
        return self.teaser_graphic[0].value
```

harrislapiroff:
@chigby Is working on the Django integration here: https://github.com/freedomofpress/pressfreedomtracker.us/tree/chart-gen-backend

chigby:
One thing I've been noticing on the node service is that it will encounter an error if you request a chart but filter the data so that it doesn't have any valid incidents in it, such as filtering on a nonexistent tag:

```
http://localhost:3000/treemap-chart.png?options={%22height%22:300,%22filterTags%22:%22does_not_exist%22}
```

I've also noticed that in the event of such an error, the whole service stops and has to be restarted. Is there a way around this?

dkaoster:
@chigby Weird, that URL works fine for me without anything crashing. But nevertheless, I can add a restart on crash flow to the dev script so you don't have to restart the whole app.

harrislapiroff:
Restart on crash is also something we can configure docker compose (on dev)/k8s (on prod) to handle, right? Though it would be nice to get to the bottom of the crash if we can.

chigby:
Okay, I'm not seeing that URL I posted causing problems anymore. Maybe I was using a slightly older version of the branch, not sure, but not seeing a huge need to investigate it. 

Here's some other thoughts:

1. Placeholder/default images. Even in the best of times, our chart pregeneration service might fail occasionally. In that case, it might be useful to have some kind of "default" or "placeholder" image that can be used if a template somewhere asks the service for an image and none can be obtained. Anyone have ideas for what this image should be? It might have to be configured via a site setting.

2. I've found another scenario that causes an error in the pregeneration server: the `branch` field on tree map charts. Right now, this field only works if you set its value to the string `"categories"`, which gets interpolated into the URL `/api/edge/${options.branch}/` to request the values for that branch. But the other options for a tree map to branch on, status of charges and assailant, don't correspond to an endpoint, so making a request into the pregeneration service with one of those branch options causes an error: http://localhost:3000/treemap-chart.png?options={%22height%22:300,%22branch%22:%22assailant%22}

    I think the solution here is to handle the branch option more like it's being handled on the interactive chart, which treats `branch` as an object rather than a string and differentiates between branches where the values are obtained via a URL and those that aren't. [This is the code](https://github.com/freedomofpress/pressfreedomtracker.us/blob/chart-gen-backend/client/charts/js/tree-map-chart.js#L30-L39). I'd love it if there were a way to somehow share more of this "preliminary" code between the interactive charts and the snapshots/pregenerated ones, but there are some architectural differences that might make that infeasible.

3. Something I noticed, and this may also fall under the category of things that we can't immediately do anything about, but it may end up causing problems in the future. When running the python test suite, the test cases use a special test database and server that only exist during the test execution. However, the pregeneration code always talks to the "actual" development Django instance, running on port 8000 and connected to the development database. This definitely confused me for a little bit, since things I was doing to my test environment had no effect on what the node code was seeing when it was asking Django for data. Just putting this out there!

4. Speaking of tests, did we decide on if we wanted to have some JS ones written specifically for the pregeneration code? My python tests mostly treat the service as a black box.

dkaoster:
> Placeholder/default images. Even in the best of times, our chart pregeneration service might fail occasionally. In that case, it might be useful to have some kind of "default" or "placeholder" image that can be used if a template somewhere asks the service for an image and none can be obtained. Anyone have ideas for what this image should be? It might have to be configured via a site setting.

I've gone ahead and added a simple blank white fallback in the case of an error as the default, if we have something else in mind I'm happy to switch it over.

> I've found another scenario that causes an error in the pregeneration server: the branch field on tree map charts. Right now, this field only works if you set its value to the string "categories", which gets interpolated into the URL /api/edge/${options.branch}/ to request the values for that branch. But the other options for a tree map to branch on, status of charges and assailant, don't correspond to an endpoint, so making a request into the pregeneration service with one of those branch options causes an error: [http://localhost:3000/treemap-chart.png?options={%22height%22:300,%22branch%22:%22assailant%22}](http://localhost:3000/treemap-chart.png?options=%7B%22height%22:300,%22branch%22:%22assailant%22%7D)
>
> I think the solution here is to handle the branch option more like it's being handled on the interactive chart, which treats branch as an object rather than a string and differentiates between branches where the values are obtained via a URL and those that aren't. [This is the code](https://github.com/freedomofpress/pressfreedomtracker.us/blob/chart-gen-backend/client/charts/js/tree-map-chart.js#L30-L39). I'd love it if there were a way to somehow share more of this "preliminary" code between the interactive charts and the snapshots/pregenerated ones, but there are some architectural differences that might make that infeasible.

Good catch on this, I've gone ahead and updated the logic to be more similar to how you have `branches` and `branchFieldName`.

> Something I noticed, and this may also fall under the category of things that we can't immediately do anything about, but it may end up causing problems in the future. When running the python test suite, the test cases use a special test database and server that only exist during the test execution. However, the pregeneration code always talks to the "actual" development Django instance, running on port 8000 and connected to the development database. This definitely confused me for a little bit, since things I was doing to my test environment had no effect on what the node code was seeing when it was asking Django for data. Just putting this out there!

Do you have an example of how I can connect to the test database and service when running the test cases? I think if we're able to override the environment variables to direct it to the right place, it could work?

> Speaking of tests, did we decide on if we wanted to have some JS ones written specifically for the pregeneration code? My python tests mostly treat the service as a black box.

Good point, I'll work on getting some added in!

dkaoster:
@chigby Looks like the safety check is failing for something sanitization related, and test coverage is down too, otherwise I've cleaned up the js related tests and we should be ready for a review!

chigby:
Checking on the coverage & bandit situation now

chigby:
Everything should be passing now

chigby:
Seems to have started failing again, I'll take a look

chigby:
The dev job is failing because the pregeneration service isn't starting, with this error:

```
tracker-node-chart-pregenerator-1  | ✘ [ERROR] Could not resolve "./now"
tracker-node-chart-pregenerator-1  | 
tracker-node-chart-pregenerator-1  |     ../node_modules/lodash/debounce.js:2:18:
tracker-node-chart-pregenerator-1  |       2 │     now = require('./now'),
tracker-node-chart-pregenerator-1  |         ╵                   ~~~~~~~
tracker-node-chart-pregenerator-1  | 
tracker-node-chart-pregenerator-1  | ✘ [ERROR] Could not resolve "./toNumber"
tracker-node-chart-pregenerator-1  | 
tracker-node-chart-pregenerator-1  |     ../node_modules/lodash/debounce.js:3:23:
tracker-node-chart-pregenerator-1  |       3 │     toNumber = require('./toNumber');
tracker-node-chart-pregenerator-1  |         ╵                        ~~~~~~~~~~~~
tracker-node-chart-pregenerator-1  | 
tracker-node-chart-pregenerator-1  | /django/chart_pregenerator/node_modules/esbuild/lib/main.js:1639
tracker-node-chart-pregenerator-1  |   let error = new Error(`${text}${summary}`);
tracker-node-chart-pregenerator-1  |               ^
tracker-node-chart-pregenerator-1  | 
tracker-node-chart-pregenerator-1  | Error: Build failed with 2 errors:
tracker-node-chart-pregenerator-1  | ../node_modules/lodash/debounce.js:2:18: ERROR: Could not resolve "./now"
tracker-node-chart-pregenerator-1  | ../node_modules/lodash/debounce.js:3:23: ERROR: Could not resolve "./toNumber"
tracker-node-chart-pregenerator-1  |     at failureErrorWithLog (/django/chart_pregenerator/node_modules/esbuild/lib/main.js:1639:15)
tracker-node-chart-pregenerator-1  |     at /django/chart_pregenerator/node_modules/esbuild/lib/main.js:1051:25
tracker-node-chart-pregenerator-1  |     at /django/chart_pregenerator/node_modules/esbuild/lib/main.js:996:52
tracker-node-chart-pregenerator-1  |     at buildResponseToResult (/django/chart_pregenerator/node_modules/esbuild/lib/main.js:1049:7)
tracker-node-chart-pregenerator-1  |     at /django/chart_pregenerator/node_modules/esbuild/lib/main.js:1078:16
tracker-node-chart-pregenerator-1  |     at responseCallbacks.<computed> (/django/chart_pregenerator/node_modules/esbuild/lib/main.js:700:9)
tracker-node-chart-pregenerator-1  |     at handleIncomingPacket (/django/chart_pregenerator/node_modules/esbuild/lib/main.js:755:9)
tracker-node-chart-pregenerator-1  |     at Socket.readFromStdout (/django/chart_pregenerator/node_modules/esbuild/lib/main.js:676:7)
tracker-node-chart-pregenerator-1  |     at Socket.emit (node:events:527:28)
tracker-node-chart-pregenerator-1  |     at addChunk (node:internal/streams/readable:315:12) {
tracker-node-chart-pregenerator-1  |   errors: [
tracker-node-chart-pregenerator-1  |     {
tracker-node-chart-pregenerator-1  |       detail: undefined,
tracker-node-chart-pregenerator-1  |       id: '',
tracker-node-chart-pregenerator-1  |       location: {
tracker-node-chart-pregenerator-1  |         column: 18,
tracker-node-chart-pregenerator-1  |         file: '../node_modules/lodash/debounce.js',
tracker-node-chart-pregenerator-1  |         length: 7,
tracker-node-chart-pregenerator-1  |         line: 2,
tracker-node-chart-pregenerator-1  |         lineText: "    now = require('./now'),",
tracker-node-chart-pregenerator-1  |         namespace: '',
tracker-node-chart-pregenerator-1  |         suggestion: ''
tracker-node-chart-pregenerator-1  |       },
tracker-node-chart-pregenerator-1  |       notes: [],
tracker-node-chart-pregenerator-1  |       pluginName: '',
tracker-node-chart-pregenerator-1  |       text: 'Could not resolve "./now"'
tracker-node-chart-pregenerator-1  |     },
tracker-node-chart-pregenerator-1  |     {
tracker-node-chart-pregenerator-1  |       detail: undefined,
tracker-node-chart-pregenerator-1  |       id: '',
tracker-node-chart-pregenerator-1  |       location: {
tracker-node-chart-pregenerator-1  |         column: 23,
tracker-node-chart-pregenerator-1  |         file: '../node_modules/lodash/debounce.js',
tracker-node-chart-pregenerator-1  |         length: 12,
tracker-node-chart-pregenerator-1  |         line: 3,
tracker-node-chart-pregenerator-1  |         lineText: "    toNumber = require('./toNumber');",
tracker-node-chart-pregenerator-1  |         namespace: '',
tracker-node-chart-pregenerator-1  |         suggestion: ''
tracker-node-chart-pregenerator-1  |       },
tracker-node-chart-pregenerator-1  |       notes: [],
tracker-node-chart-pregenerator-1  |       pluginName: '',
tracker-node-chart-pregenerator-1  |       text: 'Could not resolve "./toNumber"'
tracker-node-chart-pregenerator-1  |     }
tracker-node-chart-pregenerator-1  |   ],
tracker-node-chart-pregenerator-1  |   warnings: []
tracker-node-chart-pregenerator-1  | }
```

@dkaoster Does this make sense to you? It's working for me locally, I wonder why this is showing on up on CircleCI.

dkaoster:
@chigby looks like it has something to do with the way that I'm trying to optimize build times, let me try a few things

chigby:
Thank you!

chigby:
Current status of this issue:
1. Waiting for experimental branch readiness (which is going through network/Cloudflare troubleshooting from @maeve-fpf)
2. Waiting for code review from @harrislapiroff

maeve-fpf:
I'm seeing the following error on the deployed `chart-pregenerator` branch in experimental:

```
Traceback (most recent call last):
  File "/django/common/models/charts.py", line 159, in get_or_generate
    chart = cls.objects.get(chart_type=chart_type, query=query, snapshot_type=snapshot_type)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/query.py", line 435, in get
    raise self.model.DoesNotExist(
common.models.charts.ChartSnapshot.DoesNotExist: ChartSnapshot matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 204, in _get_response
    response = response.render()
  File "/usr/local/lib/python3.9/site-packages/django/template/response.py", line 105, in render
    self.content = self.rendered_content
  File "/usr/local/lib/python3.9/site-packages/django/template/response.py", line 83, in rendered_content
    return template.render(context, self._request)
  File "/usr/local/lib/python3.9/site-packages/django/template/backends/django.py", line 61, in render
    return self.template.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 170, in render
    return self._render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/loader_tags.py", line 150, in render
    return compiled_parent._render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/loader_tags.py", line 150, in render
    return compiled_parent._render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/loader_tags.py", line 62, in render
    result = block.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/loader_tags.py", line 62, in render
    result = block.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/defaulttags.py", line 315, in render
    return nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/loader_tags.py", line 195, in render
    return template.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 172, in render
    return self._render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/defaulttags.py", line 315, in render
    return nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/loader_tags.py", line 193, in render
    return template.render(context.new(values))
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 172, in render
    return self._render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/defaulttags.py", line 519, in render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/defaulttags.py", line 214, in render
    nodelist.append(node.render_annotated(context))
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/loader_tags.py", line 193, in render
    return template.render(context.new(values))
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 172, in render
    return self._render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/defaulttags.py", line 519, in render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/defaulttags.py", line 315, in render
    return nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 988, in render
    output = self.filter_expression.resolve(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 671, in resolve
    obj = self.var.resolve(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 796, in resolve
    value = self._resolve_lookup(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 858, in _resolve_lookup
    current = current()
  File "/django/common/utils/charts.py", line 37, in png_snapshot_mini_url
    return self.png_snapshot_mini().get_rendition('original').url
  File "/django/common/utils/charts.py", line 29, in png_snapshot_mini
    snapshot = ChartSnapshot.get_or_generate(
  File "/django/common/models/charts.py", line 166, in get_or_generate
    chart.generate()
  File "/django/common/models/charts.py", line 130, in generate
    custom_image = get_image_model().objects.create(
  File "/usr/local/lib/python3.9/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/query.py", line 453, in create
    obj.save(force_insert=True, using=self.db)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/base.py", line 739, in save
    self.save_base(using=using, force_insert=force_insert,
  File "/usr/local/lib/python3.9/site-packages/django/db/models/base.py", line 776, in save_base
    updated = self._save_table(
  File "/usr/local/lib/python3.9/site-packages/django/db/models/base.py", line 881, in _save_table
    results = self._do_insert(cls._base_manager, using, fields, returning_fields, raw)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/base.py", line 919, in _do_insert
    return manager._insert(
  File "/usr/local/lib/python3.9/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/query.py", line 1270, in _insert
    return query.get_compiler(using=using).execute_sql(returning_fields)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1415, in execute_sql
    for sql, params in self.as_sql():
  File "/usr/local/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1358, in as_sql
    value_rows = [
  File "/usr/local/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1359, in <listcomp>
    [self.prepare_value(field, self.pre_save_val(field, obj)) for field in fields]
  File "/usr/local/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1359, in <listcomp>
    [self.prepare_value(field, self.pre_save_val(field, obj)) for field in fields]
  File "/usr/local/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1310, in pre_save_val
    return field.pre_save(obj, add=True)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/fields/files.py", line 302, in pre_save
    file.save(file.name, file.file, save=False)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/fields/files.py", line 89, in save
    self.name = self.storage.save(name, content, max_length=self.field.max_length)
  File "/usr/local/lib/python3.9/site-packages/django/core/files/storage.py", line 53, in save
    name = self.get_available_name(name, max_length=max_length)
  File "/usr/local/lib/python3.9/site-packages/storages/backends/gcloud.py", line 345, in get_available_name
    return super().get_available_name(name, max_length)
  File "/usr/local/lib/python3.9/site-packages/django/core/files/storage.py", line 91, in get_available_name
    while self.exists(name) or (max_length and len(name) > max_length):
  File "/usr/local/lib/python3.9/site-packages/storages/backends/gcloud.py", line 248, in exists
    return bool(self.bucket.get_blob(name))
  File "/usr/local/lib/python3.9/site-packages/google/cloud/storage/bucket.py", line 1276, in get_blob
    blob.reload(
  File "/usr/local/lib/python3.9/site-packages/google/cloud/storage/_helpers.py", line 238, in reload
    api_response = client._get_resource(
  File "/usr/local/lib/python3.9/site-packages/google/cloud/storage/client.py", line 366, in _get_resource
    return self._connection.api_request(
  File "/usr/local/lib/python3.9/site-packages/google/cloud/storage/_http.py", line 73, in api_request
    return call()
  File "/usr/local/lib/python3.9/site-packages/google/api_core/retry.py", line 283, in retry_wrapped_func
    return retry_target(
  File "/usr/local/lib/python3.9/site-packages/google/api_core/retry.py", line 190, in retry_target
    return target()
  File "/usr/local/lib/python3.9/site-packages/google/cloud/_http/__init__.py", line 482, in api_request
    response = self._make_request(
  File "/usr/local/lib/python3.9/site-packages/google/cloud/_http/__init__.py", line 341, in _make_request
    return self._do_request(
  File "/usr/local/lib/python3.9/site-packages/google/cloud/_http/__init__.py", line 379, in _do_request
    return self.http.request(
  File "/usr/local/lib/python3.9/site-packages/google/auth/transport/requests.py", line 545, in request
    self.credentials.before_request(auth_request, method, url, request_headers)
  File "/usr/local/lib/python3.9/site-packages/google/auth/credentials.py", line 133, in before_request
    self.refresh(request)
  File "/usr/local/lib/python3.9/site-packages/google/oauth2/service_account.py", line 410, in refresh
    access_token, expiry, _ = _client.jwt_grant(
  File "/usr/local/lib/python3.9/site-packages/google/oauth2/_client.py", line 289, in jwt_grant
    response_data = _token_endpoint_request(
  File "/usr/local/lib/python3.9/site-packages/google/oauth2/_client.py", line 260, in _token_endpoint_request
    _handle_error_response(response_data, retryable_error)
  File "/usr/local/lib/python3.9/site-packages/google/oauth2/_client.py", line 69, in _handle_error_response
    raise exceptions.RefreshError(
google.auth.exceptions.RefreshError: ('invalid_grant: Invalid grant: account not found', {'error': 'invalid_grant', 'error_description': 'Invalid grant: account not found'})
```

Do you need a shared bucket set up for this, and auth configuration added to the deploys?

maeve-fpf:
Here's an additional type of exception:

```
Traceback (most recent call last):
  File "/django/common/models/charts.py", line 159, in get_or_generate
    chart = cls.objects.get(chart_type=chart_type, query=query, snapshot_type=snapshot_type)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/query.py", line 435, in get
    raise self.model.DoesNotExist(
common.models.charts.ChartSnapshot.DoesNotExist: ChartSnapshot matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 204, in _get_response
    response = response.render()
  File "/usr/local/lib/python3.9/site-packages/django/template/response.py", line 105, in render
    self.content = self.rendered_content
  File "/usr/local/lib/python3.9/site-packages/django/template/response.py", line 83, in rendered_content
    return template.render(context, self._request)
  File "/usr/local/lib/python3.9/site-packages/django/template/backends/django.py", line 61, in render
    return self.template.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 170, in render
    return self._render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/loader_tags.py", line 150, in render
    return compiled_parent._render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/loader_tags.py", line 150, in render
    return compiled_parent._render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/loader_tags.py", line 62, in render
    result = block.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/loader_tags.py", line 62, in render
    result = block.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/defaulttags.py", line 315, in render
    return nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/loader_tags.py", line 195, in render
    return template.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 172, in render
    return self._render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/defaulttags.py", line 315, in render
    return nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/loader_tags.py", line 193, in render
    return template.render(context.new(values))
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 172, in render
    return self._render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/defaulttags.py", line 519, in render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/defaulttags.py", line 214, in render
    nodelist.append(node.render_annotated(context))
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/loader_tags.py", line 193, in render
    return template.render(context.new(values))
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 172, in render
    return self._render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/defaulttags.py", line 519, in render
    return self.nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/defaulttags.py", line 315, in render
    return nodelist.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 988, in render
    output = self.filter_expression.resolve(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 671, in resolve
    obj = self.var.resolve(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 796, in resolve
    value = self._resolve_lookup(context)
  File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 858, in _resolve_lookup
    current = current()
  File "/django/common/utils/charts.py", line 37, in png_snapshot_mini_url
    return self.png_snapshot_mini().get_rendition('original').url
  File "/django/common/utils/charts.py", line 29, in png_snapshot_mini
    snapshot = ChartSnapshot.get_or_generate(
  File "/django/common/models/charts.py", line 166, in get_or_generate
    chart.generate()
  File "/django/common/models/charts.py", line 130, in generate
    custom_image = get_image_model().objects.create(
  File "/usr/local/lib/python3.9/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/query.py", line 453, in create
    obj.save(force_insert=True, using=self.db)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/base.py", line 739, in save
    self.save_base(using=using, force_insert=force_insert,
  File "/usr/local/lib/python3.9/site-packages/django/db/models/base.py", line 776, in save_base
    updated = self._save_table(
  File "/usr/local/lib/python3.9/site-packages/django/db/models/base.py", line 881, in _save_table
    results = self._do_insert(cls._base_manager, using, fields, returning_fields, raw)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/base.py", line 919, in _do_insert
    return manager._insert(
  File "/usr/local/lib/python3.9/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/query.py", line 1270, in _insert
    return query.get_compiler(using=using).execute_sql(returning_fields)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1415, in execute_sql
    for sql, params in self.as_sql():
  File "/usr/local/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1358, in as_sql
    value_rows = [
  File "/usr/local/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1359, in <listcomp>
    [self.prepare_value(field, self.pre_save_val(field, obj)) for field in fields]
  File "/usr/local/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1359, in <listcomp>
    [self.prepare_value(field, self.pre_save_val(field, obj)) for field in fields]
  File "/usr/local/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1310, in pre_save_val
    return field.pre_save(obj, add=True)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/fields/files.py", line 302, in pre_save
    file.save(file.name, file.file, save=False)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/fields/files.py", line 89, in save
    self.name = self.storage.save(name, content, max_length=self.field.max_length)
  File "/usr/local/lib/python3.9/site-packages/django/core/files/storage.py", line 54, in save
    name = self._save(name, content)
  File "/usr/local/lib/python3.9/site-packages/storages/backends/gcloud.py", line 210, in _save
    size=getattr(content, 'size', None),
  File "/usr/local/lib/python3.9/site-packages/django/utils/functional.py", line 48, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "/usr/local/lib/python3.9/site-packages/django/core/files/base.py", line 42, in size
    self.file.seek(0, os.SEEK_END)
io.UnsupportedOperation: seek
```

Unclear what's going on here, doing a `seek()` makes me think the problem is trying to use the container's read-only root for temporary files, but that stack trace shows it's still instide the django-storage gcloud backend.

soleilera:
Currently a 500 error being thrown on the front page and blog on experimental, other pages look fine. @maeve-fpf will add the QA credentials to the server so @chigby can look at the backend.

chigby:
To @maeve-fpf (or anyone else who might know): Wondering if there's a way to view the logs from `experimental.pressfreedomtracker.us` -- on logs.freedom.press the `resource.labels.cluster_name` options are only staging and production. Is there somewhere else I should be looking for that?

soleilera:
> To @maeve-fpf (or anyone else who might know): Wondering if there's a way to view the logs from `experimental.pressfreedomtracker.us` -- on logs.freedom.press the `resource.labels.cluster_name` options are only staging and production. Is there somewhere else I should be looking for that?

@chigby - are you able to see the experimental logs now? sounds like @maeve-fpf added those to the aggregate.

dkaoster:
@chigby So this week I tried to figure out why the stacked bar charts were not working properly with the chart pregenerator, and found a couple issues which I resolved in https://github.com/freedomofpress/pressfreedomtracker.us/pull/1693/commits/8e0813509c625ca850ad521b4111319471f2af8a and https://github.com/freedomofpress/pressfreedomtracker.us/pull/1693/commits/d506c437070a792a5b4d5442edba01ebd2a94611. However, this seems to have broken the tests for the chart pregenerator, which is showing up in the circleci as:

```
2023-10-19 00:16:53 [error    ] Chart pregeneration failed     [common.models.charts] 
Traceback (most recent call last):
  File "/django/common/models/charts.py", line 112, in generate
    result = request_snapshot(
  File "/usr/local/lib/python3.9/unittest/mock.py", line 1092, in __call__
    return self._mock_call(*args, **kwargs)
  File "/usr/local/lib/python3.9/unittest/mock.py", line 1096, in _mock_call
    return self._execute_mock_call(*args, **kwargs)
  File "/usr/local/lib/python3.9/unittest/mock.py", line 1151, in _execute_mock_call
    raise effect
common.utils.chart_pregenerator.api.PregenerationException
.....2023-10-19 00:16:53 [error    ] Chart pregeneration failed     [common.models.charts] 
Traceback (most recent call last):
  File "/django/common/models/charts.py", line 163, in get_or_generate
    chart = cls.objects.get(chart_type=chart_type, query=query, snapshot_type=snapshot_type)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/db/models/query.py", line 435, in get
    raise self.model.DoesNotExist(
common.models.charts.ChartSnapshot.DoesNotExist: ChartSnapshot matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/django/common/models/charts.py", line 112, in generate
    result = request_snapshot(
  File "/usr/local/lib/python3.9/unittest/mock.py", line 1092, in __call__
    return self._mock_call(*args, **kwargs)
  File "/usr/local/lib/python3.9/unittest/mock.py", line 1096, in _mock_call
    return self._execute_mock_call(*args, **kwargs)
  File "/usr/local/lib/python3.9/unittest/mock.py", line 1151, in _execute_mock_call
    raise effect
common.utils.chart_pregenerator.api.PregenerationException
```

I tried digging into the testing code to see if I can fix it, but I'm having a lot of trouble understanding how the testing code works. Can you help me figure out why the tests are failing?

.eslintignore
```diff
@@ -0,0 +1 @@
+chart_pregenerator
```
harrislapiroff: Is there a reason we can't lint our node code?

devops/docker/NodeChartPregeneratorDockerfile
```diff
@@ -0,0 +1,30 @@
+# sha256 as of 2023-09-08
+FROM node:20.8.0-alpine@sha256:80cc2c520781c1a4681e59df6791d81a432a6cb3da0c385d8d62e9f16acf8e5f
```
harrislapiroff, line 2: For whatever reason this image is not playing nice with my M1 MacBook Pro. I get:

```
node-chart-pregenerator The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
```

I was able to resolve it by switching this image to be the same one used by `NodeDockerfile` but presumably there's a better solution.

docker-compose.yaml
```diff
@@ -1,7 +1,12 @@
 ---
-version: "3"
+version: "3.4"
 networks:
   app:
+
+x-chart-pregenerator:
+  host: &chart-host ${NODE_CHART_PREGENERATOR_HOST:-chartgenerator}
+  port: &chart-port ${NODE_CHART_PREGENERATOR_PORT:-3000}
```
harrislapiroff, line 7: Do we need to do this instead of using docker compose's internal networking addresses? (It's possible this was necessary when we were loading things directly from the node service client side, but we decided not to do that, right?)

devops/docker/NodeChartPregeneratorDockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+# Make npm output less verbose
+ENV NPM_CONFIG_LOGLEVEL warn
+
+ARG USERID
+RUN getent passwd "${USERID?USERID must be supplied}" || adduser -D -g "" -u "${USERID}" pft_node
+COPY chart_pregenerator /opt/chart-pregenerator
+RUN chown -R "${USERID}" /opt/chart-pregenerator
+WORKDIR /opt/chart-pregenerator
+USER ${USERID}
```
harrislapiroff: What is this bit doing? Is it similar to the user id hacks we were doing to share node_modules between host and container? Do we need it?

docker-compose.yaml
```diff
@@ -53,6 +58,36 @@ services:
...
+      - target: *chart-port
+        published: *chart-port
+    working_dir: /django/chart_pregenerator
+    user: ${UID:?err}
+    healthcheck:
+      test: ["CMD", "curl", "--fail", "http://localhost:3000/"]
+      start_period: 10s
+      interval: 30s
+      timeout: 20s
+      retries: 4
```
harrislapiroff: love me a healthcheck

docker-compose.yaml
harrislapiroff: Or maybe we just need it for the port? Though arguably we could just hard code it like we're doing for the database...

prod-docker-compose.yaml
```diff
@@ -1,7 +1,12 @@
 ---
-version: "3"
+version: "3.4"
 networks:
   app:
+
+x-chart-pregenerator:
+  host: &chart-host ${NODE_CHART_PREGENERATOR_HOST:-chartgenerator}
+  port: &chart-port ${NODE_CHART_PREGENERATOR_PORT:-3000}
+
```
harrislapiroff, line 9: See questions on other docker-compose file

harrislapiroff (commented):
Functionality seems to work! Extremely cool. I haven't fully finished reviewing the code, but I thought I'd post my initial comments and questions before getting to that.

Some general notes:

* I'm not entirely sure how I feel about images supplanting client-side teaser images. It might be nice to either keep those as client-side generated or server-side generated SVG and *only* use the PNGs for social media cards.
* Testing this made me realize it's kind of annoying to fill in the teaser image/chart multiple times. It's nice to have the power to specify them separately but it may be that the tracker team will never actually want a different lead graphic vs teaser image. Let's ask them.

docker-compose.yaml
chigby: My thought here was that:

1. We want to be able to set the host and port in environment variables
2. We need to refer to these values in multiple places in this docker compose configuration

And setting these named anchors seemed like the least-repetitive way to do it. I think you might be right that we don't actually need the host here (it's not even being used elsewhere in the file), and for local development we can also probably do without reading the environment variables here at all, if that's what's bothering you.

devops/docker/NodeChartPregeneratorDockerfile
dkaoster: honestly not sure why this one is different than the one in `NodeDockerfile`, I think we should be able to use the same image?

devops/docker/NodeChartPregeneratorDockerfile
dkaoster: I did just copy it over from the other dockerfile, which I suspect we need for the same reason `NodeDockerfile` needs it, is there a reason we wouldn't need it here?

dkaoster:
> I'm not entirely sure how I feel about images supplanting client-side teaser images. It might be nice to either keep those as client-side generated or server-side generated SVG and only use the PNGs for social media cards.

I think that's fair, I suppose the best solution here might be to server-side generate the SVGs to be included and then have it "hydrate" on the client to become interactive?

Edit: After thinking about this some more, I think one strong argument for the teaser images being image files is that they should be much lighter on the client, especially if you imagine a blog listing page with every single blog post being a SVG, that could cause the entire page to be pretty heavy. My thinking is that since the teaser images are small and not expected to be interactive, there isn't much of a downside to having them as images?

blog/templates/blog/_blog_card.html
```diff
@@ -9,17 +9,13 @@
...
-				<div class="card-image card-image-{{ size }}">
-					{% include 'common/blocks/tree_map_chart.html' with value=photo.0.value mini_graphic=true static_graphic=true %}
-				</div>
-			{% elif photo.0.block_type == "bubble_map_chart" %}
-				<div class="card-image card-image-{{ size }}">
-					{% include 'common/blocks/bubble_map_chart.html' with value=photo.0.value mini_graphic=true static_graphic=true %}
+					<img
+						class="card-image card-image-{{ size }}"
+						alt="{{ photo.0.value.description }}"
+						src="{{ photo.0.value.svg_snapshot_mini_datauri }}"
```
harrislapiroff, line 17: In general we tend to use inline SVGs over image tags for SVG rendering, so might be something to consider revising in the future. I can't think of any concrete advantages either way at the moment though, so fine to leave as-is.

harrislapiroff (commented):
Just reviewed this with @chigby and @dkaoster explaining things to me. A few notes. 

Blockers (just one):

- [x] Since chart images can be regenerated up to once per day we should clean up old `Image`s from the database when regenerating. @chigby has volunteered to work on this. This may still leave image files accumulating in our storage buckets (I'm actually not sure on this point?), but if so that's fine for now. If it becomes a problem we could look into a management command for cleaning up unused media.

A few notes for posterity:

- @dkaoster used esbuild instead of webpack for this service. While I don't love adding Yet Another Build Tool to our pipeline, I do think we may switch to esbuild completely rather than webpack someday. (I've been using it recreationally and it has been much easier to configure than webpack.) So this seems like as good a way as any to introduce it to our project.
- In general as our sites acquire more custom-written services we may want to restructure the directories in our projects to reflect the service structure and make it clear what code is shared between services. Project for a future date.
- @dkaoster ran into some problems getting react-animated-dataset and d3 working server side and came up with some good workarounds. I'm excited about, at some future time, looking into solutions that will allow us to use these libraries frictionlessly on the server-side.
- Currently rendering a chart snapshot requires a full roundtrip between the django server and the image. Visitor requests a page from django app -> django app requests a chart snapshot from node service -> node service requests data from django app. I think this is a little overcomplicated and we could just send the data from django to the node service in the first request. @chigby mentioned he could see some reasons for doing it the way we are doing it now. In either case it's fine to leave as is for now and revisit in the future.

Some of the above should become new tickets after merge, to keep track.

soleilera:
This is currently blocked on #1795

harrislapiroff:
@chigby is QA'ing this. The experimental server seems to be working now! *knock on wood*

harrislapiroff:
New blocker: using NPM to run the service results in attempts to write files—not compatible with the read-only filesystem we're deploying to.

Attempted a few things including removing `npm` from the run at all and nothing worked.

Changes only get deployed when the files inside the service are changed—so none of those attempts were deployed. Were eventually deployed, still not working.

Can't seem to reproduce locally.

Debugging this is eating up a lot of our time, so let's put this on ice for now and @harrislapiroff will set up a call in the next week or so with @maeve-fpf and @chigby (@SaptakS optional) to debug together.

chigby:
As of now, this branch is set up so that the node chart pregeneration service uses the variables `DJANGO_HOST` and `DJANGO_PORT` to determine where to connect for fetching chart data. So these will need to be populated for that container on experimental/staging/production. cc @maeve-fpf

chigby:
Did some more checking on the experimental cluster and noticed [this error](https://logs.freedom.press/app/discover#/doc/filebeat-*/filebeat-gke-2024.04.23?id=jx3AC48BnrDkXCJW4p2A):

```
Invalid HTTP_HOST header: 'pft-wagtail.pressfreedomtracker.svc'. You may need to add 'pft-wagtail.pressfreedomtracker.svc' to ALLOWED_HOSTS.
```

So it looks like we have to add the contents of the env var `DJANGO_CHART_HOST` to either: (1) the `DJANGO_ALLOWED_HOSTS` env var or (2) the `ALLOWED_HOSTS` setting within Django settings.py code itself. @maeve-fpf Do you have a preference on how this ought to be configured?

maeve-fpf:
If you mean we can do, like

    ALLOWED_HOSTS += [CHART_HOST]

in settings and not duplicate something between two environment variables in the config, I'd rather do that.

chigby:
Great, thank you!

chigby:
This feature seems to be working as intended on https://experimental.pressfreedomtracker.us -- I think someone else should probably do some QA on this as well, and then I think it can be merged and assessed on staging.

chigby (approved):
Approving this as part of the overall review process that the rest of the web team signed off on in the Sync meeting today, even though this is a partial self-review, here.

PR #1697, opened 2023-07-19
Add placeholder for teaser images

### LINKED ISSUE #1689: More fixes to "primary chart teasers"
- [ ] Since they take a moment to load there's a visible content reflow. Let's add a placeholder before the charts load to prevent that as [described by @SaptakS here](https://github.com/freedomofpress/pressfreedomtracker.us/pull/1687#issuecomment-1621092734). Ultimately it would be really nice to generate these charts server-side with a node process... but there's some non-trivial complexity in doing that.
- [x] Let's add a slightly darker background to the charts so they appear to have a rectangular preview as [discussed here](https://github.com/freedomofpress/pressfreedomtracker.us/pull/1687#issuecomment-1621792506). We can go even a bit darker than the mockup.

dkaoster:
fixes #1689 

Adds a placeholder for teaser images so that the layout doesn't shift as much, also adds a gray background for the mini bar chart. I've left it as just a blank svg for now, but open to other placeholders if we want!

https://github.com/freedomofpress/pressfreedomtracker.us/assets/3477162/da85c4fc-483d-4338-b9c3-08a3dadd6a0e

### DIFF
--- a/client/charts/js/bubble-map-chart.js
+++ b/client/charts/js/bubble-map-chart.js
@@ -29,7 +29,11 @@ function engageCharts() {
 				dataKey={['dataset']}
 				dataParser={[(data) => d3.csvParse(data, d3.autoType)]}
 				loadingComponent={(
-					<svg viewBox="0 0 655 440" width="100%" style={{ display: fullSize ? 'none' : 'block' }} />
+					<svg
+						viewBox="0 0 655 440"
+						width="100%"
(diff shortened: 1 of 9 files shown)


SaptakS (commented):
Can we have a light grey placeholder, instead of a blank svg? I feel that might be better. cc @harrislapiroff

harrislapiroff:
I agree a light gray placeholder would be nice!

dkaoster:
Updated!

SaptakS (approved):

PR #1713, opened 2023-07-27
Add Stacked Bar Charts

### LINKED ISSUE #1705: Stacked Bar Chart
Let's support the option of adding a criteria on which to break both our bar chart blocks into "stacks" like so:

barchart vert

I think there's probably a lot of design considerations I haven't thought through here (what categories would work for stacking? do they need to be mutually exclusive categories? etc.) so feel free to make decisions on this as you implement and consult with @KAMcCudden and the tracker team on what they might find useful.

It's possible we can reuse some of the code from #1683 for this.

You can find documentation of how interactions should work in these bar charts in the [Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?type=design&node-id=2429-12782&mode=design&t=vYf3QtQiwohriiZr-4).

dkaoster:
closes #1705 

Adds a stacked bar chart to chart visualization options. Uses the segmentation from Treemap charts in #1683 to provide segmentation options. To test, open a blog post page and choose the new Stacked Bar Chart option.

Screenshot 2023-07-27 at 2 48 56 PM
Screenshot 2023-07-27 at 2 48 49 PM
Screenshot 2023-07-27 at 2 48 37 PM
Screenshot 2023-07-27 at 2 48 22 PM


harrislapiroff:
Wow this looks awesome. I can't wait to see this in action. (Not a review, just excited.)

chigby:
Is there a way to change what categorizations appear on the horizontal axis?

harrislapiroff:
@dkaoster Sorry to ask you to do an architecture restructure on this, but can we revisit this and instead of creating a new `StackedBarChart` type I'd like you to use the current `VerticalBarChart` type and add an _optional_ `group_by` field. If the `group_by` field isn't set it should behave as it does in the existing code and if it does have that field set it should provide the stacked chart you've coded here.

I think this will simplify the interface for editors and allow more code reuse for us.

harrislapiroff:
Also, just for the record here's a couple features for the future (ones that appear in the Figma but don't need to be implemented in this PR). We should open separate tickets for these:

- Select categories in the legend to render *only* those categories on the chart
- Customize the x-axis field

harrislapiroff (changes requested):
See my comment above

chigby:
If a grouping doesn't appear on the chart, should we still show it in the legend? E.g.:
image

harrislapiroff:
Yeah, let's exclude groupings with no incidents from appearing.

chigby:
I think the safety issue is resolved in `develop` now

common/blocks.py
```diff
@@ -500,10 +500,17 @@ class VerticalBarChart(blocks.StructBlock):
         'e.g., "Bar chart showing a decreasing number of assaults over the '
         'course of 2023."',
     )
+    group_by = blocks.ChoiceBlock(
+        label='Group Incidents By',
+        required=False,
+        choices=charts.IncidentBranches.choices,
+        default=None,
+    )
```
chigby, line 510: Might be nice here to add some help text explaining how this field changes how the chart is displayed. Though I'm not immediately coming up with a concise way of doing that.

incident/utils/charts.py
```diff
@@ -71,7 +71,7 @@ def branches(self):
         return json.dumps(branches_value)
 
 
-class TreeMapBranches(TextChoices):
+class IncidentBranches(TextChoices):
```
chigby, line 74: It does occur to me here that I chose the name "branches" for this concept because of how it related to the "tree" map chart, so I hope this terminology is still somewhat meaningful when applied to other visualizations. :sweat_smile:

chigby (approved):
With the updates to the display of the legend, I think this can be approved. I left a comment about adding some help text but I don't think that ought to block the merging of this. But if you think of something, it would probably be nice to have that there.

common/blocks.py
harrislapiroff: "If selected, will create a stacked bar chart colored by grouping" or something along those lines?

harrislapiroff (commented):
With @chigby's approval, I'd be happy to see this merged before deployment this week! I agree it would be nice to get in some help text before merge. (In general, I'm in favor of giving editors more help text since we don't exactly have a user's manual.)

common/blocks.py
dkaoster: @harrislapiroff how's this? 

"If selected, turns this vertical bar chart into a stacked bar chart with incidents grouped by the selected classification."

harrislapiroff: Sounds great!

chigby:
@harrislapiroff Still needs your approval also

harrislapiroff (approved):

harrislapiroff:
Whoops, approved. Merging now.

PR #1714, opened 2023-08-01
Fix Homepage Charts

### LINKED ISSUE #1711: Home Page Charts Misaligned
Screenshot 2023-07-26 at 5 30 10 PM

Sometime recently a change was made and now the charts on the homepage no longer line up properly.

dkaoster:
closes #1711 

Screenshot 2023-08-01 at 10 13 54 AM

Some of the earlier work getting the map to the right size caused some issues on the homepage alignment, this fixes that.

### DIFF
--- a/client/charts/js/components/BarChart.jsx
+++ b/client/charts/js/components/BarChart.jsx
@@ -137,15 +137,16 @@ export default function BarChart({
 				)}
 				<svg
 					ref={setSvgEl}
-					width={width}
-					height={height}
+					width="100%"
 					aria-labelledby={id}
 					style={{
 						marginTop: margins.top,
 						marginBottom: margins.bottom,
 						marginRight: margins.right,
 						marginLeft: margins.left,
+						display: "block",
 					}}
+					viewBox={[0, 0, width, height]}
 				>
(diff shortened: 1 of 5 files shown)


chigby:
Huh, this still seems slightly off on my machine: 

image

(in Firefox, same in Chromium)

chigby:
The map chart looks like its content is slightly smaller than the other two (again, on my machine). Just thinking out loud here. 
image

dkaoster:
looks like this is just flexbox being flexbox. Turns out [equal columns in flexbox is actually almost impossible to do reliably](https://css-tricks.com/equal-columns-with-flexbox-its-more-complicated-than-you-might-think/), so I guess I'll convert it to css grid.

dkaoster:
Okay, I've changed it to use css grid and modified the svgs to be actually responsive, that should do it!

chigby (approved):
Definitely looks aligned to me, great change!

chigby:
Not totally clear why https://github.com/freedomofpress/pressfreedomtracker.us/commit/9119b82216dd2ba69ea2202031b6ef2e9bf5c214 is not showing up in this PR's list of commits.

PR #1724, opened 2023-08-17
Fix homepage charts bar chart width

### LINKED ISSUE #1723: Home page charts responsive rendering between 1000 and 1018 px
While looking into #1720, I noticed something off about how the home page charts are displayed when the viewport in the range of 1000px to 1018px (inclusive).

Example from Firefox, but I also see this in Chromium:

image

dkaoster:
## Description

Fixes #1723 

Fixes the bar chart being the wrong width from about 1000px to 1018px.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

- Check out this branch, run the application and visit http://localhost:8000/
- Resize the browser viewport to about 1000px
- Verify that the width of all the charts are correct

### Post-deployment actions

### General checks

Screenshot 2023-08-17 at 9 24 27 AM

### DIFF
--- a/client/charts/js/components/BarChart.jsx
+++ b/client/charts/js/components/BarChart.jsx
@@ -343,8 +343,8 @@ export default function BarChart({
 		return (
 			<svg
 				ref={setSvgEl}
-				width={width}
-				height={height}
+				width="100%"
+				viewBox={[0, 0, width, height]}
 				style={{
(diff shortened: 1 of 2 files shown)


chigby (approved):
Fix works great for me

PR #1727, opened 2023-08-26
Add Database Filter Summary

### LINKED ISSUE #1706: Add filter summary to top of database page
Let's add little tokens showing a summary of applied filters to the database page. See the design in [Figma](https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?type=design&node-id=4403-103654&mode=design&t=JJbfhah8AHfakbC4-4).

Image

Initially we weren't going to take on this ticket this year, but I think this is a good one for someone to take on if we run out of work while I'm on PTO. I think it'll be relatively quick and a pretty good UX win. Otherwise we may push it back to a future time.

dkaoster:
## Description

Fixes #1706 

This PR adds the current filters as little pills to the top of the database page, allowing users to clear specific filters by clicking on each pill. 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

- Run the application locally, and test adding some filters on this page: http://localhost:8000/all-incidents/
- The summary should show up above the database entries.
- Ensure that desktop, mobile, and tablet all look good

### Post-deployment actions

### General checks

Screenshot 2023-08-25 at 4 54 19 PM

### DIFF
--- a/client/charts/js/components/FilterSummary.jsx
+++ b/client/charts/js/components/FilterSummary.jsx
@@ -48,7 +48,6 @@ export default function FilterSidebar() {
 							"categories", categories.filter(d => d !== category).join(',')
 						)}
 						className="btn btn-tag"
-						tabIndex={0}
 						aria-label={`Removes filter: ${category.toLowerCase()}`}
 					>
 						<div className={classNames("category", `category-${symbolMap[category]}`)}></div>
@@ -62,7 +61,6 @@ export default function FilterSidebar() {
 					<button
 						onClick={() => clearFilter("date_lower")}
 						className="btn btn-tag"
-						tabIndex={0}
 						aria-label={`Removes start date: ${date_lower}`}
 					>
 						<span>from: {date_lower}</span>
@@ -75,7 +73,6 @@ export default function FilterSidebar() {
 					<button
 						onClick={() => clearFilter("date_upper")}
 						className="btn btn-tag"
-						tabIndex={0}
 						aria-label={`Removes end date: ${date_upper}`}
 					>
 						<span>to: {date_upper}</span>
@@ -87,10 +84,9 @@ export default function FilterSidebar() {
 				<li key={tag}>
 					<button
 						onClick={() => clearFilter(
-							"tags", categories.filter(d => d !== tag).join(',')
+							"tags", tags.filter(d => d !== tag).join(',')
 						)}
 						className="btn btn-tag"
-						tabIndex={0}
 						aria-label={`Removes tag: ${tag.toLowerCase()}`}
 					>
 						<span>{tag.toLowerCase()}</span>
@@ -103,7 +99,6 @@ export default function FilterSidebar() {
 					<button
 						onClick={() => clearFilter(filterKey)}
 						className="btn btn-tag"
-						tabIndex={0}
 						aria-label={`Removes filter: ${restFilters[filterKey].replaceAll("_", " ").toLowerCase()}`}
 					>
 						<span>{restFilters[filterKey].replaceAll("_", " ").toLowerCase()}</span>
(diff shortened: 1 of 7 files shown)


client/charts/js/components/FilterSummary.jsx
```diff
@@ -0,0 +1,111 @@
...
+	return (
+		<ul className="filters-summary">
+			{categories.map(category => (
+				<li key={category}>
+					<button
+						onClick={() => clearFilter(
+							"categories", categories.filter(d => d !== category).join(',')
+						)}
+						className="btn btn-tag"
+						tabIndex={0}
```
SaptakS: A quick comment before I review the rest of the PR, can we probably add an 
```js
aria-label={`Removes filter: ${category.toLowerCase()}`}
```
or something similar to really convey the intent of the buttons? Because right now it doesn't convey to the user whether clicking the button is going to close th button, or open a new page with the filters or something else.

cc @harrislapiroff in case you have a better idea.

harrislapiroff: Seems reasonable to me!

client/charts/js/components/FilterSummary.jsx
```diff
@@ -0,0 +1,116 @@
...
```
SaptakS: There is no need to add `tabIndex=0` on a button I think since it is already interactive.

client/charts/js/components/FilterSummary.jsx
```diff
@@ -0,0 +1,116 @@
...
+						<span>to: {date_upper}</span>
+						<span className="close-icon" />
+					</button>
+				</li>
+			)}
+			{tags.map(tag => (
+				<li key={tag}>
+					<button
+						onClick={() => clearFilter(
+							"tags", categories.filter(d => d !== tag).join(',')
```
SaptakS: I think this should be `tags.filter...` ?

client/common/sass/components/_filters.sass
```diff
@@ -127,3 +127,31 @@
...
+	+desktop-up
+		&:not(:empty)
+			margin-bottom: 32px
+
+	+tablet-down
+		margin-bottom: 8px
+
+	li
+		display: inline-block
+		margin: 2px 2px 2px 0
```
SaptakS: Small knitpick but according to the figma, the margin between each element is 8px

SaptakS (changes requested):
I have left a few changes requested. After that, I think this PR should be ready to be merged.

SaptakS:
This is definitely not an issue with this PR, and I am also definitely guilty in doing this, but I feel somewhere along the line, we might have lost track of what components we puts incide `client/charts/js/components` vs `client/common/js/components` (or maybe it's just me not remembering a decision we consciously made). According to me, I feel only the chart related components should be in `clients/charts/`, and all the other components such as FilterSummary, FilterSidebar should be inside `clients/common`. @chigby do you remember if we did this because it would be easier to integrate with the charts? I remember we got some react for the charts to begin with and hence put them inside charts.

Again, this is definitely not a comment on this PR, and probably a refactoring discussion. cc: @harrislapiroff

harrislapiroff:
Hmmm I think `clients/charts/` only ever existed to give Accurat an isolated area to develop their charts in, so, prior to your comment, my instinct would have been to refactor to move all the js to one directory. If we're finding the distinction helpful though, we could come up with some criteria though. What do you all think?

SaptakS:
I feel like keeping charts in a separate folder make it easier to discover in the codebase, but definitely don't think it's necessary (especially since we have kind of been using it all in one place and there is no distinction as of now).

dkaoster:
I also think it could be cool in a future ticket to move to it's own folder, for organization's sake.

SaptakS (approved):

PR #1728, opened 2023-08-28
Fix Chart Downloader

dkaoster:
## Description

Fixes the chart downloader.

Before:
Searched and seized equipment of journalists across the U S

After: 
chart

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

Run the application locally, and try downloading any chart. The downloaded chart should look correct.

### General checks

### DIFF
--- a/client/charts/js/components/ChartDownloader.jsx
+++ b/client/charts/js/components/ChartDownloader.jsx
@@ -91,7 +91,7 @@ const ChartDownloader = ({
 							</text>
 						` : ""}
 					` : ""}
-					<svg width="${imageWidth}" height="${imageHeight}" viewBox="0 0 ${svgWidth} ${svgHeight}">
+					<svg width="${imageWidth}" height="${imageHeight}">
 						${svgStringData}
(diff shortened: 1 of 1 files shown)


chigby:
For me, this change does fix the problem of the chart being cropped.

chigby:
Though I am also noticing that sometimes the title can be cut off of the image if the title is too long: 
Camera family reach owner college unit add nearly learn(2)

Not sure if this is related to the same problem of if it's a separate issue.

dkaoster:
> Though I am also noticing that sometimes the title can be cut off of the image if the title is too long:
> 
> Not sure if this is related to the same problem of if it's a separate issue.

This is a separate issue, unfortunately automatic text wrapping is not supported in SVGs, so there are some workarounds but they are all suboptimal. If this is something we want to address I think it'd make sense to cut a new ticket.

chigby (approved):
Fix works great!

PR #1733, opened 2023-08-29
Add in-app defaults for chart host/port, remove from docker build time

maeve-fpf:
## Description

We don't have a place to add additional `CHART_HOST` and `CHART_PORT` build args in the Docker build process in CI. These don't make sense to me as args; the collectstatic run doesn't actually need them, the app just bails out on boot if nothing is specified and they're there to prevent that.

Instead, fall back to default values in the Django config code. To do this, `.get()` didn't work as expected; I had to use `getattr()`. Is that part acceptable?


chigby (approved):
This is fine, and honestly probably better than what I had there before-- my thinking was that default values would "live" in the docker configuration but it seems just as well for them to be in the settings.

chigby:
Looking over this again, I'm realizing that some of my tests of the code that was changed are failing. I can update those if you'd like.

Also, I'm slightly confused about the `.get()` vs. `getattr()` situation. Can you say more about what wasn't working? 
https://github.com/freedomofpress/pressfreedomtracker.us/blob/8ac9275aead6025bf1be9a019c4c5db4f2c0b710/common/utils/chart_pregenerator/config.py#L7-L16

The Django settings have an attribute called `CHART_PREGENERATOR` so using `getattr()` there (line 8) should work. The value of that attribute is a dictionary, so I would expect we would need `.get()` to access the `'HOST'` and `'PORT'` keys on lines 12 and 16. As it stands in the code above, I think using `getattr()` for those attributes on a dictionary will always yield the default values, even if the underlying settings change.

PR #1734, opened 2023-08-31
Improve CSRF error handling for FormPage submissions

### LINKED ISSUE #1674: Improve CSRF error handling on contact forms
[updated 6/15/2023]

Based on the conversation below this ticket is now to improve error handling on CSRF errors for contact forms:

- [ ] CSRF errors should show _on_ the form itself as form errors
- [ ] The form should retain user input so they can attempt to resubmit

We should also make sure we're logging these errors and ideally set up some sort of threshold alert (attn @chigby @maeve-fpf)

Whatever we do here should also be implemented on FPF and SDO after, so please open tickets for those sites when this is resolved.

***

Mig noticed at one point during testing that he got a csrf error that went away when he tried again... caching doesn't seem to be a problem but there is a csrf cookie that might be the culprit. We've only seen this once, and no STR yet... Value of the secret is changed each time a user logs in. If they are anonymous, it's per session.

Timebox task? See if we can repro.

chigby:
## Description

Fixes #1674

1. Adds a CSRF error handler view that's under our control
2. Configures the view so that if (1) the URL being requested is a wagtail page, and (2) the page implements the method `redirect_for_csrf_error`, the view will redirect to whatever is returned by that method. Otherwise, a 403 response is returned, which is the current behavior.
3. Adds `redirect_for_csrf_error` to `FormPage` to store the submitted data in the Django session, then redirect to back to itself.
4. Modify form pages so that, when loading the form, it looks in the session for data from step 3 to populate the form. Also, if it finds such data, add a top level message saying "Submission failed, please try again." -- this can be adjusted, though I'm not sure how specific we want to be, like "CSRF validation failed, please clear cookies and try again" (or something).

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

Test by faking a CSRF error when submitting a form to a form page (like the submit an incident page, or a custom one you create). I found that you can reliably get the error by clearing the "csrftoken" cookie before submitting. The result of this should be that you're redirected back to the form page with the same data you originally submitted, except also with an error message at the top of the form.

### Post-deployment actions

None.

### General checks

image

### DIFF
--- a/client/common/sass/components/_forms.sass
+++ b/client/common/sass/components/_forms.sass
@@ -19,6 +19,10 @@
 
 	+constrain-to-desktop
 
+	&__top-level-error
+		font-style: italic
+		color: $error
+
 .form
 
 	+desktop-up
--- a/common/tests/test_views.py
+++ b/common/tests/test_views.py
@@ -4,7 +4,7 @@ from unittest import mock
 from django.core.files.base import ContentFile
 from django.urls import reverse
 from django.utils.http import urlencode
-from django.test import TestCase, override_settings
+from django.test import TestCase, override_settings, RequestFactory
 from wagtail.documents.models import Document
 from common.models import CommonTag
 from common.wagtail_hooks import CommonTagAdmin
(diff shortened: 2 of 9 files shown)


SaptakS (approved):
Looks good to me.

PR #1736, opened 2023-09-07
Define a default value for NPM version to be used for container

chigby:
## Description

Related to https://github.com/freedomofpress/fpf-www-projects/issues/359

Changes proposed in this pull request: We have run into an issue where some environments do not set the `NPM_VER` argument to the production django dockerfile, which can cause unexpected major-version changes in the `npm install npm@${NPM_VER}` command.  This prevents that from happening.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

With this change, building a container from the node dockerfile and the production django dockerfile should succeed even if `NPM_VER` is not given as an argument. This can be tested by, for example, removing that argument from the prod-docker-compose.yaml file (see https://github.com/freedomofpress/freedom.press/blob/npm-version-default-value/prod-docker-compose.yaml#L25) and building the container with 

```
docker compose -f prod-docker-compose.yaml build
```

Without this change, that command will fail because it tries to install npm version 10 on an incompatible engine.

### Post-deployment actions

None.

### General checks

No changes

No changes
No changes.

No changes.
No changes.

No changes.

### If it's a major change

Not a major change.

No frontend changes.

### DIFF
--- a/devops/docker/NodeDockerfile
+++ b/devops/docker/NodeDockerfile
@@ -4,7 +4,7 @@ FROM node:16-alpine@sha256:1908564153449b1c46b329e6ce2307e226bc566294f822f11c5a8
 # Make npm output less verbose
 ENV NPM_CONFIG_LOGLEVEL warn
 
-ARG NPM_VER
+ARG NPM_VER=9.6.7
 
 # Upgrade npm to speicifed version
 RUN npm install npm@${NPM_VER} --location=global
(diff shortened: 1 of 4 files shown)


dkaoster:
I noticed that the version specified in [`docker-compose.yaml`](https://github.com/freedomofpress/pressfreedomtracker.us/blob/develop/docker-compose.yaml#L38) is `9.6.0`. Should we try to keep these the same?

chigby:
Good point, I can update it

dkaoster (approved):

Issue #1742, opened 2023-09-18
Database page tag filter becomes non-functional


harrislapiroff:
Reported by @KAMcCudden:

> When one arrives at the Incident Database from a Category Main Page graphic link
> 
> ![image](https://github.com/freedomofpress/pressfreedomtracker.us/assets/296631/fe82295b-b710-4c7f-89a1-1d9caf4d4a4f)
> 
> That filtering is correct. But the option to filter with tags is lost at this step
> 
> ![image](https://github.com/freedomofpress/pressfreedomtracker.us/assets/296631/7489a236-e441-48e7-a915-e3a3cbb9e135)
> 
> However, other filters are available and work, and those filters re-engage the tag filtering option.
> 
> ![image](https://github.com/freedomofpress/pressfreedomtracker.us/assets/296631/1f3c1b57-1f70-40a9-8bc9-3b8960506850)

I was able to replicate this navigating to https://pressfreedomtracker.us/all-incidents/?categories=4. I suspect this is related to #1741

Issue #1741, opened 2023-09-18
Inconsistent behavior when filtering with category ID instead


harrislapiroff:
Our database page allows either category ID or name for filtering in the GET parameters, but behaves differently depending on which you use. With [category ID filtering](https://pressfreedomtracker.us/all-incidents/?categories=4) the sidebar filters break and don't reflect the proper category:

<img width="311" alt="image" src="https://github.com/freedomofpress/pressfreedomtracker.us/assets/296631/0aac1987-6c95-4e6c-b14b-f5b880e2237b">

and the "chips" at the top of the page show the number 4 instead of a category name:

<img width="233" alt="image" src="https://github.com/freedomofpress/pressfreedomtracker.us/assets/296631/03b18c50-5dbb-4f21-bcb7-56547a5649d4">

By contrast using the [category name filtering](https://pressfreedomtracker.us/all-incidents/?categories=Arrest%2FCriminal+Charge) renders everything properly:

<img width="797" alt="image" src="https://github.com/freedomofpress/pressfreedomtracker.us/assets/296631/88ca13b7-c1cc-4a76-96e2-afb4e344320e">

I think it's good that we support both options for filtering, but we should normalize the behavior here (especially since we're actively using the ID form in some places as links). Maybe one easy way to future-proof this would be to always provide an HTTP redirect from one form to the other so we only have to design the database page templates and javascript to handle one possibility?

chigby:
Just capturing this thought: I think it would be a nice improvement to the back-end code to do some kind of normalizing of strings to IDs (or the other way, if we prefer, but I think I like IDs more).

chigby:
Another thought: I think one of the reasons this is happening is that the data used by the database/filter page front end is in CSV format. It's worth remembering that CSV data made available by the API was originally designed to be part of an "export" function -- for people who wanted to download a subset of the Tracker's incident data in a spreadsheet-compatible format, and by design this is _human readable_. Numeric ID fields for categories, journalists, institutions, charge types, etc., were not included in that by design. So, there is inevitably going to be some amount of tension when also using that same data as a basis for various website features where no human is going to be seeing the data directly.

dkaoster:
@chigby can you comment on why you like IDs more? I don't have a strong preference but I feel like using the title results in slightly more human-readable urls, but it's not a huge deal to me.

chigby:
[...]

PR #1746, opened 2023-09-20
Convert category IDs given in query params to category titles

chigby:
## Description

Part of #1741 and #1742

Adds code to the sidebar filter component on the database page to handle numeric category IDs given in the `categories` key in the querystring. I have not updated the filter summary component (aka the yellow chips) at all.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

When loading the database page, passing a category page's numeric ID in the querystring should function identically to passing its title. For example: http://localhost:8000/all-incidents/?categories=13 should cause the filter sidebar to exhibit exactly the same behavior as http://localhost:8000/all-incidents/?categories=Chilling%20Statement (if the Chilling Statement category has ID 13)

### Post-deployment actions

None.

### General checks

No changes.

No changes.

No changes.

No changes.

No changes.

No changes.

### If it's a major change

No changes.

Here's the change (it's basically the sidebar but it has data displaying instead of not displaying):

image

### DIFF
--- a/client/charts/js/components/FilterSidebar.jsx
+++ b/client/charts/js/components/FilterSidebar.jsx
@@ -45,7 +45,7 @@ export default function FilterSidebar({ serializedFilters, initialDataset = null
 		get: (searchParams, prop) => searchParams.get(prop),
 	})
 
-	let initialFilterParams = decode(urlParams)
+	let initialFilterParams = decode(urlParams, filters)
 
 	return (
 		<FiltersIntegration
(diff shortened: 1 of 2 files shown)


harrislapiroff:
Haven't tested this locally, but the code looks good. If this fixes the issue then I'm happy for us to merge it ASAP to fix the bug.

chigby:
@harrislapiroff Setting this to "Ready for review" though it looks like you've already done some of that. IMO the filter summary situation is better served in another issue and PR.

dkaoster (approved):
Seems to work well on my end!

PR #1749, opened 2023-09-22
Process numeric categories in FilterSummary.jsx

### LINKED ISSUE #1741: Inconsistent behavior when filtering with category ID instead

### LINKED ISSUE #1742: Database page tag filter becomes non-functional

dkaoster:
## Description

Fixes #1742 
Fixes #1741 

Similar to #1746, this applies the same logic to the chips used for the filter summary to display the category titles. This similarly reads from the serialized filters to get the ids matched to titles.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

Run the application locally, and use [any one](http://localhost:8000/equipment-search-seizure-or-damage/?) of the category pages to follow the [link to the database page](http://localhost:8000/all-incidents/?categories=7). The page should have the proper filter chips at the top with the category title, and the sidebar should be functional.

### General checks

Now, if you go to http://localhost:8000/all-incidents/?categories=Leak+Case%2CChilling+Statement,4

you should see:

Screenshot 2023-09-22 at 4 51 32 PM

### DIFF
--- a/client/charts/js/components/FilterSummary.jsx
+++ b/client/charts/js/components/FilterSummary.jsx
@@ -15,7 +15,11 @@ const symbolMap = {
 	"Other Incident": "other_incident",
 }
 
-export default function FilterSidebar() {
+export default function FilterSummary({ serializedFilters }) {
+	const categoryFilters = JSON.parse(serializedFilters)
+
+	const idMap = categoryFilters.reduce((acc, val) => ({ ...acc, [val.id]: val.title }), {});
+
 	const searchParams = new URLSearchParams(window.location.search)
 	const {
 		categories: categoriesStr,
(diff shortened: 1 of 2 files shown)


chigby (approved):
This works great! :sparkles: 

I'm going to rebase this branch, then merge.

PR #1753, opened 2023-09-28
Reorganize node chart-generator Dockerfile, add dev mode to script

maeve-fpf:
## Description

In production, we run containers with readonly root. This means that we do not want to run `npm install` at runtime. However, we still want to do that in development. So, we need something like we do with the Django container, where it can be optionally run in "dev mode" that allows restarting and incrementally installing dependencies. This commit adds that to the script, and changes the Dockerfile somewhat to make it work.

## Type of change

- [ ] Bug fix (sort of... I think this points to a need for infra to better explain what production is)

## Testing

I will deploy this to the experimental cluster for testing. Devs, please test it in development mode (you may want to wipe out your local container and rebuild).

### Post-deployment actions

If this looks good in dev and we merge it into `chart-pregenerator`, I can enable builds for pushes to that branch so that they will automatically deploy, since the container will not crash on startup. Further configuration will be required on that point, but let's get this in first.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,7 @@ jobs:
     environment:
       SETUPTOOLS_USE_DISTUTILS: stdlib
     machine:
-      image: ubuntu-2204:2023.04.2
+      image: ubuntu-2204:2023.07.2
     working_directory: ~/tracker
     steps:
       - checkout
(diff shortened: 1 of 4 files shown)


dkaoster (approved):
works locally for me! 👍

chigby:
Potentially useful [stack overflow answer](https://stackoverflow.com/a/76487443/1382355): is esbuild not compatible with something deeper in the docker setup here? I'll investigate more tomorrow.

chigby:
So this is now passing after I upgraded node/ubuntu on this branch, then brought the base branch up to date with `develop` and rebased this one on it. I think it should be ready to merge but @maeve-fpf if you want to look it over before pushing the button, please do.

PR #1756, opened 2023-09-28
Docker compose health checks

chigby:
## Description

Adds docker health checks for development containers.

This allows us to remove:

1. The use of the `.node_complete` file as a way of signaling to other containers.
2. Potential race conditions between webpack compilations and Django's attempting to look for the `webpack-stats.json` file
3. The explicit use of `ping` in CI jobs to wait for the Django container to be ready.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

This PR should "just work" in that `docker compose up` and its variants all still bring the docker system up, and without the need for intermediate files or anything else mentioned in the description.

### Post-deployment actions

None

### General checks

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -123,8 +123,7 @@ jobs:
           name: Ensure we can run dev-env
           command: |
             make dev-init
-            docker compose up --detach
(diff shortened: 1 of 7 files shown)


SaptakS (approved):
Looks good to me

PR #1757, opened 2023-09-29
Stacked Bar by Tag

### LINKED ISSUE #1740: Stack bar chart by tag/no tag
In addition to the current stacking options, the tracker eds would like to create stacked bar charts where the bars are segmented into "has tag" and "doesn't have tag" for a specific tag.

For example, they might want to make a chart that showed equipment search and seizure over time showing that most of these occurred at protests. They should be able to limit a bar chart to equipment search and seizure (they can already do this) and then choose a segmenting tag "protests" (this is the new part) to render a chart like this:

Artboard

(Forgive the quick mockup—neither pixel accurate nor data accurate.)

I'm leaving unspecified what the UX should look like but if you want to talk it out, put a meeting on my calendar. I think choosing a tag for segmenting needs to be mutually exclusive with the other segmenting options, so we may need to do some clever validation to enforce that. It's also possible we need to allow them custom label text since I can't think of a consistent way to label tags automatically. (I guess "protest/not protest" kinda works, but it's not as clear as "at protest" and "not at protest", but that pattern won't work for all tags either.)

dkaoster:
## Description
Adds the option for vertical bar charts to be grouped by a tag.

Fixes #1740 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

Run this branch locally, and select a tag as the group by field. Make sure that the resulting bar chart is stacked by tag.

### General checks

In wagtail, vertical bar charts now have tags included in the group by options:
Screenshot 2023-09-29 at 8 57 11 AM

Screenshot 2023-09-29 at 8 57 05 AM

Doing so generates this bar chart:
Screenshot 2023-09-29 at 8 56 30 AM


incident/utils/charts.py
```diff
@@ -75,3 +75,10 @@ class IncidentBranches(TextChoices):
     CATEGORIES = 'CATEGORIES', 'Categories'
     ASSAILANT = 'ASSAILANT', 'Assailant'
     STATUS_OF_CHARGES = 'STATUS_OF_CHARGES', 'Status of Charges'
+
+
+def IncidentBranchesAndTags():
```
dkaoster: This is probably not the best way to do it, but it was how I was able to get it working initially. Open to suggestions!

common/blocks.py
```diff
@@ -503,7 +503,7 @@ class VerticalBarChart(blocks.StructBlock):
     group_by = blocks.ChoiceBlock(
         label='Group Incidents By',
         required=False,
-        choices=charts.IncidentBranches.choices,
+        choices=charts.IncidentBranchesAndTags,
```
chigby: I was going to say that I'd kind of prefer the tags be separated into their own field on `VerticalBarChart` instead of being combined here because (1) the editor UX would be nicer to be able to use the autocomplete widget to choose the tag, and (2) because then we could avoid inferring what the user has chosen when we're rendering the chart, _but_ then I remembered that actually we can't use the autocomplete widget here since we're in a `StructBlock`! The other place where we're letting the user choose tags, in the incident set, we have a drop down. So now I'm kind of not feeling as strongly about separating this into another field. @harrislapiroff any thoughts?

harrislapiroff: I think we should separate it out into a separate field in anticipation of using an autocomplete widget in the future (How hard would it be to use an autocomplete widget for a block? That feels like it should definitely be a feature.)

dkaoster: In this case, is there a way to limit users to only choosing either a category or a tag? I don't know what the behavior should be if they happened to select group by a category and a tag?

chigby: Sadly, I don't think wagtail 4.1 supports this kind of strict validation. I'm told it's something that's out there in a future wagtail version (maybe 5+? [Here's some documentation](https://docs.wagtail.org/en/v5.1.2/advanced_topics/streamfield_validation.html)) which we are probably going to upgrade to Soon(tm).

Given this, I'd be fine with adding some help text explaining that these options are mutually exclusive. While at a system level we can't really stop them from choosing both (yet), I think for now it's fine to have one mode "win" if someone picks both a tag and a category. @harrislapiroff thoughts on this?

harrislapiroff: @chigby I agree that's a good approach. We already have places where we do this—for example button and nav fields that accept either a page, document, or URL. Maybe we can open a ticket for adding better validation pending the next version of Wagtail. (Maybe a new label "blocked on:Wagtail 5" or something?)

dkaoster: sounds good, let me see what I can do

dkaoster:
I have updated it to use a separate field, as follows:
Screenshot 2023-10-10 at 6 33 20 PM

chigby (approved):
I think this is looking really good, and thanks for adding the extra field. I'm sorry if that felt a little like it came from out of nowhere, it emerged out of some discussions in meetings while you were out of the "office." But what you've made here is excellent and I think (a) it works great and (b) lays the groundwork for those validation improvements that are (I hope) coming soon to wagtail. 

Anyway, I tested this out locally and everything seems to be in order so I'm going to merge it.

PR #1760, opened 2023-10-03
Update selenium and urllib3

chigby:
## Description

Selenium update is needed to support the latest urllib3.

Urllib3 updates fixes this vulnerability:

* https://github.com/advisories/GHSA-v845-jxx5-vc9f

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Vulnerabilities update
- [ ] Config changes


harrislapiroff (approved):
LGTM. Interesting that upgrading urllib seems to have removed our dependency on `pyopenssl` and `cryptography` 👀

PR #1773, opened 2023-10-18
Adds robots.txt and sitemap.xml

### LINKED ISSUE #1761: Create robots.txt
We're seeing a lot of crawler traffic, which can slow down other site operations. Let's create a carefully crafted robots.txt that defines what 's OK to crawl and what's not.

SaptakS:
## Description

Fixes #1761.

Adds a robots.txt which disallows all bots access to `/all-incidents/?*` but allows `/all-incidents/?page=*`. Also adds a sitemap.xml and adds the sitemap link to robots.txt. A sitemap lets the crawlers know which pages to crawl (which helps search engines know about different pages) and also addresses [WCAG SC 2.4.5](https://www.w3.org/WAI/WCAG22/Understanding/multiple-ways)

PS: Not all crawlers follow robots.txt fully, so this PR might not fully solve the issue of being crawled.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

- Check `/robots.txt` shows the content  that it should.
- Check `/sitemap.xml` shows the sitemap correctly.

### DIFF
--- a/common/tests/test_template_tags.py
+++ b/common/tests/test_template_tags.py
@@ -9,6 +9,7 @@ from common.templatetags.common_tags import (
     add_as_string,
     richtext_aside,
     query_transform,
+    get_absolute_url,
 )
 
 
@@ -108,3 +109,11 @@ class TestTemplateTags(TestCase):
         qd = QueryDict(result)
         self.assertEqual(qd['page'], '1')
         self.assertEqual(qd['sort'], 'title')
+
+    def test_get_absolute_url(self):
+        request = RequestFactory().get('/')
+        context = {'request': request}
+
+        result = get_absolute_url(context, 'sitemap')
+
+        self.assertEqual(result, 'http://testserver/sitemap.xml')
--- a/common/tests/test_template_tags.py
+++ b/common/tests/test_template_tags.py
@@ -116,4 +116,4 @@ class TestTemplateTags(TestCase):
 
         result = get_absolute_url(context, 'sitemap')
 
-        self.assertEqual(result, 'http://testserver/sitemap.xml')
+        self.assertEqual(result, f'http://{request.get_host()}/sitemap.xml')
--- a/tracker/templates/robots.txt
+++ b/tracker/templates/robots.txt
@@ -1,3 +1,5 @@
 User-agent: *
 Disallow: /all-incidents/?*
 Allow: /all-incidents/?page=*
+
+Sitemap: https://pressfreedomtracker.us/sitemap.xml
--- a/tracker/templates/robots.txt
+++ b/tracker/templates/robots.txt
@@ -1,0 +1,3 @@
+User-agent: *
+Disallow: /all-incidents/?*
+Allow: /all-incidents/?page=*
--- a/tracker/templates/robots.txt
+++ b/tracker/templates/robots.txt
@@ -1,6 +1,7 @@
+{% load common_tags %}
 User-agent: *
 Disallow: /all-incidents/?*
 Allow: /all-incidents/?page=*
 Crawl-delay: 10
 
-Sitemap: https://pressfreedomtracker.us/sitemap.xml
+Sitemap: {% get_absolute_url 'sitemap' %}
--- a/tracker/templates/robots.txt
+++ b/tracker/templates/robots.txt
@@ -1,5 +1,6 @@
 User-agent: *
 Disallow: /all-incidents/?*
 Allow: /all-incidents/?page=*
+Crawl-delay: 10
 
 Sitemap: https://pressfreedomtracker.us/sitemap.xml
--- a/tracker/urls.py
+++ b/tracker/urls.py
@@ -4,6 +4,7 @@ from django.apps import apps
 from django.conf import settings
 from django.urls import include, path, re_path
 from django.contrib import admin
+from django.views.generic import TemplateView
 from drf_spectacular.views import SpectacularAPIView, SpectacularRedocView, SpectacularSwaggerView
 from wagtailautocomplete.urls.admin import urlpatterns as autocomplete_admin_urls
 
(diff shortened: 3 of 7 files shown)


tracker/urls.py
```diff
@@ -29,6 +31,8 @@
     path('health/ok/', common_views.health_ok),
     path('health/version/', common_views.health_version),
     path('csrf/', common_views.get_csrf_token, name='csrf_token'),
+    path('sitemap.xml', sitemap),
```
chigby: I wonder if we need to give this view some special treatment with regards to the cloudflare cache.

tracker/templates/robots.txt
```diff
@@ -0,0 +1,5 @@
+User-agent: *
+Disallow: /all-incidents/?*
+Allow: /all-incidents/?page=*
+
+Sitemap: https://pressfreedomtracker.us/sitemap.xml
```
mig5: Could I make a recommendation to add a [`Crawl-Delay`](https://en.wikipedia.org/wiki/Robots.txt#Crawl-delay_directive) directive, just to aid the performance matter? 

It is strictly-speaking, a non-standard directive, however, some well-intentioned bots respect it.

chigby: I think there is probably a way to avoid hard-coding the domain, the sitemap name, or both here. This may or may not appeal to you, but if you give the URL a name, such as 

```
path('sitemap.xml', sitemap, name='sitemap'),
```
Then you can write:

```
Sitemap: https://pressfreedomtracker.us/{% url 'sitemap' %}
```
So that it will refer to whichever path we have defined in our URLs file.
* [`url` docs](https://docs.djangoproject.com/en/4.2/ref/request-response/#django.http.HttpRequest.build_absolute_uri)

To get the absolute URL is a little more trouble, which I was surprised by, but it would involve writing a custom template tag, so that we can write

```
Sitemap: {% absolute_url 'sitemap' %}
```

We'd have to add, for example, to `common_tags.py`, something like this:

```py
@register.simple_tag(takes_context=True)
def absolute_url(context, view_name, *args, **kwargs):
    return context['request'].build_absolute_uri(
        reverse(view_name, args=args, kwargs=kwargs)
    )
```

And then we'd have to add `{% load common_tags %}` to the robots.txt template. 
* [`build_absolute_uri` docs](https://docs.djangoproject.com/en/4.2/ref/request-response/#django.http.HttpRequest.build_absolute_uri)
* [`reverse` docs](https://docs.djangoproject.com/en/3.2/ref/urlresolvers/#reverse)

This is not particularly a lot of additional code in either case, but I'll let you be the judge of how enticing either of these additions sounds.

tracker/urls.py
SaptakS: You mean cache bursting when a new page is added? That's a good question. I am not sure how we would do that without messing with the sitemap internals of wagtail.

tracker/templates/robots.txt
SaptakS: I am aware of the name parameter of `path`, I was just like I don't see any scenario where we would want to change the hostname or the sitemap path name (since both of them are standards). But I guess for future proofing, we can remove hardcoding.

tracker/urls.py
chigby: Oh totally, I don't think we should try to mess around with the internals, but it might be something we'd want to apply headers to, like we are doing here:

https://github.com/freedomofpress/freedom.press/blob/02989d22d5c5ab7ad681bcb64e91970ca8a42f1d/bundles/models/bundles.py#L523-L524

or here

https://github.com/freedomofpress/freedom.press/blob/develop/common/views.py#L56-L57

These decorators can be applied directly to the view in the URL line, something like:

```py
path('sitemap.xml', cache_control(private=True)(sitemap), name='sitemap'),
```

But I also am not sure how strongly we feel about caching this sitemap overall.

SaptakS: Yes. Honestly I was thinking that this page should ideally be cached (I am not fully sure how adverse it is performance wise). Else the bots checking sitemap link shouldn't bring down our website. But I know cache bursting will be an issue as well. So I am not sure either. @harrislapiroff thoughts?

harrislapiroff: Definitely should be cached. Can we register a signal to bust the cache for this path whenever a page is published? Otherwise we could just set the TTL shorter for this endpoint.

SaptakS:
@chigby I have added cache purging for sitemap, and also added tests for `get_absolute_url` templatetags. Should be ready for re-review.

harrislapiroff (approved):
The changes made in response to comments here look great to me! Thanks @SaptakS. @chigby do you want to take a quick additional pass?

chigby:
One comment I have that's not extremely related to this PR but just generally how we're using the `post_delete` signal instead of the wagtail specific `page_unpublished`. I'm not sure how often pages are actually deleted, so like if a page is unpublished without being deleted, then the cache won't be purged.  I'm not sure we need to address that here, since a bunch of other signal handlers are also set up to be using `post_delete` and maybe this change would be better incorporated into a larger reworking of this code which could include, for example, purging pages in batches.

harrislapiroff:
@chigby That's a really good point @chigby. Do you want to open a ticket to rework our signals code more generally? It has been a source of some issues in the past (even now I think the homepage of FPF doesn't always get purged consistently when a new blog post is published and I don't know why—possibly cleaning up this code would lead us to some solutions)

common/tests/test_template_tags.py
```diff
@@ -108,3 +109,11 @@ def test_query_transform_ignores_unknown_keys_set_to_none(self):
...
         self.assertEqual(qd['page'], '1')
         self.assertEqual(qd['sort'], 'title')
+
+    def test_get_absolute_url(self):
+        request = RequestFactory().get('/')
+        context = {'request': request}
+
+        result = get_absolute_url(context, 'sitemap')
+
+        self.assertEqual(result, 'http://testserver/sitemap.xml')
```
chigby: This line is fine, I will just point out that we can get `testserver` from the request object itself with `request.get_host()`

chigby:
IMO this can be merged once it's up-to-date with the base.

PR #1778, opened 2023-10-20
Smart search phase one

### LINKED ISSUE #1767: Smart Search Phase 1
Okay, let's actually try to get phase one of the smart search in! Here's the spec

1  Basic search

- [ ] When in focus the search bar pops up a dropdown showing the same list of tags that are used for the homepage data viz
- [ ] A user can type text into the search bar, causing the dropdown to disappear
- [ ] A user can click a tag from the dropdown causing the tag to appear in the search bar with a # next to it and the dropdown to disappear
  - [ ] This isn't in the above spec but let's also render a × next to the tag that will clear the field when clicked, since there's no other way to clear it right now (the "next steps" in the spec seem to apply to a later iteration)
- [ ] A user can also select a tag using the arrow keys and enter button
- [ ] Submitting the search form behaves differently depending on its contents:
  - [ ] If a tag is contained within, it goes to the database page with a tag filter applied
  - [ ] If free text is contained within, it goes to the database page with a text search

In developing this be attentive to making the code flexible, clean and ready to adapt to new features. You can look at the figma to see what the next phases look like: https://www.figma.com/file/H3uTXxT88ofhdebUFgzEo9/PFT-%E2%80%93-Hand-off-Guide?type=design&node-id=4495-110643&mode=design&t=7XimrARBwTXy6QZb-4

dkaoster:
## Description

Fixes #1767

Adds the smart search phase one by showing a dropdown of tags when the search bar is active.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing
Run the app locally, and in the homepage or the incident page, use the search bar by clicking it and expanding the options. Ensure that the search bar works and satisfies accessibility requirements.

### General checks

Screenshot 2023-10-20 at 2 26 18 PM

### DIFF
--- a/client/common/js/components/Search.jsx
+++ b/client/common/js/components/Search.jsx
@@ -1,4 +1,4 @@
-/* eslint-disable jsx-a11y/label-has-associated-control */
+/* eslint-disable jsx-a11y/label-has-associated-control, no-case-declarations */
 import React, { useState } from 'react'
 import classNames from 'classnames'
 import PropTypes from 'prop-types'
@@ -24,6 +24,28 @@ export default function Search({ data = [] }) {
 		setSearchText(text)
 	}
 
+	const handleArrowKeys = (event) => {
+		const currentId = event.target.id
+		switch (event.keyCode) {
+		case 38: // up
+			// select prev tag
+			const allTags = [...document.querySelectorAll('.search-dropdown--tag')].reverse()
+			const currentTagIndex = allTags.findIndex((tagEl) => tagEl.id === currentId)
+			if (allTags[currentTagIndex + 1]) allTags[currentTagIndex + 1].focus()
+			event.preventDefault()
+			break
+		case 40: // down
+			// select next tag
+			const nextEl = document.querySelector(
+				`#${currentId} ~ .search-dropdown--tag, #${currentId} ~ .search-dropdown .search-dropdown--tag`,
+			)
+			if (nextEl) nextEl.focus()
+			event.preventDefault()
+			break
+		default:
+		}
+	}
+
 	const handleSubmit = (event) => {
 		event.preventDefault()
 		if (searchText) {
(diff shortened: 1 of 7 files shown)


dkaoster:
@SaptakS I think I have a good start on accessibility here, would be good to get your thoughts!

SaptakS:
I will take a look.

SaptakS (changes requested):
Looks mostly great to me! There are few accessibility issues that I noticed.
- The smart search doesn't work with keyboard when selecting one of the options using keyboard. The bhaviour I expect is: I select an option -> I press enter -> the search with the option happens. Right now there are 2 issues - selecting the option adds a tag which is after the search button, so the tab order gets messed up; and the focus leaves the form hence pressing enter doesn't cause a search to happen. The solution I think is to move the search-tag-pill before the search button, and once an option is selected, focus the search button
- I think the semantic for the options can be improved a little bit to make it a list of options (You can check the semantics here for some ideas: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/). Also, I think the list of options don't need to be tabbable and arrow keys are okay, especially since the tabs for the options come after the search button
- The close button for the tags can be made more accessible. This is what you are looking for: https://www.htmhell.dev/20-close-buttons/#solution-5-font-awesome

client/common/js/components/Search.jsx
```diff
@@ -97,28 +119,14 @@ export default function Search({ data = [] }) {
 							className="search-dropdown--tag"
 							onClick={() => updateSelectedTag(tag)}
 							onKeyDown={handleArrowKeys}
-							tabIndex="0"
+							tabIndex="-1"
```
SaptakS: Instead of adding a tabIndex=-1, can we just use a different HTML structure? I am thinking `<ul><li role="option">`

dkaoster: Unless I'm missing something, doing this still makes the buttons tabbable?

SaptakS: After discussing with @dkaoster I think it's fine to keep it as is. I was thinking of removing the button element and just using li, but he pointed out that might be more fiddly than needed.

SaptakS (approved):
Looks good to me

PR #1779, opened 2023-10-23
Fix vertical bar charts when incident has no tags

dkaoster:
## Description

Fixes https://freedomofpress.slack.com/archives/C05DGNV886R/p1698088793401849 for when there are incidents with no tags, `tags = null` which causes problems with the group by tag logic.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

- To test, check out this branch and edit an incident by removing all of the tags.
- Go to a blog page and create a vertical bar chart that is grouped by tag.
- Ensure that the vertical bar chart still shows up.


chigby (approved):
This solved the problem for me locally. Let's test on staging tomorrow.

PR #1782, opened 2023-10-26
Updates name for actor field in equipment damage

SaptakS:
## Description

Request from @stephaniesugars on slack:
> the "actor" field on equipment damage currently displays as "actor who seized equipment" (which is why we stopped using it for the ED category in the first place). Can you change it to just read "Actor"?

This PR addresses the above request.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

- Update an incident to have `equipment-damage` category
- Add an actor field
- Check that the metadata table shows "Actor" instead of "Actor who seized equipment"

### Post-deployment actions

### General checks

### DIFF
--- a/incident/models/category_fields.py
+++ b/incident/models/category_fields.py
@@ -49,7 +49,7 @@ CATEGORY_FIELD_MAP = {
     ],
     'equipment-damage': [
         ('equipment_broken', 'Equipment Broken'),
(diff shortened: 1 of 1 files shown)


chigby (approved):
Looks good!

PR #1792, opened 2023-11-08
Add `mistakenly_released_materials` under Prior Restraint category

### LINKED ISSUE #1478: USPFT Metadata changes
In the vein of #1018, the Tracker team has reviewed the other categories as well and is requesting these updates ([detailed in full here with draft mock-ups](https://docs.google.com/document/d/1HPEAO641opu034WelRSkzaWC3lN4CIDz3z6m41R7z5o/edit#))

General asks:
- [x] Alphabetize the metadata categories (Legal case can/should be at bottom just before “related cases”)
- [x] On database filters page, display text for date filters is messed up: “took place between after” and “took place between before” (Switch to “took place after” and “took place before”)
- [x] Remove parenthetical “incl. legal case” from “leak prosecution” and “prior restraint”

Assault metadata
- [x] rename entry line from "physical assault" to just "assault"

Border Stop metadata
Remove: (will result in deleting this data from the database)
- [x] Stopped at border?
- [x] Did authorities ask for social media username?
- [x] Did authorities ask for social media password?
- [x] Were devices searched or seized?

~~Chilling Statement metadata (to be newly created)~~
~~- [ ] Add 'actor': (politician / public figure) [no recording of name]~~
~~- [ ] Add 'source of statement': (investigation / legal action / legislation / public statement / social media)~~
_Removed 2023-10-31 by @chigby for not being part of relevant Google doc_

Denial of Access metadata
- [x] Add “type of denial” (select all that apply): (changes in policy or practice / government events / other / press credential or media list) -- In progress https://github.com/freedomofpress/pressfreedomtracker.us/pull/1781
- [x] Rephrase existing language from 'Politicians or public officials involved' to 'Government agency or public official involved'

Equipment Damage metadata
- [x] Rename “Equipment Broken” to “Equipment Damaged”
- [x] Add actor (with the same general options that we use for assaults) above equipment entry (this should mirror equipment search/seizure) 
- [x] Rename button from “add equipment broken” to just “add equipment” with matching display text

Equipment Search or Seizure metadata
- [x] Merge “Equipment Seizure or Damage” into the existing “Equipment Searched or Seized" menu
- [x] Rename button from “add equipment search or seized” to just “add equipment” 
- [x] Remove 'Actor" as only actor viable for counting under the category is law enforcement

Prior Restraint metadata
~~- [ ] Add date entry fields for date ordered and date lifted~~
~~- If possible, we would love for this to essentially be used to field a "length of restraint" display metadata / potentially make "length of restraint" searchable. Date itself does not need to be added to filter.~~ Moved to https://github.com/freedomofpress/pressfreedomtracker.us/issues/1784
- [ ] Add "Mistakenly released materials?" checkbox

Subpoena/Legal Order metadata
_For "subpoena of journalism" entry field_
- [x] Rename as “Subpoena/legal order targeting journalist”
- [x] Akin to the arrest category changes: will have a "add legal order" button
- [x] From there, a “Select legal order type” [select one]: (motion to compel / subpoena / warrant / other)
- [x] Then, "Select information requested" : “communications or work product / testimony about confidential source / other testimony”
- [x] Status selection w/ date and button for adding additional statuses 

_For “Legal order for journalist’s records” entry field_
- [x] Rename as "Subpoena/legal order to third party"

chigby:
## Description

Resolves #1478 

Add field `mistakenly_released_materials` under the Prior Restraint category. The field is a straightforward true/false boolean (as opposed to a "maybe boolean" we have in a few places).

### Outstanding question

I noticed that this boolean filter doesn't really work with the filter summary as of right now. The box just reads "1" since that's what is in the URL query string. But I'm not sure what the filter summary should be? I tried this text, and it doesn't really look right either, and has kind of a "key: value" structure instead of just a "value" structure that the rest of the summary buttons have.

image

Should it just say "Mistakenly released materials"? What's the negative case then, "No mistakenly released materials"? Not sure what to do here.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes
- [x] This change requires an admin update after deploy

## Testing

Make sure that the field operates as we expect in these contexts:

1. Admin. Is it in the right place in the editor UI? Is the default value correct? Does updating and previewing the field work?
2. The front end. Is this field visible in the incident cards? On the incident detail page? Does it display how we want it to? 
3. Filtering. Does the field display in the right place on the filter sidebar? Is the UI there correct? When applied, does the filter work correctly for filtering down the list of incidents? (**Note**: this involves adding it as a filterable field, as described in the next section).
4. API. Does the field render correctly in the JSON API? How about the CSV API?

### Post-deployment actions

Yes, if we want the field to be filterable, it needs to be added to the Prior Restraint category page in the appropriate spot:

image

### General checks

- [x] Verify that API responses are correct
- [x] Verify that visualizations using the API endpoints are functional

- [x] Verify incident export works correctly
- [x] Verify incident filters are rendered correctly
- [x] Verify incident filters show correct incidents
- [x] Verify categories work
- [x] Verify incidents are discoverable by search

No changes.

~~- [ ] Verify that it renders correctly in homepage, if applicable~~ (not applicable)
~~- [ ] Verify that it renders correctly in blog index page, if applicable~~ (not applicable)
~~- [ ] Verify that it renders correctly in individual blog page, if applicable~~ (not applicable)

No changes.

No changes

### If it's a major change

#### Incident index page

image

#### Incident detail page 

image

#### Admin 

image


dkaoster:
@chigby In response to your question about the boolean, I will note that the filters for date range do show up as `from: 2023-11-08` and `to: 2023-11-08` so I'd say that the `key: value` pattern isn't totally foreign. I think I'd be fine with implementing it `Mistakenly released materials: true`.

harrislapiroff:
I think ideally the yellow token would say "Mistakenly released materials" (no question mark, no need for key-value, just present or not). Do we need to add a bit of additional configuration to filter fields to generate summary strings like that?

chigby:
@harrislapiroff I think the issue with that is you are also allowed to apply the negative version of the filter, you can say `?mistakenly_released_materials=0`

chigby:
I guess that one could say "No Mistakenly Released Materials" -- maybe?

harrislapiroff:
Oooh it's one of those three-state booleans. Yeah I think you're correct that my preferred solution would be to have "Mistakenly released materials" and "No mistakenly released materials." If that's technically challenging, I'd accept the key-value structure as an alternative.

chigby:
I'm kind of wondering if we maybe do want some kind of config here. Like, here are the fields we want to apply this to. Verbose name, question mark removed:

* Unnecessary use of force
* Search warrant obtained
* Denied entry
* Stopped previously
* Charged under espionage act
* Mistakenly released materials

Some of these want "no" and some seem to want "not"

harrislapiroff:
I agree we want slightly different verbiage for each of those. Maybe we should treat this somewhat like `verbose_name_plural` where there's a default behavior that can be overriden. Default should probably be something like `Denied entry? No` but we probably want to override it with `Not denied entry`.

chigby:
Added the ability to override these fields, and now the filter summaries for boolean fields looks like this:

image

All are set to negative above, except "Unnecessary use of force."

SaptakS (approved):
Looks good to me.

PR #1803, opened 2023-11-29
Adjust frequent tags

### LINKED ISSUE #1801: Smart Search Phase One (Part II)
I think there was a bit of missed communication on my part in the implementation of smart search phase one. @dkaoster rigged it up so that it always displays the most frequently used tags, but we actually want it to display the tags selected in the admin if there are any.

image

dkaoster:
## Description

Fixes #1801

Sets the homepage to the tags specified in the admin, and the search page to be the tags that are most frequently used.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

Run the app locally, and double check http://localhost:8000/ and http://localhost:8000/all-incidents/ to make sure that the suggested tags in the search bars are correct.

### General checks

- [x] Verify that it renders correctly in blog index page, if applicable
- [x] Verify that it renders correctly in individual blog page, if applicable
- [x] Verify that it renders correctly in individual special blog page, if applicable

Screenshot 2023-11-28 at 7 15 14 PM

Screenshot 2023-11-28 at 7 15 03 PM

### DIFF
--- a/client/charts/js/components/HomepageSelection.jsx
+++ b/client/charts/js/components/HomepageSelection.jsx
@@ -9,8 +9,13 @@ import {
 	filterDatasetByFiltersApplied,
 } from '../lib/utilities.js'
 
-export function chooseMostFrequentTags(dataset, numberOfTags) {
-	const tags = dataset.map((d) => d.tags).filter((d) => d !== null)
+export function chooseTrendingTags(dataset, numberOfTags) {
+	const currentDate = new Date();
+	const filterStartDate = (new Date()).setFullYear(currentDate.getFullYear() - 1);
+	const tags = dataset
+		.filter(({ date }) => date >= filterStartDate)
+		.map((d) => d.tags)
+		.filter((d) => d !== null)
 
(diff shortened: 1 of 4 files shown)


harrislapiroff (approved):
This looks great, @dkaoster, thank you! The only thing I'd add is the thing I suggested [on the ticket](https://github.com/freedomofpress/pressfreedomtracker.us/issues/1801#issuecomment-1830346950), to change the default behavior/behavior on the search page by changing the `chooseMostFrequentTags` function to a `chooseTrendingTags` function—which does the same logic, but limits itself to incidents from the past year.

The only reason I suggest this is that 2020 was such an exceptional year for PFT that as long as we're doing all time most popular tags they're probably going to be the same 5 tags from 2020. But my priority is getting this merged and deployed this week, so only make this change on this branch if you can get it in today. If it turns out to be complicated, we'll save it for a future PR.

harrislapiroff (approved):
Looks great! :shipit:

PR #1808, opened 2023-12-05
Migrate to new Docker Compose syntax in README

eloquence:
## Description
From https://docs.docker.com/compose/install/

> Effective July 2023, Compose V1 stopped receiving updates and is no longer in new Docker Desktop releases. Compose V2 has replaced it and is now integrated into all current Docker Desktop versions. For more information, see [Migrate to Compose V2](https://docs.docker.com/compose/migrate).

From that page:

> Update scripts to use Compose V2 by replacing the hyphen (-) with a space, using docker compose instead of docker-compose.

Our `Makefile` targets already were updated in c4f89bcd9fe52f950b43e3bdadd9d736ced33425 but the README still used the old syntax.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,13 +23,13 @@ Prerequisites
 The installation instructions below assume you have the following software on your machine:
 
 * `docker <https://docs.docker.com/engine/installation/>`_
-* `docker-compose <https://docs.docker.com/compose/install/>`_
+* `docker compose <https://docs.docker.com/compose/install/>`_
 
 Local Development instructions
 ------------------------------
 
 When you want to play with the environment, you will be using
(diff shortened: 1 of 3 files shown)


chigby (changes requested):
We might as well update the `TROUBLESHOOTING.md` document while we're at it.

eloquence:
Updated the remaining mentions -- not sure the `savedb.sh` script is still used, but if it is it'll surely need the new syntax as well.

chigby (approved):

chigby:
I think that the save/restore db scripts are no longer used, having been made mostly obsolete by docker.

PR #1810, opened 2023-12-07
Smart search phase two

### LINKED ISSUE #1804: Smart Search Phase Two
Here's the spec from Accurat for Smart Search Phase Two—feel free to ask me any questions about the intention of any parts of the design or any gaps we need to fill in that aren't accounted for in this spec:

2  One filter option

dkaoster:
## Description

Fixes #1804

Adds smart search phase two by editing the dropdown to include categories when the search bar is active.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing
Run the app locally, and in the homepage or the incident page, use the search bar by clicking it and expanding the options. Ensure that the search bar works as described in the mocks and satisfies accessibility requirements.

### General checks

Screenshot 2023-12-06 at 6 18 27 PM

### DIFF
--- a/client/charts/js/components/FilterSummary.jsx
+++ b/client/charts/js/components/FilterSummary.jsx
@@ -1,19 +1,5 @@
 import React from 'react'
-import classNames from 'classnames'
-
-const symbolMap = {
-	"Arrest / Criminal Charge": "arrest",
-	"Border Stop": "border_stop",
-	"Denial of Access": "denial_of_access",
-	"Equipment Search or Seizure": "equipment_search",
-	"Assault": "assault",
-	"Leak Case": "leak_case",
-	"Subpoena / Legal Order": "subpoena",
-	"Equipment Damage": "equipment_damage",
-	"Prior Restraint": "prior_restraint",
-	"Chilling Statement": "chilling_statement",
-	"Other Incident": "other_incident",
-}
+import CategoryIcon, { categorySymbolMap } from '../../../common/js/components/categoryIcon'
 
 export default function FilterSummary({ serializedFilters }) {
 	const categoryFilters = JSON.parse(serializedFilters)
@@ -44,7 +30,7 @@ export default function FilterSummary({ serializedFilters }) {
 	const categories = [...new Set(
 		(categoriesStr ? categoriesStr.split(',').map(d => d.trim()) : [])
 			.map(category => {
-				if (symbolMap[category]) return category;
+				if (categorySymbolMap[category]) return category;
 				if (idMap[category]) return idMap[category];
 				return null;
 			})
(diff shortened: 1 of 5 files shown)


harrislapiroff:
We discussed some UX concerns in web sync today, some of which predate this specific PR, but which we would like to address before we put further smart search development on ice:

* Entering plain text and selecting it from the dropdown doesn't seem to do anything
* Selecting either a tag or a category makes it impossible to continue typing, since our search field currently only supports a single token

These will be resolved by future development on the smart search to support multiple tokens, but since those may be a ways into the future I am going to propose the following mitigations to these issues:

- [ ] Selecting _anything_ should execute the search immediately (we should make it easy to switch off this change since we won't want that behavior when we ultimately implement multi-token support). In this case it might also make sense for the plaintext option to be selected by default which demonstrates that hitting "enter" will confirm that selection.
- [ ] Let's indicate this behavior by adding a "Go →" label

   Artboard

- [x] When something is selected, let's gray out the text box to indicate it's not typable and realign and label the clear button. I propose:

   Artboard

   (I suspect @SaptakS will want to at least make some color contrast modifications here—I threw this together quickly 😅)

I'm open to other thoughts on either of these if people have different ideas.

@dkaoster I know you volunteered to help get this over the line as a volunteer contributor—and we're very appreciative of the offer!—but I think we can handle these last changes in house and we'll let you know if we have any questions. Thanks for everything!

chigby:
I've added the greying-out effect for when further selections can't be made to the smart search bar. Let me know what you all think! For the specific grey color I used something we already had for the inactive tag background color, but this can be easily changed. 

This change also seemed incompatible to me with the "immediately perform a search upon selection" strategy, so I did _not_ implement that. Let me know if I assumed incorrectly, though. I prefer the greying-out approach since adding the "search immediately" feature is something we will have to remove later, once multiple selections are supported, and I think we should do as few big changes to the primary way users interact with these controls as possible. The grayed-out state doesn't change the workflow of how the search bar is used (i.e. type, select, submit search), it makes the state of the controls more obvious and might even be something we want to keep if we need such a state as the capabilities of this feature evolve.

Anyway, this is ready for review.

harrislapiroff:
@chigby I made an incorrect assumption which was that we would need a way to display a filled in field after the search had been executed. I see now that's not the case. I think you're right that they're incompatible requests since we don't need that. Personally my inclination is to prefer the "execute search immediately" approach between the two of them because:

1. It reduces an extra step
2. In the moment before a user makes a selection, the "Go" label will eliminate any confusion about whether or not they will be able to add more than one tag, category, etc.

But I'm open to discussion about it! Given that it may be half a year or more before we get back to smart search I'm inclined to prioritize the experience now if it won't be too challenging to implement.

chigby:
Gotcha! I don't think it's too challenging.

chigby:
Updated the code to immediately execute tag and category searches upon selection.

SaptakS (approved):
I definitely think the changes are performing the actions correctly. I did expect to see the search bar filled when I clicked on a tag/category, but I guess since they are shown as labels in the search incident page, maybe it's fine to not show the value in search bar.

I kind of want to say that once we have implemented the search bar to some extent (which I understand will take some time), we should user test this part, rather than few of us in the team trying to guess what works best.

PR #1814, opened 2023-12-27
Fix table alignment

### LINKED ISSUE #1813: Fix alignment in incident data table
In some pages like https://pressfreedomtracker.us/all-incidents/florida-journalists-barred-from-covering-gov-desantis-signing-of-controversial-new-election-law/, the target row is overflowing and not aligning correctly with the rest of the rows. 
Screenshot 2023-12-21 at 12-51-16 Florida journalists barred from covering Gov  DeSantis signing of controversial new election law

I think a solution could be using grid in CSS instead of flex? But haven't really looked closely.

dkaoster:
## Description

Fixes #1813 

Fixes the alignment in incident data table.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

- Run this branch locally, open an incident page
- In the "targets" list, right click and duplicate nodes until it overflows. Make sure that the overflow stays within the column of the grid.

### General checks

Screen Shot 2023-12-26 at 8 05 57 PM

### DIFF
--- a/client/common/sass/components/_details-table.sass
+++ b/client/common/sass/components/_details-table.sass
@@ -12,7 +12,6 @@
 
 		+tablet-up
 			display: flex
-			flex-wrap: wrap
 			align-items: center
 
 		+desktop-up
@@ -22,6 +21,7 @@
 	&__label
 		font-size: $font-size-small-p
 		line-height: 1.43
+		flex-shrink: 0
 
 		+tablet-up
 			width: 17.5rem


SaptakS (approved):
Though I would prefer to move all the tables to a grid based layouting, I think this fix works for now (given we have proper breakpoints already dealing with some of the edge cases)

PR #1815, opened 2023-12-27
Update python django wagtail

### LINKED ISSUE #1785: Upgrade to Wagtail 5.2.x, Django 4.2.x and Python 3.12
Wagtail has just [released](https://docs.wagtail.org/en/stable/releases/5.2.html) their new LTS that also allows us to move to the latest LTS of Django. Also, as discussed in meeting, we planned to move to the latest stable Python release that we can migrate to.

Deliverables:
- [ ] Upgrade to Wagtail 5.2.x
- [ ] Upgrade to Django 4.2.x
- [ ] Upgrade to Python 3.12 (or 3.11 in case all our dependencies don't yet support 3.12)

Once this is done, we would also need to let the editorials know about the changes in UI (and any new [features](https://wagtail.org/blog/curate-content-better-and-make-it-greener-in-wagtail-52/) they might be interested in)

SaptakS:
## Description

Fixes #1785.

- Updates to django 4.2 and make necessary code changes
- Updates to Wagtail 4.2 and make necessary code changes
- Updates to Python 3.12 and updates other python packages to support Python 3.12

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Dependency update
- [ ] Config changes

## Testing

It might be worth checking the changelogs to see if I missed anything. Other than that, try building the container locally with the new dependencies, check all the pages and admin section and tests work without throwing any errors or deprecation warning.


SaptakS:
The coverage here is failing because I had to add a template tag for copyinh some removed wagtail templates that we are using in `/admin/incident-search/`. Haven't looked closely to know what tests to add for that, will take a look.

Also, this repo also seems to use Wagtail modeladmin and removing it seems more tricky (or rather more decision involved) rather than just changing some imports, so I might need some help on people who had added that to ensure I am not missing anything. I am also thinking if removing modeladmin should be handled in separate PRs, since all the repos seem to have it, and it's more of a structural change in all the cases , rather than just simply changing some imports to make it behave like snippet.

harrislapiroff:
@SaptakS It looks like [6 years ago](https://github.com/freedomofpress/pressfreedomtracker.us/commit/bee2006f181ae056b4ca22ec99cafa0a658c4003) we added the `incident-search` endpoint and made the admin search point to it in order to make searching incident pages *only* the default. It might be worth asking @KAMcCudden if this behavior is still desirable before we put a lot of work into migrating it.

As far as converting things to ModelAdmin, seems fine to do in a separate PR, but let's open tickets to track. Would be sad to get to the next upgrade cycle and realize we'd forgotten to do it 😅

harrislapiroff:
I don't know if we're still working on this, but I confirmed with @KAMcCudden that removing `/incident-search/` and references to it is fine. They understand the default admin search will be a combined search of all pages when we do so.

common/tests/test_tweet_embed_block.py
```diff
@@ -13,16 +13,16 @@ def test_clean_non_twitter_url(self):
...
 
     def test_clean_http_www_twitter_url(self):
         block = TweetEmbedBlock()
         with self.assertRaises(ValidationError) as cm:
             block.clean({
                 'tweet': EmbedValue('http://www.twitter.com/WagtailCMS/status/1413141835711606786'),
             })
-
-        self.assertEqual(cm.exception.params, {'tweet': ['Please enter a valid Twitter URL.']})
+        print(cm)
```
chigby: Tidying up: extra print statement here

devops/docker/ProdDjangoDockerfile
```diff
@@ -18,8 +18,8 @@ RUN npm install --global gulp-cli
 COPY ./ /src-files
 RUN cd /src-files && ( npm install && npm run build )
 
-# sha256 for python:3.9.18-slim-bookworm on linux/amd64 as of 2023-09-15
-FROM python:3.9-slim-bookworm@sha256:44b7f161ed03f85e96d423b9916cdc8cb0509fb970fd643bdbc9896d49e1cad0
+# sha256 for python:3.12.1-slim-bookworm on linux/amd64 as of 2023-12-13
+FROM python:3.12.1-slim-bookworm@sha256:815f9e7b04768469d7c1346adaaa9f7b1853b28a870e23c82ae18bd08777ff1d
```
chigby: It looks like there actually is a python 3.12.2 at this point, maybe we should bump all the versions?

blog/models.py
```diff
@@ -140,7 +140,7 @@ def serve(self, request, *args, **kwargs):
         # of the blog (including paginated and filtered URLs) simultaneously
         response['Cache-Tag'] = self.get_cache_tag()
 
-        if request.is_ajax():
+        if request.headers.get('x-requested-with') == 'XMLHttpRequest':
```
chigby: It's not clear to me that we need this code anymore. The comment inside this conditional refers to varying the layout template on ajax vs. non-ajax, but this behavior looks like it was removed in the redesign (see 396e24393ead4561ba6eacedc899f78d77518332)

common/models/pages.py
```diff
@@ -570,7 +570,7 @@ def get_context(self, request, *args, **kwargs):
         filters.pop('categories', None)
         context['filtered'] = bool(filters)
 
-        if request.is_ajax():
+        if request.headers.get('x-requested-with') == 'XMLHttpRequest':
```
chigby: This is another case where I think this varying with ajax is no longer code we need. Does `base.ajax.html` even exist anymore? I couldn't find it.

devops/docker/DevDjangoDockerfile
```diff
@@ -1,5 +1,5 @@
-# sha256 for python:3.9.18-slim-bookworm on linux/amd64 as of 2023-09-15
-FROM python:3.9-slim-bookworm@sha256:44b7f161ed03f85e96d423b9916cdc8cb0509fb970fd643bdbc9896d49e1cad0
+# sha256 for python:3.12.1-slim-bookworm on linux/amd64 as of 2023-12-13
+FROM python:3.12.1-slim-bookworm@sha256:815f9e7b04768469d7c1346adaaa9f7b1853b28a870e23c82ae18bd08777ff1d
```
chigby: See other note about python 3.12.2

forms/wagtail_hooks.py
```diff
@@ -2,7 +2,7 @@
 from wagtail import hooks
 
 
-@hooks.register('insert_editor_css')
+@hooks.register('insert_global_admin_css')
```
chigby: I'm also kind of wondering if the CSS added by this hook is still valid. To be honest, I can't really remember what this was supposed to do, but it does assume a particular CSS class hierarchy in the wagtail admin that may or may not still exist. I wonder if we can remove this, too, or if we should attempt to see what it was doing and fix it if it's no longer working?

incident/models/incident_index_page.py
```diff
@@ -235,7 +235,7 @@ def get_context(self, request, *args, **kwargs):
         # Could also be:
         # context['incident_count'] = incident_qs.count()
 
-        if request.is_ajax():
+        if request.headers.get('x-requested-with') == 'XMLHttpRequest':
```
chigby: Same comment here as for the blog index page. I basically think it's safe to remove setting this context variable to anything: `layout_template` doesn't seem like it's used anywhere and the `base.ajax.html` is not present anymore (as far as I can tell).

incident/models/incident_index_page.py
```diff
@@ -255,7 +255,7 @@ def serve(self, request, *args, **kwargs):
         # simultaneously
         response['Cache-Tag'] = self.get_cache_tag()
 
-        if request.is_ajax():
+        if request.headers.get('x-requested-with') == 'XMLHttpRequest':
```
chigby: Same as above. If we remove the conditional in `get_context`, we don't need it here, either.

chigby (changes requested):
Left a few comments, mostly it is other things we can remove instead of update. They're not blockers, but if you agree I think we might as well do them now. Also: do we need to upgrade `wagtail-django-recaptcha`? The changelog implies only the new version is compatible with wagtail 5.2

forms/wagtail_hooks.py
SaptakS: This CSS is definitely supposed to be used in the nested form fields in the form group of a form page. However, like you correctly pointed out, they are not getting applied because of the change in wagtail classname structure. Honestly, I don't feel like any of the CSS designs that we have mentioned here are needed (or even makes sense). The only one that kind of makes sense is for the multiple option field, we _can_ make it `width:100%`, but I don't know with the new design if it's really needed. CC: @harrislapiroff 

This is how it looks without width: 100%:
image


And this is with width: 100%:
image

SaptakS:
The Circle CI builds are failing because cimg for Python 3.12.2 [has not been released yet](https://github.com/CircleCI-Public/cimg-python/issues/227). Should we wait for them to release and then rerun the CI to make it pass? I don't feel the best decision to stay to 3.12.1 is because Circle CI didn't still release their latest Python image.

chigby:
Good point about image being a Circle CI image rather than an official Python one. I think the differences between 3.12.1 and 3.12.2 are not going to matter too much for CI testing and verification purposes, so one vote from me for keeping our own version on the latest and not letting Circle CI's releases block us.

forms/wagtail_hooks.py
chigby: My inclination here is to remove our CSS. The newer admin design seems like it doesn't need it anymore. We were just ahead of the curve, maybe :slightly_smiling_face:

harrislapiroff: I agree. My vague recollection is that nested inlines used to be very poorly supported style-wise (maybe not supported) and we were probably accounting for that?

chigby (approved):
Marking this as approved to expedite the completion process. Please merge once the CI 3.12.1 image is restored & working and the admin CSS is removed.

PR #1817, opened 2024-01-02
Updates selenium

SaptakS:
## Description

Updates selenium to fix NULL Pointer Dereference vulnerability. Selenium is only used in dev and doesn't affect prod.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Vulnerabilities update
- [ ] Config changes


chigby (approved):
Update looks good to me!

PR #1837, opened 2024-03-06
Adds a install setuptools step in circle CI

SaptakS:
## Description

We are getting CI fails in safety check because of missing setuptools. So this PR adds a step to install setuptools.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [x] Config changes


SaptakS:
@chigby updated this PR to use Python 3.11 in safety check job

chigby (approved):

PR #1844, opened 2024-04-09
Fix mobile BarChart Slider 🐞 

### LINKED ISSUE #1840: Slider not working on mobile bar charts
image

Expected behavior: scrubbing across the slider changes which bar is selected and shows details for it

Actual behavior: scrubbing across the slider doesn't move the slider

This makes it look like a weird design element and like we're weirdly highlighting one bar for no reason!

zrobert7:
## Description

The BarChart component has a slider on mobile. This was not working because it relied on mouse events. On mobile, we need to use touch events. As far as I can tell, this Slider is currently _only_ used on mobile, but I opted to make the same component functional on Desktop or Mobile in case that changes down the line. 

Fixes #1840 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

I tested with the Chrome & Firefox mobile device emulators. 

### General checks

https://www.loom.com/share/0fe41edb438047489ba24f6b5e985a54

### DIFF
--- a/client/charts/js/components/Slider.jsx
+++ b/client/charts/js/components/Slider.jsx
@@ -4,7 +4,7 @@ import { clamp, first, last } from 'lodash'
 
 export default function Slider({ elements, xScale, y, setSliderSelection, sliderSelection, idContainer }) {
 	const onSliderReleaseRef = useRef(null)
-	const onMouseMoveRef = useRef(null)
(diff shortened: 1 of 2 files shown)


chigby (approved):
This change is looking great to me, and as described, it works in more contexts as well. Thanks!

PR #1894, opened 2024-08-28
Updates webpack and micromatch

SaptakS:
## Description

Updates webpack. https://github.com/advisories/GHSA-4vvj-4cpr-p986
Updates micromatch: https://github.com/advisories/GHSA-952p-6rrq-rcjv

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Vulnerabilities update
- [ ] Config changes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,7 @@ jobs:
     environment:
       SETUPTOOLS_USE_DISTUTILS: stdlib
     machine:
-      image: ubuntu-2204:2023.07.2
+      image: ubuntu-2204:current
     working_directory: ~/tracker
     steps:
       - checkout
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -59,7 +59,7 @@ jobs:
     environment:
       SETUPTOOLS_USE_DISTUTILS: stdlib
     machine:
-      image: ubuntu-2204:2023.07.2
```
harrislapiroff, line 62: Is this intentionally part of this commit? (Not sure if it makes a difference one way or another, but curious about it)

harrislapiroff (approved):
Curious about the motivation for the ubuntu upgrade, but it seems like a good idea! All LGTM

.circleci/config.yml
SaptakS: Yes! Circle CI has deprecated some images and these are one of them. So the CI was failing without updating the images.

SaptakS: https://discuss.circleci.com/t/linux-image-deprecations-and-eol-for-2024/50177

harrislapiroff: Ah, makes sense!

PR #1896, opened 2024-09-04
Updates django to 4.2.16 and cryptography

SaptakS:
## Description

Updates Django to fix security releases. https://www.djangoproject.com/weblog/2024/sep/03/security-releases/
Updates cryptography: https://github.com/freedomofpress/pressfreedomtracker.us/security/dependabot/109

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Vulnerabilities update
- [ ] Config changes


chigby (approved):
Update looks good

PR #1907, opened 2024-12-09
Update geonames fixtures to correct the name of "New York"

### LINKED ISSUE #1797: Fix "New York City" geoname entry to match our incident database entries
In our incidents database, there are a number of incidents with the `city` field set to `New York`. There is no corresponding geoname for this city. The one in the `geonames.geonames` table is `New York City`. We want to update the geoname to match our use of the city name.

In my opinion, this is best done by running this query:

```sql
UPDATE geonames_geoname SET name = 'New York' WHERE name = 'New York City';
```

In the Django shell, this is:

```python
from geonames.models import GeoName
GeoName.objects.filter(name="New York City").update(name="New York")
```

To fix the incidents currently in production:

1. Run the above query/shell commands.
2. Either:
    - Run the `./manage.py populate_geodata` command. This will update all incidents to have their lat/long data reflect the updated geonames DB, including the ones with the city "New York"
    - Refresh the incident lat/long data one-by-one loading the ones you want to fix in the wagtail editor and issuing a "Publish" command.

Additionally, we will want to apply this change everywhere we can, so we should update the fixtures to use the correct name.

The easiest way to do this is to:

1. Ensure that the fixtures have been loaded in the local dev environment (`docker compose exec django ./manage.py loaddata --app geonames cities500-us-only.json.xz`)
2. Update the DB using the query or shell commands from above.
3. Regenerate the fixtures: `docker compose exec django ./manage.py dumpdata geonames -o geonames/fixtures/cities500-us-only.json.xz`

chigby:
## Description

Fixes #1797

Updates the geonames fixtures to change "New York City" to "New York"

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

1. Load the fixtures in the dev environment (`docker compose exec django ./manage.py loaddata --app geonames cities500-us-only.json.xz`)
2. Find an incident, and edit the city to be `New York` and the state to be `New York (NY)`, then save the page.
3. In the Django shell or DB shell, look up the longitude and latitude given to that incident page:
```
In [1]: from incident.models import IncidentPage
In [2]: i = IncidentPage.objects.get(pk=131)
In [3]: i.latitude
Out[3]: 40.71427

In [4]: i.longitude
Out[4]: -74.00597
```

It should also be visible if you create a bubble map somewhere and group incidents by city.


harrislapiroff (approved):
I feel like there should be a GitHub tool for diffing compressed files, but, it's fine probably.

PR #1919, opened 2025-01-08
Removes deprecated fields

### LINKED ISSUE #1914: Remove deprecated Arrest and Subpoena fields
Hi team, there are a few fields remaining in the backend for PFT incidents that we no longer use. Could you remove them when possible? 

Arrest/Criminal Charge fields: remove Status of charges, Current Charges, and Dropped Charges.
Screenshot 2025-01-02 at 4 36 36 PM

Subpoena/Legal Order fields: remove the first "Legal Order Type" field right below the "Subpoena/Legal Order" heading.
Screenshot 2025-01-02 at 4 37 29 PM

Thank you!

SaptakS:
## Description

Fixes #1914.

* Removes status of charges, current charges and dropped charges from arrest/criminal charge
* Removes legal order type from subpoena fields

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

Check all the functionalities in the website and admin work related to incidents

### General checks


incident/utils/incident_filter.py
```diff
@@ -867,11 +820,9 @@ class IncidentFilter(object):
         },
         'targeted_institutions': {'filter_cls': TargetedInstitutionsFilter, 'text_fields': ['title']},
         'arresting_authority': {'filter_cls': RelationFilter, 'verbose_name': 'Arresting authority'},
-        'status_of_charges': {'filter_cls': StatusOfChargesFilter},
```
chigby, line 870: So, I think this filter should not be removed. We still want users of the site to be able to filter incidents by the status of their charges, and the functionality of `StatusOfChargesFilter` has already been changed to apply to the new style of charge metadata. I think what needs to happen here is the definition for this filter needs to be moved from the `filter_overrides` section to the `_extra_filters` section. `filter_overrides` have to correspond to the name of a field on the `IncidentPage` model, and since `status_of_charges` has been removed as a field, it won't work in this section. What we will want here is something resembling how `legal_order_status` is handled.

incident/utils/incident_filter.py
```diff
@@ -867,11 +820,9 @@ class IncidentFilter(object):
...
         'venue': {'filter_cls': RelationFilter, 'verbose_name': 'venue'},
         'state': {'text_fields': ['abbreviation', 'name']},
         'charges': {'filter_cls': ChargesFilter, 'text_fields': ['title'], 'verbose_name': 'Charges'},
-        'legal_order_type': {'filter_cls': LegalOrderTypeFilter},
```
chigby, line 874: Same comment as `StatusOfChargesFilter` above. I think we want to preserve this filter on the front-end, so it will have to be moved to `_extra_filters`.

incident/tests/test_filtering.py
```diff
@@ -353,84 +353,6 @@ def test_should_filter_charges_by_title_as_one_field(self):
...
-        )).get_queryset(strict=True)
-
-        self.assertQuerysetEqual(incidents, [])
-
-        incidents = IncidentFilter(dict(
-            categories=str(category.id),
-            status_of_charges=desired_status,
-        )).get_queryset(strict=True)
-        self.assertQuerysetEqual(incidents, [target1])
-
```
chigby: If we are keeping the `status_of_charges` filter (`StatusOfChargesFilter`) then we will want to keep these test cases, too.

incident/api/tests/test_csv.py
```diff
@@ -288,14 +287,6 @@ def test_arresting_authority_is_correct(self):
...
         )
 
-    def test_status_of_charges_is_correct(self):
-        self.assertEqual(
-            self.result['status_of_charges'],
-            ', '.join(
-                charge.updates.order_by('-date').first().get_status_display() for charge in self.incident.charges.all()
-            )
-        )
-
```
chigby: I think we do want to preserve this test, as it does not depend on the `status_of_charges` field directly, and instead uses the new style charge metadata.

incident/api/serializers.py
```diff
@@ -327,4 +321,3 @@ class CSVIncidentSerializer(VariableFieldSerializer):
     links = serializers.CharField(source='link_summary')
     equipment_broken = serializers.CharField(source='equipment_broken_summary')
     equipment_seized = serializers.CharField(source='equipment_seized_summary')
-    status_of_charges = serializers.CharField(source='status_of_charges_summary')
```
chigby: I think we can keep this here, as `status_of_charges_summary` does not use the removed `status_of_charges` field and instead uses the new style charge metadata.

incident/tests/test_filtering.py
```diff
@@ -2190,34 +2015,6 @@ def test_targeted_institution_filtering_via_targeted_journalists(self):
...
-        incidents = IncidentFilter({
-            'legal_order_type': self.desired_type,
-        }).get_queryset(strict=True)
-
-        self.assertQuerysetEqual(
-            incidents,
-            [self.legal_order1.incident_page],
-        )
-
-
```
chigby: Similarly to the `status_of_charges` filter, I think we will want to keep the `legal_order_type` filter, which uses the `LegalOrderTypeFilter` class, and that does not involve the `legal_order_type` field.

incident/tests/test_get_serialized_filters.py
```diff
@@ -41,32 +41,6 @@ def test_serialize_general__search_only(self):
...
-        ])
-        general, category = get_serialized_filters()
-        legal_order_type_filter = category['filters'][0]
-        self.assertEqual(legal_order_type_filter, {
-            'name': 'legal_order_status',
-            'title': 'Legal order status',
-            'type': 'choice',
-            'choices': LegalOrderStatus.choices
-        })
-
```
chigby: I don't think we want to remove either of these tests, as both `legal_order_status` and `legal_order_type` are filters we want to preserve.

chigby (changes requested):
I'm requesting some changes on this PR because I think there's been a slight amount of confusion as to what is being removed, and I think too much has been. Unless I'm misunderstanding the issue (and there's a chance I am, so let me know if I'm mistaken here, please), we want to remove the database fields for `status_of_charges` and `legal_order_type` (as well as dropped/current charges, but those are not affected by this situation), but not the corresponding filters. Those filters have kept the names of the fields for backwards compatibility purposes, but the filter code uses the new, inline-based model organization rather than the fields of those names.

The code changes needed to preserve the filters is relatively simple (moving the definitions from `filter_overrides` to `_extra_filters` -- I explain a little more in a comment near the relevant code), but there is a decent amount of code that's deleted here involving those filters that should be brought back, I think. 

Let me know what you think of this, I hope it is not too annoying to orchestrate this, and that everything I've said makes sense. The filter system and how they're defined is definitely not as intuitive as it could be, and probably made more sense originally, when the filters more straightforwardly operated directly on the fields of `IncidentPage` and that connection was more obvious. I think if we were designing the system now, it would make it clearer what was going on.

SaptakS:
Hmm. From the issue, I got the impression that they were not being used at all, but maybe I misunderstood. I will update the PR.

harrislapiroff:
I think Cameron is correct. If I understand correctly filters are only related to the fields insofar as they share the same name, but they don't actually query those fields, they query the "new" way of storing that data, so those should be left in place. It's just the old database fields that represent our old way of storing the data that should be removed. Apologies for the confusion!

incident/api/tests/test_csv.py
```diff
@@ -289,6 +289,7 @@ def test_arresting_authority_is_correct(self):
         )
 
     def test_status_of_charges_is_correct(self):
+        print(self.result)
```
chigby: Cleanup: this print statement can be removed.

incident/tests/test_get_serialized_filters.py
```diff
@@ -223,7 +222,7 @@ def test_serialize_general_and_category_fields(self):
                         'title': 'Status of charges',
                         'type': 'choice',
                         'name': 'status_of_charges',
-                        'choices': STATUS_OF_CHARGES,
+                        'choices': None,
```
chigby: So, this is ringing some bells for me. The serialization of the `status_of_charges` filter should still offer choices.

chigby: I've done some digging around in the code and it seems like there is some code lingering around in the way `ChoiceFilter` (and `MultiChoiceFilter`) works that will break `StatusOfChargesFilter` if the `status_of_charges` field is removed. You can see this by going to the All Incidents page locally (http://localhost:8000/all-incidents/) and it will throw an error.

I can think of two ways to handle this:

1. Define the choices on the `model_field` attribute when initializing the `StatusOfChargesFilter`, like this:
 
```py
model_field=CharField(choices=choices.ARREST_STATUS),
```

(on line 894 of `incident_filter.py`). The reason `LegalOrderStatusFilter` and `LegalOrderTypeFilter` don't need this is that they override how the `serialize` method works, which `StatusOfChargesFilter` does not. 

2. The second way involves re-writing `ChoiceFilter` and its friend `MultiChoiceFilter` to be more flexible. This reduces some duplication and means we don't have to override `serialize` on the above filters, either, we can define the choices more directly.

The code for this is more involved, so I made [my own branch with the details](https://github.com/freedomofpress/pressfreedomtracker.us/commit/22ff9c2832e0d196bbe919cd57b3d00872342f4c). Feel free to use this code or merge the branch or anything you like. It is not as simple as the other change, but it does make these three similar filters work in a more similar way.

chigby (changes requested):
I left another comment because I noticed that the All Incidents page was not working locally for me. The fix is not obvious but I've left the code for two possible ways of doing it in the comment. Let me know what you think. Hopefully this is the last lingering change that needs to happen :crossed_fingers:

SaptakS:
@chigby merged your branch and cleaned up the print statements

chigby (approved):

PR #1932, opened 2025-02-10
Add spacing between nav & social links

eloquence:
## Description

Follow-up to #1928 as things otherwise look a bit crammed with the number of navigation links we have in prod.

## Type of change

- [x] Cosmetic fix

## Testing

1. Populate nav to look like prod
2. View dev env and ensure spacing is appropriate

### DIFF
--- a/client/common/sass/components/_footer-info.sass
+++ b/client/common/sass/components/_footer-info.sass
@@ -31,6 +31,7 @@
 		display: flex
 		align-items: flex-end
 		justify-content: space-between
+		gap: 2rem
 
 		+desktop-up
 			align-items: center


SaptakS:
@eloquence it does add the spacing, but it also wraps the item. Is that the desired style? 
image

The other options can be reducing the gapping between the social icons or increasing the max width of the container, to avoid the wrapping.

harrislapiroff (approved):
lgtm

eloquence:
Thanks for noting that @SaptakS. I have the menu replicated in my dev env; let me poke at it a bit locally to see if increasing the max-width of the container would have any unwanted side effects - if not, I'll open a follow-up PR to tweak that a bit further.

eloquence:
I think visually we want to probably keep the text and footer navigation aligned.

Assuming so, here's what it looks like with the width bumped to 800:

image

And here's the current state:

image

My inclination would actually be to preserve the current state. Yeah, the wrap is a bit inelegant, but the lines are already running a bit above the common [readability recommendation](https://baymard.com/blog/line-length-readability) of 50-75 characters.

@harrislapiroff @SaptakS Let me know if you feel otherwise, happy to revisit/tweak.

PR #1954, opened 2025-04-15
Replace images from the deprecated Unsplash URL with Lorem Picsum

harrislapiroff:
see: https://picsum.photos/

## Description

Does what it says on the tin. I was getting failures from `./manage.py createdevdata` locally before this change.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

1. Delete your dev database with `docker compose rm -v postgres`
2. Run `docker compose run ./manage.py createdevdata` and confirm that it succeeds
3. Run `docker compose up` and confirm that you see a website at locahost:8000 which has images.

### Post-deployment actions

### General checks

### DIFF
--- a/common/management/commands/createdevdata.py
+++ b/common/management/commands/createdevdata.py
@@ -159,7 +159,7 @@ class Command(BaseCommand):
         )
 
     def fetch_image(self, width, height, collection):
-        url = f'https://source.unsplash.com/{width}x{height}?animal'
+        url = f'https://picsum.photos/{width}/{height}'
         response = requests.get(url, timeout=5)
         if response and response.content:
             CustomImageFactory(


SaptakS:
In FPF website, we started using [local pictures and fixtures](https://github.com/freedomofpress/freedom.press/pull/1618/commits/8b8273a35388819d3784016ac51956bd8d822a35) for our createdevdata instead of depending on a 3rd party services. We might want to consider that for all our repos.

harrislapiroff:
I would be OK with that approach instead. I can look at the code and try to adapt it for this repo.

zrobert7:
Update: let's merge this as an interim solution and then switch to local images

SaptakS (approved):

PR #1962, opened 2025-04-25
Aligns images and embeds correctly

### LINKED ISSUE #1912: Video block alignment issues in blog post
When using the Video building block to embed an FPF youtube video on [this page](https://pressfreedomtracker.us/blog/journalist-arrests-surge-in-2024/), @KAMcCudden found that the left/right/full width alignment buttons that appear as options in the backend were not seemingly different on the front end other than a few millimeters. See in screenshots very small differences when using left align (the caption runs to 2 lines but the text won't run around it) versus full width (caption at 1 line). She published with fill width, but would have opted for a left / right alignment and let the text run around it instead if that option was available. 

Level of urgency from (1) “get to it when you have a spare moment” to (5) “must be done ASAP”: 1
Level of importance from (1) “nice-to-have” to (5) “we need this to function”: 4

Screenshot 2024-12-13 at 13 03 57
Screenshot 2024-12-13 at 13 04 16

SaptakS:
## Description

Fixes #1912.

- Makes left and right aligned media only take up a maximum of 60% width. Without this constraint, depending on the size of the media and the caption, it was trying to take up 100% space and hence the text was not wrapping around it
- Makes the width of the media independent of the caption size. Before this change, the width of embed media was dependent on the width of the caption size.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

- Add an image or youtube embed
- Apply left or right alignment
- See that the text actually wraps around

### General checks

image

### DIFF
--- a/client/common/sass/components/_media.sass
+++ b/client/common/sass/components/_media.sass
@@ -1,16 +1,19 @@
 .inline-media
-	&.full-width
-		margin-inline: 0
+	width: 100%
+	margin-inline: 0
 
-	&.left
-		float: left
-		margin-bottom: 1em
-		margin-right: 1em
+	+tablet-up
+		&.left
+			margin-bottom: 1em
+			margin-right: 1em
+			float: left
+			max-width: 60%
 
-	&.right
-		float: right
-		margin-bottom: 1em
-		margin-left: 1em
+		&.right
+			margin-bottom: 1em
+			margin-left: 1em
+			float: right
+			max-width: 60%
 
 	&__caption
 		line-height: 1


client/common/sass/components/_media.sass
```diff
@@ -1,16 +1,19 @@
...
-		margin-left: 1em
+	width: 100%
+	margin-inline: 0
+
+	+tablet-up
+		&.left
+			margin-bottom: 1em
+			margin-right: 1em
+			float: left
+			max-width: 60%
```
SaptakS, line 10: This value is slightly randomly decided. I just felt the media should take up slightly more space than the text even when the text wraps. But if people feel strongly about keep this to 50% so that they take up almost equal space, or even increasing it to make the media take up more space than now, let me know.

chigby (approved):
I tried this out locally with various lengths of captions and it works for all of them. I think the 60% width is visually appealing, so let's keep it.

PR #1990, opened 2025-07-02
Scale homepage bubble map bubbles by providing domain

### LINKED ISSUE #1924: Alternative to bubble map on homepage
The bubbles on the map on the Tracker's homepage are likely to become unreadable soon as incidents rise. Is there a good alternative to the bubbles, like a heat map? 

Completion criteria:

- [ ] Bubbles should be scaled to the size of the data. So, pick a max reasonable bubble diameter and match that up to the largest city.

zrobert7:
## Description

Fixes #1924 

- Fetch the max city size to calculate the domain before scaling
- Bound the range at radius 60 

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

Before & after this change, where the largest bubble is for a state with 500 incidents and the smallest is for 1

Pasted Graphic 1
Pasted Graphic

### DIFF
--- a/client/charts/js/components/USMap.jsx
+++ b/client/charts/js/components/USMap.jsx
@@ -27,7 +27,7 @@ const defaultPaddings = {
 
 const markerSize = {
 	min: 5,
-	max: 30,
+	max: 60,
 }
 
 const markerBorder = {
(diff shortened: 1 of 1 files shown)


client/charts/js/components/USMap.jsx
```diff
@@ -72,8 +72,8 @@ export default function USMap({
 	const hasLatLon = ({ latitude, longitude }) => latitude && longitude
 
 	// Scale markers size depending on the number of incidents in a city
-	const markerScale = d3.scaleSqrt().domain([0, 30]).range([markerSize.min, markerSize.max])
-
+	const values = dataset.map((d) => d.numberOfIncidents);
```
zrobert7, line 75: hide whitespace during review! Real change is here

SaptakS (approved):
I think the code does what is expected. I think it might be useful for @harrislapiroff or stakeholders to check if this solution actually satisfies the need.

harrislapiroff (approved):
Screenshot and code lgtm!

PR #1996, opened 2025-07-23
Stabilize ordering for legal orders

### LINKED ISSUE #1951: Subpoenas switching place
### Description

I just published this incident - https://pressfreedomtracker.us/all-incidents/texas-tv-station-subpoenaed-in-murder-case-order-set-aside/ - that includes two legal orders. On the back end, the legal order that includes a "quashed" entry comes first. BUT on the front end, that legal order appears second. It's weird!

Image
Image

### Expected Behavior

I'd like the back and the front end to match, and for the "quashed" one to come first.

### Browser and OS
Firefox, MacOS.

### Urgency

### Notifications

_No response_

chigby:
## Description

Fixes #1951

This PR defines a sort ordering for the `LegalOrder` model by `pk`. This means when we display it on the front end, they will be sorted by their database ID in ascending order. Right now, there is no defined order, so it's no clear what will come first from page to page. 

I expect that this ordering will match the one in the admin area. I did some local testing and that seemed to be the case. If it continues to re-occur, we can revisit this, I think.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

Try to find some pages with legal orders and check that the front end display matches the one in the admin.
If you want to be thorough, try finding ones where the query results are returned not in ID order. I used a `manage.py shell` script to find some of those:

```python
from django.db.models import Exists, OuterRef
from incident.models import LegalOrder, IncidentPage

legal_orders = LegalOrder.objects.filter(incident_page=OuterRef("pk"))
xs = IncidentPage.objects.annotate(has_legal_orders=Exists(legal_orders)).filter(has_legal_orders=True)

for x in xs:
     lo_ids = [lo.pk for lo in x.legal_orders.all()]
     is_sorted = all(b >= a for a, b in zip(lo_ids, lo_ids[1:]))
     if not is_sorted:
         print(repr(x), x.pk, lo_ids)
         break
```

Maybe this will help you find a page whose legal order ordering changes when the changes from this PR are applied.


chigby:
@sophieongithub Has told me this solution is fine for the overall issue, compared to making Legal Orders fully re-orderable at will.

SaptakS (approved):

PR #1998, opened 2025-07-25
Allow multiple authors [1/2]

### LINKED ISSUE #1982: Allow multiple authors
Split from #1566:

> Can we build out the functionality to allow multiple authors? The current approach has been creating a new author page with both individuals' names, but that breaks some of the searching functionality. Here are some examples:
> - https://pressfreedomtracker.us/blog/trump-crisis-mode-tweets-his-2000th-attack-press/
> - https://pressfreedomtracker.us/blog/federal-judge-reinstates-cnn-reporters-press-pass-after-trump-revoked-it-critical-coverage/
> - https://pressfreedomtracker.us/blog/government-documents-show-fbi-cleared-filmmaker-laura-poitras-after-six-year-fishing-expedition/

# Acceptance criteria

- Multiple authors from the same organization can be set. There is no need for a multi-author/multi-organization combination for now.

# Urgency and priority

Medium

zrobert7:
## Description

Fixes #1982 

- New BlogAuthor model (modeled after the `IncidentAuthor` type
- Default migration to add BlogAuthor and authors field to Blog
- Custom migration to add existing `author` to `authors` field

In PR 2/2, I will remove the author field and change the frontend to rely on the `authors` field

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

image
image

### Post-deployment actions

Check the staging & prod runs of the migration


blog/models.py
```diff
@@ -157,6 +157,19 @@ class BlogIndexPageFeature(Orderable):
     ]
 
 
+class BlogAuthor(Orderable):
+    parent_page = ParentalKey('BlogPage', related_name='authors')
+    author = models.ForeignKey('wagtailcore.Page', on_delete=models.CASCADE, related_name='+')
```
chigby: I think we will want to limit the possible page types that are allowed to be an author. Specifically, I think `common.PersonPage` makes sense here. I'm not sure there are types besides that one that we want to allow to be configured as blog authors. Does that make sense? Is there another type of author page I'm not thinking of?

blog/models.py
```diff
@@ -157,6 +157,19 @@ class BlogIndexPageFeature(Orderable):
...
+
+    @property
+    def summary(self):
+        return self.author.title
```
chigby, line 166: This method is being flagged in the coverage check. Do you think it would be possible to write a simple test for what's happening here? Maybe put it in the `blog/tests/test_pages.py` file (or make a new `test_models.py` file? Up to you). We're doing something very similar actually in this test case: 

https://github.com/freedomofpress/pressfreedomtracker.us/blob/5e57337a5c2b5189d3c9332b993b56a55dd2f12d/incident/tests/test_models.py#L92-L103

blog/tests/factories.py
```diff
@@ -104,3 +104,23 @@ class Params:
...
+    def authors(self, create, extracted, **kwargs):
+        if not create:
+            return
+
+        if extracted:
+            for author in extracted:
+                BlogAuthorFactory(
+                    parent_page=self,
+                    author=author,
+                )
```
chigby: This method is also being flagged by coverage. IMO, the simplest thing to do here is update some of the uses of `BlogPageFactory` in `createdevdata` to use `authors` (which invokes this `post_generation` method) rather than `author` -- and passing in a list of authors rather than a single one. 

The `if not create:` branch would still not be reached in that case, and it might be fine to just remove it. `create` will be `False` when you call `BlogPageFactory.build` which I'm not sure we're doing anywhere in our code. I know we do have this little `if not create: return` tag all over the rest of our factories, but I suspect it's fine to not include it. I'm kind of wary of building too much extra code on top of `factory_boy` here anyway, but maybe that's a discussion topic for another day.

chigby (dismissed):
As with [`IncidentAuthor`](https://github.com/freedomofpress/pressfreedomtracker.us/blob/develop/incident/models/incident_page.py#L63), which is limited to `PersonPage` in what pages it's allowed to select, I think we should put some limitation on `BlogAuthor`. But other than that I think this PR looks good.

Oh, and the coverage check is failing. I left some comments that I think detail the simplest way forward on that front.

blog/models.py
chigby: Oh, I see the situation here. The `authors` foreign key on `BlogPage` is also set to `'wagtailcore.Page'`. Hmm. Let me take a look at what exists on staging.

chigby: Every blog page on staging with an author has a person page linked.

chigby:
```
>>> from blog.models import BlogPage
>>> from collections import Counter

>>> c = Counter([b.author.content_type.name for b in BlogPage.objects.select_related('author__content_type').exclude(author=None).all()])
>>> c
Counter({'person page': 112})
```

harrislapiroff (approved):
This looks good! I don't want to merge it until both PRs because it leaves the admin in a confusing state. Will look at the next one momentarily.

PR #1999, opened 2025-07-25
Remove author field & replace with authors [2/2]

### LINKED ISSUE #1982: Allow multiple authors

zrobert7:
## Description

Fixes #1982 

- Migration to remove the single author field
- Change all references to author to be authors

We can merge the first PR in this stack, test that it worked as expected, and then merge this one 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

Verify all visual & functional blog author behavior

### Post-deployment actions

### General checks

- [ ] Verify categories work

### If it's a major change

image


blog/templates/blog/_special_blog.html
```diff
@@ -26,8 +26,10 @@ <h1 class="special-blog__hero-title article-title" id="title">{{ page.title }}</
 		{% if page.publication_datetime %}
 			<p class="special-blog__hero-meta">
 				<span class="special-blog__hero-date">{{ page.publication_datetime|date:"F j, Y" }}</span>
-				{% if page.author %}
-					<span class="special-blog__hero-author"> — {{ page.author }}</span>
+				{% if page.authors %}
+					{% for author in page.authors.all %}
+						<span class="special-blog__hero-author"> — {{ author.summary }}{% if not forloop.last %}, {% endif %}</span>
+					{% endfor %}
```
chigby: This ends up looking a little odd in practice. I wonder if the dash should go outside the loop, here?

image

blog/templates/blog/_special_blog_card.html
```diff
@@ -17,9 +17,10 @@ <h2 class="special-blog-card__title">{{ post.title }}</h2>
 		</a>
 		<p class="special-blog-card__metadata">
 			{{ post.publication_datetime|date:"F j, Y" }}
-			{% if post.author %}
-				— {{ post.author }}
-			{% endif %}
+			{% if post.authors %}
```
chigby: This `{% if ... %}` tag doesn't appear to have a closing `{% endif %}`

blog/templates/blog/_blog_card.html
```diff
@@ -34,10 +34,12 @@ <h3 class="card-title card-title-{{ size }} blog-card--title">
 						{{ post.publication_datetime|date:"F j, Y" }}
 					</time>
 				</dd>
-				{% if post.author %}
+				{% if post.authors %}
```
chigby: This replacement, which occurs elsewhere in the code, too, is something that feels like it should work, but because of how Django handles relationships, does not actually work.

For single-relationships, `post.author` holds the object of the relationship. The `PersonPage` in this case. And if there isn't a value there, then it evaluates to `None` and the contents of this `if` block don't render.

For multiple relationships, the situation is more complex. `post.authors` holds the _relationship_ object. This is a regular Python class that has methods for manipulating the relationship set, and usually we have to call methods on it to do anything useful, for example calling `all` to iterate over the members of that relationship. It will never evaluate to `None` or `False` so this `if` block will always be rendered, even if the relationship is empty. It's unintuitive, I think. You can test this out, locally, by removing all the authors from a page and seeing it will still render "Written by" here.

The solution, I think, is to replace this with `post.authors.count`, which will give us a number of authors in the related set. If there are none, it's `0` and the `if` block will properly be skipped. If it gets used a few times, it might be worth storing the count in a variable using the [`with` tag](https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#with), but I leave that up to your judgment.

chigby (dismissed):

blog/templates/blog/_special_blog.html
zrobert7: good call!
image

harrislapiroff (approved):
Looks good! Let's get em both on staging for more in depth QA

PR #2012, opened 2025-08-18
Fix charges filter

### LINKED ISSUE #1991: Searching for a specific charge isn't pulling up the incident with that charge listed
### Description

Hi team, Stephanie noticed that a filter search for the charge "publishing: revealing, disclosing or divulging grand jury information" (https://pressfreedomtracker.us/all-incidents/?charges=publishing%3A+revealing%2C+disclosing+or+divulging+grand+jury+information&categories=Arrest%2FCriminal+Charge) is not pulling up either of the incidents with that charge listed (https://pressfreedomtracker.us/all-incidents/alabama-publisher-charged-over-report-on-grand-jury-investigation/ and https://pressfreedomtracker.us/all-incidents/alabama-reporter-faces-felony-charge-for-article-on-grand-jury-investigation/). Could you look into this for us? Thank you!

### Expected Behavior

_No response_

### Browser and OS

Firefox, macOS

### Urgency

### Notifications

_No response_

SaptakS:
## Description

Fixes #1991.

I implemented the first option mentioned in https://github.com/freedomofpress/pressfreedomtracker.us/issues/1991#issuecomment-3188375828. So this PR makes every filter except Tags & Categories split on `|`, but keep the Tags & Categories still split on `,`. I also added validatons to disallow `,` in title fields of Tags & CategoryPage.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

- Create a charge under Arrest/Criminal Charges in an incident page with a comma, then go to the frontend and check that it filters correctly with this charge.
- Check that multiple tags and multiple categories filter still work from frontend (both functionally and visually)
- Try creating a tag from admin using CommonTags, and check that it gives an error if there is a comma
- Try creating a CategoryPage from admin and check that it gives an error if there is a comma in the title

### General checks


chigby:
Should we be disallowing the pipe character in the filterable fields of the objects that are affected by this change?

SaptakS:
> Should we be disallowing the pipe character in the filterable fields of the objects that are affected by this change?
Probably a safe idea. I will look into that.

chigby:
Reviewing this PR is definitely making me wish that we had the time to give the front end of the incident search page the attention it needs. I think it's okay that this change is not backwards-compatible with the old API because the front end does not currently have a way to input search queries for multiple charges using the UI. You have to manually type those into a URL. 

It's also making me remember that the way the front end worked prior to the redesign was by ID numbers. So querying for two charges looked like `charges=33,100`. The API still supports this format, but the redesigned search front end was changed to use names/titles. It's something I wanted to fix towards the end of the redesign but we didn't have the capacity. In that world, we wouldn't have this "magic character" problem, either.

chigby (approved):
I've tested this and it works well. Charges (and other related names) with commas now filter correctly in the search results. I appreciate that this PR doesn't affect the front-end JS components, as that would be adding even more complications.

I do think the original issue here is the result of a lot of technical debt on both the front end and the back end of this system. The changes in this PR work great to fix this but, but are not necessarily something I'd feel strongly about preserving if/when we made deeper, more structural changes to the back end to better support a more feature-rich front end.

SaptakS:
> I do think the original issue here is the result of a lot of technical debt on both the front end and the back end of this system.

I agree. A lot of stuff from migration to new design and tech debt.

SaptakS:
@chigby this is ready for re-review

chigby (approved):

PR #2033, opened 2025-09-29
Add a stub template for Person Pages to prevent errors on preview

chigby:
## Description

PersonPages are technically subclasses of `Page` but they aren't used like pages. This meant they weren't given a template in the site redesign, but without a template the live preview feature will attempt to render it while editing.  This stub template will now prevent that error.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

When editing a person page in the admin, if the preview panel is opened, the panel should display a message about the impossibility of previewing this page type. There should be no error in the console or displayed in the preview pane.

image

### General checks


SaptakS (approved):
This makes sense to me... But I think at some point, we have to get rid of this tech debt and move to snippets.

Issue #2030, opened 2025-09-23
Increased granularity on PFT home page data viz


KAMcCudden:
### Description

General agreement from in-person 9/18 meeting that the home page data v's could use more useful "granularities"
E.g. "Past 6 months" not super useful/unfortunate that it defaults to that. Other options like "Past week" "Past month" or "place" 

### Urgency

Low (get to it whenever you have a free moment)

### Impact

Nice-to-Have

### Timing

there is not a specific deadline for this feature. I marked it Low and "Nice-to-have", but we'll want to consider upgrading it to "important" when building out features for https://github.com/freedomofpress/pressfreedomtracker.us/issues/2029 in Q1 as they are sorta connected

### Notifications

not yet — will see if it gets integrated to other tickets, etc.

harrislapiroff:
Proposed options:

Last 7 days | All Time | 2025 ▼

where the last item is a dropdown that lets you select a year

harrislapiroff:
Some concerns came up during web sync today:

* Will Last 7 Days be the default?
* Concern that last 7 days won't have much content given reporting lag

harrislapiroff:
Thanks for the PR @zrobert7. Let’s do a little more discussion of the questions above before we merge. My newest thought is it might make sense to do “past month” or “past 30 days” and bin the bar chart by week? We can label the ticks with the first day of the week (e.g., a bar for the week after September 1 would just be labeled “Sep 1”)

harrislapiroff:
Asked editorial for input

KAMcCudden:
Thanks for flagging. Understood 7 days may be empty some of the time, esp since incidents seem to be batched right now, meaning, e.g. a bunch will happen on 1 day rather than 1 a day. This weekend example: 6 assaults on Saturday; nothing Friday or Sunday. 

Using 'last month' as default with ability to see Xweek when chosen seems like a nice solution here

chigby:
@zrobert7 will submit a PR with the years presented in a drop-down format, and we will ponder what to do about "Last Month" vs "Last 30 days" vs "Last 7 days" in future discussions.

ryanwhitney:
For the granularity part of this, I've finally written down some options.

Thinking we could add two new windows: "past 4 weeks" and "past 3 months". Those give us recent past in a more expected timeframe than "past 6 months". When we have pre-pub data, we can look at 7 days.

Options would be:

| segment | # of bars | bucketing | |
[...]

PR #2042, opened 2025-10-15
PFT Homepage Viz granularity

### LINKED ISSUE #2054: Make year filter on homepage data viz a dropdown

zrobert7:
## Description

Related to #2030 
Resolves #2054

- move years to select drop down 
- change 6 months to 7 days 
- style open dropdown [TODO]

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

### Post-deployment actions

### General checks

- [ ] Verify categories work

### If it's a major change

https://github.com/user-attachments/assets/5a9ffc76-97db-43b7-b34d-1b45b69c4a52

### DIFF
--- a/client/charts/js/components/ButtonsRow.jsx
+++ b/client/charts/js/components/ButtonsRow.jsx
@@ -1,5 +1,7 @@
 import React from 'react'
 import Button from './Button'
+import DropdownButton from './DropdownButton'
+import { toNumbers } from 'canvg'
 
 const labelStyle = {
 	fontFamily: 'var(--font-base)',
@@ -9,6 +11,7 @@ const labelStyle = {
 export default function ButtonsRow({
 	label,
 	buttonLabels,
+	dropDownLabels = [],
 	defaultSelection = null,
 	updateSelection,
 	isButtonSelectable,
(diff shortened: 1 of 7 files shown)


zrobert7:
@harrislapiroff I definitely worry about the last 7 days having insufficient data - what if we have it as an option but keep 6 months as the default?

ryanwhitney (changes requested):
Sorry for the delay -- was looking at this stuff in figma this week but realize its not needed to unblock this!

Image
<br/><br/>

There is some variance across browsers (left to right: safari, mobile safari, chrome, ff). I think you might just have to do the full-on `appearance: none` to reset the select and use an svg icon for the handle since there isn't full support for styling these yet. It looks like the existing arrow-down.svg should work fine at 50%, but prob need to copypaste it so you can toggle stroke color with css.  

Should be able to get rid of the `top:` and `padding:` style overrides on the select once its appearance is reset too. I think this works well otherwise. lmk if I can be helpful at all with the css futzing!

zrobert7:
@ryanwhitney I would love your help css futzing to get this across the finish line! I updated to use the arrow-down.svg but I can't get the color to change on hover or click :/ 
image
image

ryanwhitney:
@zrobert7 

It is annoying but I think only inline SVGs can have path/color/stroke changed by CSS, so it has to be copy pasted into the markup and then positioned \*next\* to the select instead of using background-image. Did it on a separate branch: https://github.com/freedomofpress/pressfreedomtracker.us/commit/aad1c85169bd2fde19f37d35c7f7de7a7a4dba04

Or I am also just realizing we have arrow-down.svg and arrow-down-white.svg… so you could alternatively also just swap between those.

ryanwhitney (approved):
Looks good! 👏👏

PR #2092, opened 2026-02-13
Adds Draftail plugin to curlify quotation marks

### LINKED ISSUE #1957: Port smart quotes feature ("curlify")
https://github.com/freedomofpress/freedom.press/pull/2274 implements automatic conversion of quotation marks to "smart quotes", which are generally preferred in professional publishing settings. 

Once that feature has had time to land and we've gotten some end user feedback, we should consider porting it over to PFT to ensure the user experience is consistent across these two sites (securedrop.org, being more technically focused, probably benefits from not being too clever about quotation marks).

SaptakS:
## Description

Fixes #1957.

This implements similar code to https://github.com/freedomofpress/freedom.press/pull/2274/changes. I have temporarily also added the draftail codes to our eslint exception since they were complaining a lot. We can probably handle that separately if needed.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

Edit a rich text field that has curlify enabled as a feature. For example: the about field in the home page. But also any RichTextField that uses the default list of features or specified an editor. All of these fields should have curly quotes enabled. The way this works is, when typing text into the field:
- `'` should be replaced with either `‘` or `’` 
- `"` should be replaced with either `“` or `”`

### General checks

image


chigby (approved):
Tested this locally and everything looks to be in order. Thanks for moving over the rich text block definitions to use `editor` definitions, too, I think that makes the code a lot cleaner.

PR #2109, opened 2026-03-17
Adds content warning image component

### LINKED ISSUE #2046: Content warning screen for embedded images
### Description

Could we get a "content warning" screen for embedded images in incidents, so site visitors have to click through the screen in order to see the image? Thank you!

### Urgency

### Impact

Important

### Timing

_No response_

### Notifications

_No response_

SaptakS:
## Description

Fixes #2046.

Adds option in the backend to add a content warning to `AlignedCaptionedImageBlock`. I have used ListBlock to give the adding UI feel and also since it kind of handles both the checkbox and the default warning text behaviour with a single block. It does add a little bit of weird UX that kind of allows editors to add more warning (though the validation would show error). If someone feels this UI is bad and the checkbox combined with a charfield would be better, happy to change, but I kind of like it.

Also, I have added a ContentWarningImage web component to handle the blurring behaviour and unblurring on clicking the button. Currently, if the page is reloaded, the image becomes blurred again

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

- Edit an incident page, add an AlignedCaptionedImageBlock, add a content warning.
- The image should appear blurred with the text warning showing on top. Clicking the text warning should unblur the image
- Ensure the text warning as set in the backend is what gets shown on top.

### General checks

Admin:
image

Content warning image:
image


common/templates/common/blocks/image_block.html
```diff
@@ -1,4 +1,10 @@
 {% load wagtailimages_tags %}
 
 {% image value width-828 preserve-svg as my_image %}
-<img src="{{ my_image.url }}" width="{{ my_image.width }}" height="{{ my_image.height }}" alt="{% if my_image.image.attribution %}{{ my_image.image.attribution }}{% else %}{{ my_image.image.title }}{% endif %}">
+{% if content_warning %}
+<cw-image data-warning-text="{{ content_warning }}">
+{% endif %}
+  <img src="{{ my_image.url }}" width="{{ my_image.width }}" height="{{ my_image.height }}" alt="{% if my_image.image.attribution %}{{ my_image.image.attribution }}{% else %}{{ my_image.image.title }}{% endif %}">
```
paulschreiber, line 7: question: is attribution the creator of the image, such as the photographer? If so that's not especially useful as alt text. Is there another field (description?) we can use?

SaptakS: Yeah I agree. It is not really an useful alt text, and Wagtail in the recent versions has better alt text supports. But I think it might be beyond the scope of this PR, and should be handled as a separate issue/PR to:
- Either, just migrate all `ImageChooserBlock` in the model to `ImageBlock` which is the wagtail block with better alt-text support
- Or (and I prefer this) have custom image models and custom image blocks similar to the work done by @willbarton in freedom.press to have consistency.

client/common/js/cwImage.js
```diff
@@ -0,0 +1,52 @@
...
+ * It adds the cw-image-container class to address the styling and to toggle
+ * the content warning blurring. It also adds a button to remove the
+ * content warning blurring
+ *
+ * Example usage:
+ *
+ *     <cw-image>
+ *         <label for="field-id">Field Name</label>
+ *         <input type="text" id="field-id" />
+ *     </cw-image>
```
harrislapiroff: I *love* a Custom Element 😍 

**Nit:** Internal HTML should probably show an image or picture tag though

common/blocks.py
```diff
@@ -87,12 +87,30 @@ class AlignedCaptionedImageBlock(blocks.StructBlock):
...
+            label='Warning text',
+            help_text='Warning text shown on the button above blurred media',
+            default='This image may be disturbing. Tap or click to reveal.',
+            required=True,
+        ),
+        min_num=0,
+        max_num=1,
+        default=[],
+        label='Add content warning'
+    )
```
harrislapiroff, line 101: I'm sort of ambivalent on this pattern. I think we should leave it, but I want to register ambivalence so we can think about better approaches to this sort of thing in the future. Issues:

* The editor experience is *fine* but has some confusing issues, like you *can* add more than one content warning but it gives you an error when you have more than one
* I don't like defining our data model based on the least resistance way to implement the UX 😅 

A better approach would be something like a checkbox to add a content warning that caused the text box to show up and then just stored the content warning as either null or a string... but we don't currently have the scaffolding to create custom block interfaces. I was working on something like this in https://github.com/freedomofpress/freedom.press/pull/2689 so maybe we can generalize that later

harrislapiroff (approved):
Looks great! Very happy with the custom element implementation especially 🎉 

Added one nitpick which you can fix and merge as well as one comment for posterity.

The design here could be spruced up (the blur in particular doesn't handle image edges very well and the note could be styled in a more harmonious way) but I think this is good enough for a version one and if we want we can have @ryanwhitney take a design pass upon return.

PR #2113, opened 2026-03-24
Instagram + Bluesky Embeds

### LINKED ISSUE #2034: Add embed power for other social media platforms
### Description

In light of the crumbling of X, we'd like to be able to embed posts from other social media platforms - e.g., Bluesky, Instagram, etc. - in USPFT incidents.

Completion Criteria:

- [ ] Editors have the ability to embed Bluesky posts in articles
- [ ] Editors have the ability to embed Instagram posts in articles
- [ ] CSP allows all necessary domains for both embeds through

### Urgency

Low (get to it whenever you have a free moment)

### Impact

Nice-to-Have

### Timing

_No response_

### Notifications

_No response_

zrobert7:
## Description

Fixes #2034 

- New InstagramEmbedBlock
- New BlueskyEmbedBlock

**NOTE:** Extending the Wagtail EmbedBlock to support Instagram now needs approval + an API key from Meta: https://docs.wagtail.org/en/v4.1.7/advanced_topics/embeds.html?utm_source=chatgpt.com#facebook-and-instagram, so instead I added a custom Instagram block. I am not sure we want to be adding instagram embed scripts on our site --> see issue for discussion.

**NOTE**: Unfortunately for Bluesky, there is no way to directly embed from the post url. Bluesky needs the  `at://...` URI, or an extra call to the oEmbed API (https://docs.bsky.app/docs/advanced-guides/oembed). I feel like having editors copying the full embed via the below flow seems simpler than having editors find the URI within the embed code - thoughts?
image
image

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

### Post-deployment actions

### General checks

- [ ] Verify categories work

### If it's a major change

https://github.com/user-attachments/assets/ffe9f93d-fe41-4bb7-bee2-f4a1a0f6ea7a


redshiftzero:
Just waiting on tests and then this is ready for review

common/templates/common/blocks/instagram_embed.html
```diff
@@ -0,0 +1,8 @@
+<div class="instagram-embed">
```
paulschreiber, line 1: question: should the class be `instagram-embed-container` to match the bluesky behavior?

common/templates/common/blocks/instagram_embed.html
```diff
@@ -0,0 +1,8 @@
...
+    <blockquote
+        class="instagram-media"
+        data-instgrm-permalink="{{ value.url }}"
+        data-instgrm-version="14"
+        style="width:540px; margin:auto;">
```
paulschreiber: suggestion: add a rule for `.instagram-embed .instagram-media` to embed.scss instead of putting the style inline.

common/blocks.py
```diff
@@ -123,6 +123,31 @@ class Meta:
         label = 'Video'
 
 
+class BlueskyEmbedBlock(blocks.StructBlock):
+    embed_html = blocks.TextBlock(
+        required=True,
+        help_text="Paste the full Bluesky embed code, including &lt;blockquote&gt; and &lt;script&gt;."
```
paulschreiber: question: do you need to manually encode the HTML entities here? or does the rendering method handle that for you? if it doesn't, that seems like an XSS problem.

common/blocks.py
```diff
@@ -123,6 +123,31 @@ class Meta:
...
+
+    class Meta:
+        icon = "pick"
+        label = "Bluesky embed"
+        template = "common/blocks/bluesky_embed.html"
+
+
+class InstagramEmbedBlock(blocks.StructBlock):
+    def validate_instagram_url(value):
+        if "instagram.com" not in value:
```
paulschreiber:
```suggestion
        if not value.startswith("https://www.instagram.com/"):
```
Avoids false positives such as `https://www.example.com/blah/intagram.com/`

common/blocks.py
paulschreiber: suggestion: can authors paste in a post URL i.e. `https://bsky.app/profile/mshelton.bsky.social/post/3mickc2tw4k2h` which is is easier for them and less fragile (i.e. in case the embed markup changes).

common/templates/common/blocks/instagram_embed.html
zrobert7: see common/templates/common/blocks/tweet_embed.html! 

For instagram + twitter we have editors supply a URL. We can't do this with Bluesky, so I think its helpful to denote that difference in bluesky_emded.html

common/blocks.py
zrobert7: I want to allow pastes like https://instagram.com/p/C0DPdy98e4c as well, so I am going to leave this as is

common/blocks.py
zrobert7: This is needed to render the carrots in the help text 

Image

zrobert7: And regarding the URL pasting for bluesky - unfortunately no. Check out the PR description!

common/blocks.py
paulschreiber: How about this?
```python
if not re.match("^https://(?:www\.)?instagram\.com/", value):
```

paulschreiber:
> I feel like having editors copying the full embed via the below flow seems simpler than having editors find the URI within the embed code - thoughts?

@zrobert7 "Copy link to post" seems easiest. That matches how we embed YouTube, Twittter, IG, etc. Having Bluesky as a full code block will be confusing and lead to errors.

willbarton:
Since @zrobert7 is out, I've updated this to add a custom Wagtail oembed finder to support bsky.app URLs. This should enable us to paste Bluesky share URLs (via "Copy link to post") and paste them directly, just like tweets.

image

I've also updated the migrations for this change.

Maybe a note: we don't include any help text with the Instagram or Twitter embeds, so I haven't included any for this update to the Bluesky embed. Should we?

common/blocks.py
```diff
@@ -124,10 +124,7 @@ class Meta:
...
 
 class BlueskyEmbedBlock(blocks.StructBlock):
-    embed_html = blocks.TextBlock(
-        required=True,
-        help_text="Paste the full Bluesky embed code, including &lt;blockquote&gt; and &lt;script&gt;."
-    )
+    bluesky_post = EmbedBlock()
 
     class Meta:
         icon = "pick"
```
harrislapiroff, line 153: This is probably not something to resolve for this PR since it was already how `TweetBlock`s are structured, but I do wonder–is there a compelling reason for these to be `StructBlock` subclasses with a single sub-block instead of being direct subclasses of `EmbedBlock`? (Maybe this future-proofs somewhat in that it will let us add more fields in the future if we, e.g., add layout options to embeds?)

common/blocks.py
harrislapiroff, line 158: I think we may also want to give Instagram the oembed treatment? https://developers.facebook.com/docs/instagram-platform/oembed/

harrislapiroff (changes requested):
I think we want to use oembed for instagram too—let me know if I overlooked something that makes that impractical.

harrislapiroff:
> Maybe a note: we don't include any help text with the Instagram or Twitter embeds, so I haven't included any for this update to the Bluesky embed. Should we?

I think help text would be great!

common/blocks.py
willbarton: I wondered a little bit about that... I can see the future-proofing argument if we might want to add fields.  

`EmbedBlock` is a [subclass of `URLBLock`](https://github.com/wagtail/wagtail/blob/7f95b10ff731dd43b4dd909d2fd8e8fbdbb9d111/wagtail/embeds/blocks.py#L31C18-L31C33) (which is itself a [`FieldBlock`](https://github.com/wagtail/wagtail/blob/7f95b10ff731dd43b4dd909d2fd8e8fbdbb9d111/wagtail/blocks/field_block.py#L311)). So, if we encountered a need to add additional fields to this block surrounding the embed, I think this gives us more flexibility without too much complicated migration. But we aren't using it right now.

I chose what seemed to be the lightest-touch to getting Bluesky to use the built-in embed support, but happy to go deeper if we want to.

common/blocks.py
willbarton: @zrobert7 noted in the PR description:

> NOTE: Extending the Wagtail EmbedBlock to support Instagram now needs approval + an API key from Meta: https://docs.wagtail.org/en/v4.1.7/advanced_topics/embeds.html?utm_source=chatgpt.com#facebook-and-instagram, so instead I added a custom Instagram block.

We can pursue that, but we'll need keys (and to pass them in via the deployment) from a Meta Developer Account (which I don't know if we have). At that point, it might be best to split Instagram off from this PR and work on it separately (I know we went back and forth on that a bit).

harrislapiroff: @willbarton as far as you can tell, are there advantages to using oembed for instagram beyond cleaner code? It's a bit confusing to me that they'd require an API key for the embed endpoint but still offer a different way to embed that's equally good 😅

willbarton: @harrislapiroff I don't think so, unless the embed pattern changes to the point of breaking. AFAICT Instagram offers two routes to embedding: oembed via developer account, and the "Embed" option on a public post ("..." → "Embed"). 

What @zrobert7 has done here is to bring in the code that the "Embed" option provides and make a Wagtail block around it. That seems reasonable to me as an alternative to a developer account. I could see us needing to update this if Instagram changes their embedding patterns? 

I don't know what the risks/timeline for them doing that is, ***but*** given that it currently provides a bunch of HTML with inline styles, any upstream change would break what we have here would also break a lot of already-embedded posts everywhere else the code Instagram prompts people to copy has been used.

willbarton:
> I think help text would be great!

Help text included in [84dd7bb](https://github.com/freedomofpress/pressfreedomtracker.us/pull/2113/commits/84dd7bbedb08a452656c65c12c1b5285ee999826).

common/blocks.py
willbarton: Done in [1096994](https://github.com/freedomofpress/pressfreedomtracker.us/pull/2113/commits/1096994dd99ec1f1d2af3dd70c0ca98e3cc45588)

common/blocks.py
```diff
@@ -123,6 +123,36 @@ class Meta:
...
+    )
+
+    class Meta:
+        icon = "pick"
+        label = "Bluesky embed"
+        template = "common/blocks/bluesky_embed.html"
+
+
+class InstagramEmbedBlock(blocks.StructBlock):
+    def validate_instagram_url(value):
```
harrislapiroff: I think this should have an `@staticmethod` decorator or be outside the class

tracker/templates/base.html
```diff
@@ -9,6 +9,7 @@
 		{% block main %}{% endblock %}
 	</main>
 	{% include "common/_footer.html" %}
+	<script async src="https://www.instagram.com/embed.js"></script>
```
harrislapiroff: I hadn't realized this earlier, but we're now loading instagram js on _every_ page. In my mind that's a blocking privacy concern. We should limit this to only pages where that an ig embed is present.

client/common/sass/components/_embed.sass
```diff
@@ -1,2 +1,5 @@
-.tweet-embed
+.tweet-embed, .instagram-embed
 	max-width: 555px
+
+.bluesky-embed iframe
```
harrislapiroff: Instead of applying this to `.bluesky-embed` (which seems to be generated by Bluesky) can we apply it to `.bluesky-embed-container`? I want to limit the amount our code can break via third-parties changing their html.

blog/models.py
```diff
@@ -189,6 +193,8 @@ class BlogPage(MetadataPageMixin, MediaPageMixin, Page):
         ('image', AlignedCaptionedImageBlock()),
         ('raw_html', blocks.RawHTMLBlock()),
         ('tweet', TweetEmbedBlock()),
+        ('instagram', InstagramEmbedBlock()),
+        ('bluesky', BlueskyEmbedBlock()),
```
harrislapiroff: I realize this is an additional requirement I'm adding that wasn't in the original scope, but now that our list of blocks is getting long, let's start using block groups!

Let's put

* Instagram, Bluesky, and Tweet in a "Social Media" group
* Vertical bar chart, tree map chat, bubble map chart, hex bin map chart in a "Charts" group
* Everything else can stay in the default group for now

(Make this change for all pages where it applies)

common/blocks.py
harrislapiroff: One thing we do get from using an oembed block is validation that the URL goes to a real post. It would be a nice-to-have to get this on the instagram block if possible, but is not a blocker:

Image

harrislapiroff (dismissed):
This is close to ready, but there's some surprising depth here. I've left some comments for changes I want, at least one of which is a blocker for me.

common/blocks.py
willbarton: So, I have some additional thoughts on this — and I agree, this is surprisingly deep in places.

I think we could wrap the existing Instagram work in a custom Wagtail Embed finder, rather than a `StructBlock`, and then it could behave a lot more like the Twitter and Bluesky embeds. I'm going to sketch that out and see what it looks like, but I *think* it might make this kind of behavior a little more consistent across these embeds for admin users.

common/embeds.py
```diff
@@ -0,0 +1,31 @@
...
+        html = render_to_string('common/embeds/instagram.html', {
+            'url': clean_url,
+        })
+        return {
+            'provider_name': 'Instagram',
+            'type': 'rich',
+            'html': html,
+            'width': max_width,
+            'height': max_height,
+        }
```
willbarton, line 31: So, this is my attempt to square the Instagram embed-without-oembed with how Wagtail embeds generally behave. This is a [custom Wagtail `EmbedFinder`](https://docs.wagtail.org/en/latest/advanced_topics/embeds.html#custom-embed-finder-classes) that does what the Instagram embed block was doing, but within the Wagtail embeded content API. It has a couple advantages, first that the validation cycle works just like other embeds, if the `accept()` method fails, then the user gets the "Cannot find an embed for this URL" message. It also opens up the ability to iterate on the implementation without having migrations at the block level, if we need to.

While I think this is cleaner, I think there's potentially some downsides here. If you look at the commit and files, I've broken the Instagram embed out to a template that this loads, and then the embed block still has its own template, so this pattern introduces a few new little files. 

But that also gets me to something I'm not sure I've reasoned through enough: Wagtail embeds get fetched once and stored in the database (`manage.py purge_embeds` clears that table). This means that we're rendering and storing HTML from that template that just really includes the Instagram URL because we don't *fetch* anything, we just rely on Instagram's JavaScript to render it on the page in browser. This is conceptually different to me than Bluesky or Twitter embeds that are mostly text, but applies to any media in those embeds, and applies to media in YouTube embeds, etc, as well. So... I think that's fine? But worth calling out.

willbarton:
One thing I also notice that I'd like to look at is, right now the Bluesky and Instagram embeds also trip Wagtail checks in the admin:

image

I think it'd be nice to not have that happen, will see what's under our control with that.

common/blocks.py
willbarton: Moved to the `InstagramEmbedFinder` in [aa8b347](https://github.com/freedomofpress/pressfreedomtracker.us/pull/2113/commits/aa8b3476021f4b438a40774ccf2f3d466059787f)

tracker/templates/base.html
willbarton: Moved to the `InstagramEmbedFinder` template in [aa8b347](https://github.com/freedomofpress/pressfreedomtracker.us/pull/2113/commits/aa8b3476021f4b438a40774ccf2f3d466059787f)

blog/models.py
willbarton: Done in [f1e3abb](https://github.com/freedomofpress/pressfreedomtracker.us/pull/2113/commits/f1e3abb4e574b7ec4cb5843d9bd72054d26e30be)

client/common/sass/components/_embed.sass
willbarton: Done in [56ab4d5](https://github.com/freedomofpress/pressfreedomtracker.us/pull/2113/commits/56ab4d5e6e86cde02e8831cf09c60087286186dc)

common/blocks.py
willbarton: This custom embed finder is in [aa8b347](https://github.com/freedomofpress/pressfreedomtracker.us/pull/2113/commits/aa8b3476021f4b438a40774ccf2f3d466059787f)

common/models/pages.py
```diff
@@ -609,6 +611,8 @@ class SimplePage(MetadataPageMixin, Page):
         ('image', AlignedCaptionedImageBlock()),
         ('raw_html', blocks.RawHTMLBlock()),
         ('tweet', TweetEmbedBlock()),
+        ('instagram', InstagramEmbedBlock()),
+        ('bluesky', BlueskyEmbedBlock()),
```
harrislapiroff: Let's add the same grouping here that we have on other page types

common/templates/common/embeds/instagram.html
```diff
@@ -0,0 +1,6 @@
+<blockquote
+    class="instagram-media"
+    data-instgrm-permalink="{{ url }}/?utm_source=ig_embed"
+    data-instgrm-version="14">
```
harrislapiroff: Maybe there should be a fallback here? Like a button that says "view post on Instagram"? If it's easy and it works.

harrislapiroff (approved):
This looks good to me! Two small changes suggested and then please merge at will!

PR #2117, opened 2026-03-27
Adds middleware to add Onion-Location header with full path

### LINKED ISSUE #2104: Add support for returning `Onion-Location` in Wagtail
### Description

To support TOR users on our Wagtail sites after the migration to Traefik, we'll need to add functionality for Wagtail to return the `Onion-Location` header to the client.

The value of the `Onion-Location` header needs to be in this format:
```
{scheme}://{hostname}{path}
```
Where:
* `{scheme}` is `http` (as we don't use SSL connection on our wagtail onion sites)
* `{hostname}` is the onion hostname from `DJANGO_ONION_HOSTNAME`
* `{path}` is the URI of the current resource on the site

For example, the page at `https://pressfreedomtracker.us/blog/a-troubling-picture-of-press-freedom-pressures-to-start-2026/` should return this header:
```
Onion-Location: http://pftvpxbg4jdvviyqbuxw26yo66t36t4uy64fa6bkgxre3wzqm3yrpqid.onion/blog/a-troubling-picture-of-press-freedom-pressures-to-start-2026/
```

### Urgency

High (within the upcoming two-week sprint)

### Impact

Nice-to-Have

### Timing

We are aiming to go-live with the Nginx -> Traefik migration at the end of March (2025-03-31) which is when our existing Nginx controller goes EOL. We could potentially delay a week beyond that but would not want to do anymore than that.

### Notifications

@freedomofpress/fpf-infra

SaptakS:
## Description

Fixes #2104.

Adds a middleware to set Onion-Location header with full path. I have added the middleware only in production settings and not base or dev settings since I am guesseding we are using Onion-Location feature only in prod. But if someone feels strongly that this should always be present (staging and local), I am happy to move it to base.py.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

- Run docker compose with the prod-docker-compose.yaml
- Check in network tab of your browser Onion-Location is correctly set along with path.
- I have added some tests as well, ~~but for extensive testing, run ngrok (or some other tunneling, reverse proxy,etc.), open tor browser and check that the Onion-Location button shows up.~~ (The .onion button shows up only if the onion hostname is valid it seems)

### General checks


zrobert7 (approved):
This looks good but the commit is not verified

PR #2133, opened 2026-04-08
fix: update eslint to version 9, use flat config; update brace-expansion to v5

paulschreiber:
## Description
Update eslint to version 9, use flat config.

The old version of eslint (8) was not compatible with modern tooling. This meant we could not update to brace-expansion v5, which we needed to do to fix a CVE.

There is a small bit of code rewrapping. I added `.prettierrc` to minimize the change.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [X] Vulnerabilities update
- [X] Config changes


chigby:
So, I think what's going on here is that our Selenium tests are correctly flagging that at least some of the JS-based features on the site are not working on this branch. I haven't tried to coax Selenium into showing me a recording of what it's doing, but it seems like it's possible to see this on the dev version of the site yourself. When I go to http://localhost:8000/about/ I get an error message in the JS console: `Uncaught ReferenceError: __WEBPACK_DEFAULT_EXPORT__ is not defined`

image

This error seems to be short-circuiting our Javascript from loading/running. You can see this directly in the home page charts that aren't loading and the subscription form in the footer that is not submitting via an asynchronous JS request and is falling back to submitting as an HTML form (this is the feature that's being tested via Selenium that's failing to behave how we expect).

I'm not familiar with this error message, and haven't really tried to troubleshoot what's going on. Maybe it's related to some of the other package changes that happened with the eslint upgrade? Babel? I can keep poking at it if that'd be helpful, but also this information might give you what you need to go on.

redshiftzero:
Per discussion in sprint planning, @chigby suggested it may make sense to start again from scratch to identify the minimal changes to update `brace-expansion`

ryanwhitney:
per the above comment, looks like we \*can\* do minor and patch version bumps to avoid making further changes that aren't required to address the cve: https://github.com/freedomofpress/pressfreedomtracker.us/pull/2152

PR #2140, opened 2026-04-15
Adds a check to see the type of content_warning block

SaptakS:
## Description

Fixes issue with content warning text where only the first character is shown. I was actually not able to reproduce it locally, but seems like it is happening in staging: https://staging.pressfreedomtracker.us/all-incidents/alaska-state-senator-slaps-journalist/

My theory for adding that was because locally that block always returns a ListValue list of strings, but I am guessing in some cases it just returns a string, in which case maybe it just takes the first character. I added a check that it only tries to take the first element if the value is of type ListValue. This should ensure that if it is a string, it won't take the 0th element. I was not able to test it since I can't reproduce the error locally.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

### General checks

### DIFF
--- a/common/blocks.py
+++ b/common/blocks.py
@@ -14,7 +14,7 @@ import bleach
 from common.choices import BACKGROUND_COLOR_CHOICES
 from common.models.charts import (
     BubbleMapChartOptionsSchema,
-    HexbinMapChartOptionsSchema
+    HexbinMapChartOptionsSchema,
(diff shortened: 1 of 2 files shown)


harrislapiroff (approved):
This approach looks good, let's just get flake8 passing

PR #2147, opened 2026-04-23
fix: correct spelling of django in check.yaml

paulschreiber:
## Description
correct spelling of django in check.yaml

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [X] Config changes


harrislapiroff:
@enpaul This looks like a good fix to me, but wanted to flag it to you in case there's any follow up or additional work/checks that need to be done for it

harrislapiroff (approved):

PR #2156, opened 2026-04-30
Django 5.2.13 upgrade

### LINKED ISSUE #2122: Update Django from 4.2 to 5.2
### Description
Django 4.2 will be EOL at the end of April 2026:
https://www.djangoproject.com/download/

We should update to Django 5.2.

chigby:
## Description

Fixes #2122 

1. Upgrades Django to version 5.2.13
2. Applies the `django-update` script to our code
3. Upgrades most of the Django-related dependencies we're using to a version that is compatible with Django 5.

Exceptions to point 3, I did not upgrade:
- `django-recaptcha`. I did test that this still worked, but (1) we have an upper-version-boundary in the requirements.in file, and (2) it seemed very coupled to the package `wagtail-django-recaptcha`, which I'm not sure can be upgraded without a wagtail upgrade as well.
- `django-settings-export`. No newer version exists. The code seems to work, though!

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [x] Config changes

## Testing

Test basic site functionality. Might be nice to test this in production mode, too!

### Post-deployment actions

### General checks

- [ ] Verify categories work

### If it's a major change

### DIFF
--- a/tracker/settings/base.py
+++ b/tracker/settings/base.py
@@ -369,7 +369,7 @@ CONTENT_SECURITY_POLICY = {
         "default-src": [SELF],
         "form-action": [SELF],
         "frame-ancestors": [SELF],
-        "frame-src": (
+        "frame-src": [
             SELF,
             # For Social Widgets
             "https://platform.twitter.com",
@@ -379,7 +379,7 @@ CONTENT_SECURITY_POLICY = {
             "https://www.google.com/recaptcha/",
             "https://recaptcha.google.com/recaptcha/",
             "https://www.youtube.com",
-        ),
+        ],
         "img-src": [
             SELF,
             "https://analytics.freedom.press",
@@ -400,7 +400,7 @@ CONTENT_SECURITY_POLICY = {
             "DJANGO_CSP_REPORT_URI",
             "https://freedomofpress.report-uri.com/r/d/csp/enforce",
         ),
-        "script-src": (
+        "script-src": [
             SELF,
             "https://analytics.freedom.press",
             "https://www.google.com/recaptcha/",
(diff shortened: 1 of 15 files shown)


tracker/settings/base.py
```diff
@@ -337,82 +352,84 @@
...
-    CSP_IMG_SRC.extend(csp_media_origins)
-    CSP_OBJECT_SRC.extend(csp_media_origins)
-    CSP_CONNECT_SRC.extend(csp_media_origins)
-
-# Report URI must be a string, not a tuple.
-CSP_REPORT_URI = os.environ.get(
-    "DJANGO_CSP_REPORT_URI", "https://freedomofpress.report-uri.com/r/d/csp/enforce"
-)
+CONTENT_SECURITY_POLICY = {
+    "DIRECTIVES": {
```
willbarton, line 356: In these directives, is there a reason some of them are lists and some are tuples? From a glance at django-csp, it doesn't seem to care (but its docs and examples use lists), but it would be nice if they were consistent. Not a blocker.

willbarton (approved):
Looks good to me. I checked locally for any CSP issues and compared against the headers production currently returns, and I think that looks fine. 

```shell
curl -sI http://localhost:8000 | grep -i "content-security-policy"
curl -sI https://pressfreedomtracker.us | grep -i "content-security-policy"
```

The only thing I'm seeing different in CSP is that www.instagram.com, scontent.cdninstagram.com, and embed.bsky.app appear locally (and media.pressfreedomtracker.us appears in production), but I believe that's because the embed PR hasn't been deployed yet.

Otherwise, everything looks good.

I had one comment about the CSP configuration mixing lists and tuples.

tracker/settings/base.py
chigby: I think the reason lists and tuples are mixed is they all started off as tuples (not sure why, maybe a historical accident, maybe that's what the examples looked like at the time?) but then at some point some of them needed to be mutable, e.g.: 

https://github.com/freedomofpress/pressfreedomtracker.us/blob/a6e0db9db8596f2236c660eea1d86ef447154913/tracker/settings/base.py#L429-L431

It probably does make more sense to just have them all be lists!

paulschreiber: I pushed a change to ensure CSP is using lists to match the examples. Looks like I missed a couple.

PR #2159, opened 2026-05-05
fix: update Wagtail to 7.4 LTS

### LINKED ISSUE #2129: Upgrade to Wagtail 7.0 LTS

paulschreiber:
# Description
Update Wagtail 6.3.8 to 7.4.0

Also update:
- wagtail-factories 4.1 > 4.4
- wagtail-honeypot 1.2 > 1.3
- wagtail-inventory 3.1 > 3.3

Part of https://github.com/freedomofpress/fpf-www-projects/issues/508

**Notes from work done by @SaptakS** 
- Enabled deferred validation to support autosaves by adding null=True wherever makes sense. I have left out some of the orderables since they do seem risky. Adding null=True does remove the db level contraint, though Wagtail itself makes the separation such that null=True is used for deciding drafts, whereas blank=True is used to determine whether the field is actually optional.
- Sets WAGTAILSNIPPETS_MENU_SHOW_ALL to True so that all the snippets are shown in the "Snippets" menu in admin. 

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [X] Config changes

## Testing

Full QA of site.


paulschreiber:
This test is failing:
```
======================================================================
FAIL: test_serialize_general_and_category_fields (incident.tests.test_get_serialized_filters.GetSerializedFiltersTest.test_serialize_general_and_category_fields)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/django/incident/tests/test_get_serialized_filters.py", line 190, in test_serialize_general_and_category_fields
    with self.assertNumQueries(7):
AssertionError: 6 != 7 : 6 queries executed, 7 expected
```

SaptakS:
There are few different errors and failures that I am trying to fix before making it ready for review.

One of the [errors](https://github.com/freedomofpress/pressfreedomtracker.us/actions/runs/25600207765/job/75152805251?pr=2159#step:10:18409) is caused due to the [change in handling of OPTIONS](https://github.com/wagtail/wagtail/commit/66f1e817eb7d218d1ae9a43f03b2a0f37cd37d4a) request Wagtail since 6.4. Using hooks, the OPTIONS response is served separately, and actually never reaches our `RoutablePageMixin` view that implements the additional headers. The `get` and `head` methods still work like before. We can either update the test to use `head()` if we only care about testing the headers. But if we care about testing and allowing OPTIONS method, then maybe need to override that hook or something.

cc-ing @chigby since he might have some thoughts and ideas.

SaptakS:
For the [Site does not exist errors](https://github.com/freedomofpress/pressfreedomtracker.us/actions/runs/25600207765/job/75152805251?pr=2159#step:10:18406), it weirdly only happens when all the tests are run, and not when the individual test suites are run. I am trying to look into it to figure out what might be the cause since I don't really see anything in the changelog that should be causing this. Especially since it only happens when all the tests are run together and not individually.

I feel like this is happening due to cache pollution or something where the default site gets deleted in between which is why it only happens when the entire test suite is run. The error is happening in the `WagtailMetadataMixin` where it is trying to get the default site white is somehow already deleted (or site_id is changed somewhere in cache). So I am considering just creating the Site in the `setUpTestData` everytime so that the MetadataMixin can always find the default site during testing. Unless someone else has some better ideas.

willbarton (approved):

SaptakS:
@willbarton I have rebased the PR. Let me know if there is anything else that we should do before merging.

willbarton:
@SaptakS nope, I think it's good!

PR #2161, opened 2026-05-06
fix: replace psycopg2 with psycopg 3

paulschreiber:
## Description
Replace psycopg2 with psycopg[c]. 
- matches the config on FPF
- psycopg 3 is newer
- psycopg[c] is faster.

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes


willbarton (commented):
It looks like there's a [direct import from pyscopg2](https://github.com/freedomofpress/pressfreedomtracker.us/blob/develop/incident/models/incident_page.py#L43) that will need fixing.

I also wonder, is there a reason we want to build the C version each time instead of using `psycopg[binary]`?

paulschreiber:
@chigby You added the `[c]` version in ac913684946d5c77b85980fdfdf2e72ff7fa101f. Is `[binary]` ok?

paulschreiber:
@willbarton do the `Range` changes look ok?

willbarton (approved):
Yep, `Range` fix looks good, psycopg 3 [adapts to `daterange` automatically](https://www.psycopg.org/psycopg3/docs/basic/pgtypes.html#psycopg.types.range.Range).

harrislapiroff (approved):
LGTM

PR #2162, opened 2026-05-06
fix: update Python to 3.14

paulschreiber:
## Description
Update Python to 3.14
- updates Wagtail to 7.4 (and wagtail*)


paulschreiber:
Same error seen in https://github.com/freedomofpress/pressfreedomtracker.us/pull/2159

willbarton (approved):
Also looks good to me. I didn't check deprecation warnings here, because the tests are very... verbose.

PR #2166, opened 2026-05-07
fix: update Debian Bookworm to Trixie

paulschreiber:
## Description
Update Debian Bookworm to Trixie.
- Active support for Bookworm ends in June 2026.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [X] Config changes

### DIFF
--- a/devops/docker/ci-python-dockerfile
+++ b/devops/docker/ci-python-dockerfile
@@ -1,5 +1,5 @@
 # sha256 for python:3.14.5-slim-trixie on linux/amd64 as of 2026-05-20
-FROM python:3.14.5-slim-trixie@sha256:sha256:a7185a8e40af01bf891414a4df16ef10fc6000cee460a404a13da9029fe41604 AS python-base
+FROM python:3.14.5-slim-trixie@sha256:a7185a8e40af01bf891414a4df16ef10fc6000cee460a404a13da9029fe41604 AS python-base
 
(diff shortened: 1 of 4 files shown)


paulschreiber:
Same error in #2159 and #2162. Should be addressed with Wagtail upgrade.

devops/docker/ci-python-dockerfile
```diff
@@ -1,10 +1,10 @@
...
 
 RUN apt-get update &&\
-    apt-get install --no-install-recommends git -y &&\
-    apt-get clean && rm -rf /var/lib/apt/lists/* &&\
-    rm -rf /usr/share/doc/ && rm -rf /usr/share/man &&\
-    rm -rf /usr/share/locale
+apt-get install --no-install-recommends git -y &&\
+apt-get clean && rm -rf /var/lib/apt/lists/* &&\
+rm -rf /usr/share/doc/ && rm -rf /usr/share/man &&\
+rm -rf /usr/share/locale
```
willbarton: Can we back out this change to formatting? The indentation is nice for readability when these are all part of the same `RUN` instruction.

willbarton (approved):
Approving, because the trixie changes look good and work for me, but I have one request that I would like to see changed before this is merged but I don't want to block on a re-review.

PR #2175, opened 2026-05-11
Fix Wagtail 7.4 test failures: Site cache pollution and OPTIONS CORS

harrislapiroff:
## Summary
- **Clear Wagtail's site root paths cache between tests.** Many test classes delete the home page (CASCADE-removing the default Site) and create a new one in `setUpTestData`. With Wagtail 7.4, `MetadataPageMixin.get_meta_image()` now calls `self.get_site()` during template rendering, which resolves the site id via an in-process cache. The class-teardown rollback removes the new Site from the DB but does not fire the signals that would clear the cache, so subsequent tests hit `Site.DoesNotExist`. Fix: clear the cache after every test in `common/test_runner.py`.
- **Restore CORS headers on the `/incidents/export/` OPTIONS response.** Wagtail 7.4 added a `before_serve_page` hook that intercepts OPTIONS requests before they reach `RoutablePageMixin` routes, so the existing `export_view_OPTIONS` is no longer called. Override `handle_options_request` on `IncidentIndexPage` to attach the same CORS headers the test (and external clients) expect.

## Empirical verification of root cause
Confirmed the cache-pollution mechanism by isolation/composition:
- `IncidentFieldTest` alone → passes
- `TestPages` (deletes home page in `setUpTestData`) followed by `IncidentFieldTest` → fails with `Site.DoesNotExist`

## Test plan
- [x] `docker compose exec django ./manage.py test --noinput` on a fresh test DB — `Ran 713 tests in 201.817s, OK (skipped=15)`
- [x] Reverted just the OPTIONS fix and confirmed `test_OPTIONS` fails without it, then re-applied
- [x] Reviewer: run the full test suite locally to confirm

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/common/test_runner.py
+++ b/common/test_runner.py
@@ -2,7 +2,27 @@ from xmlrunner.extra.djangotestrunner import XMLTestRunner
 import factory.random
 
 from django.conf import settings
+from django.test import TestCase
 from django.test.runner import DiscoverRunner
+
+
+def _install_wagtail_site_cache_reset():
+    # Wagtail caches site root paths in process memory. Tests that delete or
+    # replace the default Site (commonly via `Page.objects.filter(slug="home").delete()`
+    # in setUpTestData) populate the cache with the new Site's id, but the
+    # transaction rollback at class teardown reverts the DB without firing the
+    # signals that would clear the cache. A later test then resolves page URLs
+    # against a Site id that no longer exists and raises Site.DoesNotExist.
+    # Clearing the cache after every test guarantees a cold cache for the next.
+    from wagtail.models import Site
+
+    original_post_teardown = TestCase._post_teardown
+
+    def post_teardown(self):
+        original_post_teardown(self)
+        Site.clear_site_root_paths_cache()
+
+    TestCase._post_teardown = post_teardown
 
 
 class WithSeedMixin(object):
@@ -17,6 +37,7 @@ class WithSeedMixin(object):
         if seed:
             factory.random.reseed_random(seed)
             print(f"Using seed: {seed}")
+        _install_wagtail_site_cache_reset()
         super().setup_test_environment()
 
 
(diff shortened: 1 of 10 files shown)


common/test_runner.py
```diff
@@ -2,9 +2,29 @@
...
+    # signals that would clear the cache. A later test then resolves page URLs
+    # against a Site id that no longer exists and raises Site.DoesNotExist.
+    # Clearing the cache after every test guarantees a cold cache for the next.
+    from wagtail.models import Site
+
+    original_post_teardown = TestCase._post_teardown
+
+    def post_teardown(self):
+        original_post_teardown(self)
+        Site.clear_site_root_paths_cache()
```
willbarton: Fascinating! Just to add a bit more context, [here is where a Wagtail page checks the Site cache](https://github.com/wagtail/wagtail/blob/e120dcddd086c24bcd6cb733caf792d1d3e964f3/wagtail/models/pages.py#L1320-L1331).

This line is probably exactly the fix if we can find a way to maybe incorporate it without monkeypatching `TestCase`?

incident/models/incident_index_page.py
```diff
@@ -148,6 +148,16 @@ def export_format_json(self, incidents, fields):
...
 
+    def handle_options_request(self, request, *args, **kwargs):
+        """
+        Override Wagtail's default OPTIONS handling to add CORS headers,
+        required for cross-origin access to the export API endpoint.
+        """
+        response = super().handle_options_request(request, *args, **kwargs)
+        response["Access-Control-Allow-Origin"] = "*"
+        response["Access-Control-Allow-Methods"] = "GET,OPTIONS,HEAD"
+        return response
```
SaptakS, line 159: This is what I definitely was thinking for overriding OPTIONS. Thanks!

SaptakS:
@willbarton @harrislapiroff @chigby I updated the code to remove monkeypatching and modify the tests to either use default_site or where it needs to recreate the homepage, add a teardown function. I feel this is better than monkeypatching. Let me know if this makes sense.

SaptakS:
As discussed in web sync, given the tests are passing, I am going to merge this PR so that the Wagtail 7.4 upgrade PR can be reviewed in it's entirety.

PR #2178, opened 2026-05-12
chore: remove manual vulnerability checks

paulschreiber:
## Description
Remove manual vulnerability checks. Avoids duplicate notifications when dependabot is enabled.
- remove safety check script
- remove safety from dependencies file
- update Makefile

This likely never worked because safety isn't a dependency in this repo/project.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [X] Config changes


harrislapiroff (approved):
Doesn't look like we've been running this at all except if manually invoked. Agree we can get rid of it with no loss of security.

PR #2218, opened 2026-06-01
Adds chart PNG in newsletter template

### LINKED ISSUE #2171: [Newsletter] Display charts as static images
It would be ideal if we could display charts as static images, using the chart pregeneration code that already exists. We'd need to render charts with axes and labels (like the download as PNG feature—though unfortunately for this use case, I believe that's generated client-side).

SaptakS:
## Description

Fixes #2171.

Uses pregenerated PNG image in the newsletter template for charts. PS: I tried testing it by sending it to my email, but remembered/realized that emails use a redirected link which won't support a localhost. So I am hoping it works in staging and then testing with an email sent from staging.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

Add a chart in a BlogPage and check it renders correctly in newsletter mode.

### General checks

image

### DIFF
--- a/blog/templates/blog/blog_page_newsletter.html
+++ b/blog/templates/blog/blog_page_newsletter.html
@@ -341,9 +341,11 @@
 							{% if block.value.title %}
 								<p><strong>{{ block.value.title }}</strong></p>
 							{% endif %}
-							<a href="{% fullpageurl page %}" class="link-external">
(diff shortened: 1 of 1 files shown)


zrobert7 (approved):
so simple so clean 🧼 let's test on staging today

PR #2219, opened 2026-06-01
fix: have django automatically format migrations with ruff

paulschreiber:
## Description
Have django automatically format migrations with ruff using `django-migrations-ruff-formatter`. Avoids manual work by developers.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [X] Config changes

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -409,6 +409,10 @@ django-filter==23.5 \
     # via
     #   -r requirements.txt
     #   wagtail
+django-migrations-ruff-formatter==0.1.3 \
+    --hash=sha256:5b81152a51e5a50dd9149e4a6bd78793319442b69cb16a9b61b3fab264cf7684 \
+    --hash=sha256:c53f9807b3d0386c92c3549057c859dd5eb21c9ac70861d92e37e108839f619c
+    # via -r requirements.txt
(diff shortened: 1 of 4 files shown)


harrislapiroff (commented):
I want to think about this a bit. I see the problem here, but I'm a little nervous about adding obscure application level packages to solve a formatting issue. Maybe we should have this package _only_ in dev settings, since we only create migrations in dev? That would at least limit its radius. Interested in thoughts from other @freedomofpress/fpf-web-developers

willbarton:
We could also just exclude generated files like migrations from formatting requirements? 

[Wagtail does this](https://github.com/wagtail/wagtail/blob/main/ruff.toml#L6), as does [Mozilla Bedrock](https://github.com/mozilla/bedrock/blob/main/pyproject.toml#L4). This would feel more correct to me.

harrislapiroff:
Thanks, @willbarton! Yes, I think that seems like the right approach. The Django migration generator is well-tested and widely-used. Let's let it do what it does.

harrislapiroff (changes requested):
Instead of this, let's add migrations to ruff's ignore config, as suggested by @willbarton

PR #2234, opened 2026-06-10
fix: update Wagtail version in CLAUDE.md, update deny paths

paulschreiber:
## Description
Update Wagtail version in CLAUDE.md

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [X] Config changes


.claude/settings.json
```diff
@@ -4,19 +4,19 @@
       "Read(**/.env)",
       "Read(**/.env.*)",
       "Read(**/local_settings.py)",
-      "Read(**/freedompress/settings/local.py)",
+      "Read(**/tracker/settings/local.py)",
```
paulschreiber, line 7: was `freedompress` a copy/paste error or is it supposed to be like that?

harrislapiroff (approved):

PR #2231, opened 2026-06-09
Prepublication/Unconfirmed incident view

chigby:
## Description

Fixes #2075 

Also includes parts of #2077 

Here's a visual of what it looks like on my machine, so far:

image

### Here's what's working:

1. Creates a stand-alone view (not a routable page) at `/unconfirmed-incidents/` that renders a table with the set of prepublication incidents given by the settings timespan value. 
2. Groups these incidents by date and location and collects category data about them.
3. Displays a functional (but maybe not pretty) data visualization in the form of a stacked bar chart comparing unconfirmed and confirmed incidents.

### What's not implemented:

1. The circled part of the design here:
image

The variable length bar with the number in it with the tooltip showing the categories represented by the number is not implemented. I have collected that data: each row has a `category_counts` field that looks something like this:

```
'category_counts': Counter(
    {
        'Assault': 5,
        'Arrest/Criminal Charge': 3,
        'Equipment Damage': 1
    })
```

But I did not have time to incorporate that into the design as shown. We do have some good tools for displaying a tooltip on hover, for rendering the category icons, and the data is there, it's a matter of putting that all into practice on the page.

2. Some smaller details on the stacked bar chart data viz

I mostly was reusing the existing bar chart, and I think it looks all right but there are some details that could be adjusted.

- Colors. The design has white / peach. The bar chart looks odd (or looked odd to me) with white as the color, so I made it #EEEEEE. But we might want to play with it. #2077 also says we might want to change the color entirely.
- Axis labels. The design has x-axis labels only on "today" and the oldest day of the chart (30 days ago). I didn't remember how to do this using the `BarChart` component. I think it's doable, though.
- The legend (confirmed/unconfirmed with the colors on the right side of the design) did not match the default way the `BarChart` component displays a legend. I'm not sure if this is needed necessarily, but it's something I noticed.

3. The FAQ page is not linked anywhere in the descriptive text. Probably not too hard to add.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires an admin update after deploy
- [ ] Includes a database migration removing or renaming a field

## Testing

[...]

PR #2263, opened 2026-06-18
Home page unconfirmed incidents table

### LINKED ISSUE #2029: Home page display for pre-publication incidents
### Description

Problem: Keeping up with "real time" documentation of incidents, especially during bursts of activity
Solution: Will need to be supported by multiple teams, but for web, focusing on the site's home page to find a way to communicate with the outside world the volume of incidents being worked on by reporting team _before_ they are all the way into the database. 
 
Harris: "more tools to display data, especially on the home page, that don’t necessarily pull from data on the site"

Possible workflow: pull from [active newsroom](https://docs.google.com/spreadsheets/d/1o-Hf3SaqfZhXo-xz2D7KHXRjCZT2fGllCi_w3lb9Ak0/edit?pli=1&gid=2121586989#gid=2121586989&fvid=2071106054) / a different spreadsheet the info needed to support language similar to "Our team is currently working to verify at least **XX press freedom aggressions** in **Los Angeles** on **9/23**. As they become verified, they will be published in the database."  

Attached: [Notes from initial 9/18/25 mtg](https://docs.google.com/document/d/1akhbmtfMXbvQUSORl9Ql6VkNUMt66Zbd6MVRAFtUMeE/edit?tab=t.0#heading=h.wk00at98m14z) 

Per TT, for Q1 2026 

### Urgency

### Impact

Important

### Timing

"Real-time documentation" is one of the Tracker's active strategic goal (thru 2026)    

### Notifications

@stephaniesugars

chigby:
## Description

Fixes #2029 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

Here's what's going on:

image

The way to test this is similar to how #2231 was tested. It's the same basic code, but it's in a different place. 

Some thoughts:

- The design mentions a scrollable table, but I found that idea a little daunting, given that the existing table from /unconfirmed-incidents/ does not have that feature, so I just made it cut off at 3 rows.
- You have to enable the home page feature in the Prepublication Settings page now. There's a checkbox or it won't display.
- The styles for the "View all..." button are a little sketchy, but I wasn't show how else to both (1) use the design, and (2) use our pre-existing button component. Maybe there's an easier way?

### Post-deployment actions


SaptakS (changes requested):
Code wise mostly looks good to me. However I noticed 2 issues in the visual design that I think needs to be addressed:

- For me in firefox, the table footer are not taking full width probably because the rest of the border is applied in the table element and here it is in the content. Looks little weird and I think should be fixed

Image


- I think in smaller screen, the table should wrap below the text. Currently it is broken

Image

SaptakS:
Also, I will note that there will be a conflict between this PR and https://github.com/freedomofpress/pressfreedomtracker.us/pull/2270 since both makes changes to the table part of the template, so whichever gets merged first, the other PR will probably need some rebasing and reworking

chigby:
1. I can rebase this branch on your PR's branch
2. I added some basic responsive code to put the table below the text when on the home page on mobile. I also adjusted the column widths on the home page version of the table, since it was seeming a little cramped.
3. I played around with the button in the table footer's border a little bit to make it borrow the outer border from the table and I think it's looking okay? I don't love manipulating the base site components a ton, but I'm not sure how else to get it to work. If you have other ideas about how to make the button behave well while in the footer of the table, I'd love to hear them.

SaptakS (approved):
Looks good to me.

PR #2206, opened 2026-05-27
Increase granularity on homepage visualizations 

ryanwhitney:
## Description

Fixes #2030.

Changes proposed in this pull request:

- Add 4 and 12 week options for viewing homepage charts
  - Add new weekly view to bar charts to support this 
  - 4 week becomes the default view
- Add 7 day viewing option
  - Does not show unless the chart container is rendered with a `sevenDayEnabled` attr
  - Also does not show unless there's data within the last 7 days 
  - When enabled, this becomes the default.
  - Requires teensy followup backend/wagtail work to add an admin setting to enable. (Could also do it here, but I figured I'd do it on top of @chigby's work.)
- One later [commit](4aaacc9be6921092d0fd11e68b041e726cabc345) updates the `filtersApplied` state object on HomepageMainCharts to add timePreset and a corresponding enum. Previously there were booleans for each time option available and we had to manually ensure only one was true at a time. Using a single value makes more sense to me, but it's a bigger diff. Easy to roll back if we think it's not necessary.
- Small fix: the year dropdown no longer resets when switching between time periods.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires an admin update after deploy
- [ ] Includes a database migration removing or renaming a field

## Testing

> [!TIP]
>  I wrote this for my prior chart PR, but I think it should still be valid: If you want to test with real data, put these two files from another branch on top of this PR. You can easily futz with dates in the csv to make more recent data.  `git checkout origin/backup-pre-drop -- client/charts/js/bubble-map-chart.js client/charts/js/data/incidents.csv.js`

- [ ] Ensure all new date options show data as expected
- [ ] Test at various breakpoints: slimmer desktop views + mobile view. 
- [ ] Ensure all charts link to the expected ranges in the incident database
- [ ] The 7 day stuff isn't relevant and won't ever appear until there's an admin switch. You can manually enable it and test it if you'd like though! 

### Post-deployment actions

In case this PR needs any admin changes or run a management command after deployment, mention it here:

### General checks

- [ ] Linting and tests pass locally
- [ ] The website and the changes are functional in Tor Browser
- [ ] There is no conflicting migrations
- [ ] Any CSP related changes required has been updated (check at least both firefox & chrome)
- [ ] The changes are accessible using keyboard and screenreader

### If you made changes to API flow:

[...]

PR #2270, opened 2026-06-22
Viz mobile improvements

### LINKED ISSUE #2077: Add visualization for Unconfirmed Incidents page
### Description

Implement this viz for the unconfirmed/confirmed incidents on the unconfirmed incidents page (#2075):

Image

Important: We like the peach color for confirmed incidents but want to modify the unconfirmed incident color to grey/peach, which we should run past editorial. 

### Urgency

### Impact

Important

### Timing

_No response_

### Notifications

_No response_

### LINKED ISSUE #2239: Implement mobile design for unconfirmed incidents page
Currently, the page looks something like this in the mobile screens:

Image

It's not "broken" but it's not great either. We should probably think a little bit on how to update the table for smaller screens.

SaptakS:
## Description

Fixes #2077.
Fixes #2239 

### Changes proposed in this pull request:
Things needed to be done from https://github.com/freedomofpress/pressfreedomtracker.us/issues/2077#issuecomment-4732646067
#### Bar chart changes
- [x] Shrink height and width (max 670w) on page 
- [ ] ~~Add legend/key with corresponding colors [per initial design](https://www.figma.com/design/agOk78n9QpCcNCjqiZ6ObO/Press-Freedom-Tracker?node-id=309-10371&t=yKPnoCBz2uXAzDEj-4) (needed as we haven't used stacked bars previously)~~ (_To be addressed through https://github.com/freedomofpress/pressfreedomtracker.us/issues/2276_)
- [x] Remove filter buttons
- [x] Mobile: Switch to a slider and reduce bar width, padding, and margins as it shrinks to get to the _widths_ [seen in the prior comment](https://github.com/freedomofpress/pressfreedomtracker.us/issues/2077#issuecomment-4711246802)
- [ ] ~~Adjust tooltips per above guidance (one tooltip per entire bar, entire bar selectable)~~ (_To be addressed through https://github.com/freedomofpress/pressfreedomtracker.us/issues/2276_)
- [ ] ~~Remove <title> attributes as they compete with tooltips. (I believe they already have `aria-label` as I made a similar change in #2206)~~ (_To be addressed through https://github.com/freedomofpress/pressfreedomtracker.us/issues/2276_)
- [x] **TBD:** @ryanwhitney needs to provide guidance on visual hover states. Initial guidance is: since we have stacked bars, we don't want to white out all other bars when hovering on one as that obscures useful data. Instead we'll need to rely on something like a different color/width outline (or other type of indication) on the hovered bar while leaving others alone. 

#### Table changes
- [x] Shrink width on page: max 670 per other pft elements
- [x] Mobile: Reduce padding in (and around) unconfirmed incidents table per above guidance
- [x] Mobile: Have the bars adjust sizing at smaller sizes to ensure fit at 320px viewport
- [x] Number not vertically aligned in incident count bar
- [x] Table missing row coloring: In the design assets, even rows get a #F8F8F8 background. (`$section-background` in pft)
- [x] [Tooltips can go offscreen](https://github.com/user-attachments/assets/24cd7c44-1cb7-4cf6-88a7-378cdb63d315): We solved this elsewhere, should be able to reuse that

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

### Post-deployment actions

### General checks

image

### DIFF
--- a/client/charts/js/components/BarChart.jsx
+++ b/client/charts/js/components/BarChart.jsx
@@ -82,6 +82,7 @@ export default function BarChart({
 	setSvgEl = () => {},
 	interactive = true,
 	disableAnimation = false,
+	showCategoryButtons = true,
 }) {
 	if (!data.length) return null
 	const dataset = data.map((d, i) => ({ ...d, index: i }))
@@ -119,7 +120,7 @@ export default function BarChart({
 	)
 
 	const categoryButtonsLabels = calculateCategoriesLabelsLegend(datasetStackedByCategory, paddings, width)
-	const buttonsHeight = isMobileView || !allCategories
+	const buttonsHeight = isMobileView || !allCategories || !showCategoryButtons
 		? 0
 		: calculateButtonsHeight(categoryButtonsLabels)
 
@@ -218,7 +219,7 @@ export default function BarChart({
 					viewBox={[0, 0, width, height]}
 				>
 					{description ? (<desc>{description}</desc>) : null}
-					{allCategories && (
+					{allCategories && showCategoryButtons && (
 						<CategoryButtons
 							interactive={interactive}
 							categoryButtonsLabels={categoryButtonsLabels}
@@ -391,15 +392,15 @@ export default function BarChart({
 				{
 					// If there are multiple categories, ie a stacked bar chart,
 					// we show the category buttons
-					allCategories && (
+					allCategories && showCategoryButtons && (
 						<CategoryButtons
 							interactive={interactive}
 							datasetStackedByCategory={datasetStackedByCategory}
+							categoryButtonsLabels={categoryButtonsLabels}
 							paddings={paddings}
 							width={width}
 							hoveredElement={!hoveredElement?.x && hoveredElement?.y}
 							setHoveredElement={(el) => setHoveredElement(el ? { y: el } : el)}
-							setButtonsHeight={setButtonsHeight}
 							findColor={findColor}
 							textStyle={textStyle}
 						/>
--- a/client/charts/js/components/BarChart.jsx
+++ b/client/charts/js/components/BarChart.jsx
@@ -30,6 +30,7 @@ const paddingsInternal = {
 }
 
 const borders = {
+	thin: 3,
 	normal: 5,
 	hover: 7,
 	grid: 1,
@@ -168,6 +169,7 @@ export default function BarChart({
 		const isHoveredByKey = !hoveredElement?.x && hoveredElement?.y === key
 		return isHoveredByDate || isHoveredByKey ? barColor : !hoveredElement ? barColor : defaultColor
 	}
+	const mobileThinBorderThreshold = 24
 
 	const selectedElement = dataset.find((d) => d[x] === sliderSelection)
 	const incidentsCount = selectedElement !== undefined ? selectedElement[y] : 0
@@ -394,10 +396,7 @@ export default function BarChart({
 					allCategories && (
 						<CategoryButtons
 							interactive={interactive}
-							datasetStackedByCategory={datasetStackedByCategory}
 							categoryButtonsLabels={categoryButtonsLabels}
-							paddings={paddings}
-							width={width}
 							hoveredElement={!hoveredElement?.x && hoveredElement?.y}
 							setHoveredElement={(el) => setHoveredElement(el ? { y: el } : el)}
 							findColor={findColor}
@@ -476,7 +475,7 @@ export default function BarChart({
 										: sliderSelection === null
 											? getBarColor(branchBars.key, d.data, 'white')
 											: 'white',
-								strokeWidth: borders.normal,
+								strokeWidth: dataset.length > mobileThinBorderThreshold ? borders.thin : borders.normal,
 								stroke: (d) => (sliderSelection === d[x] ? '#E07A5F' : 'black'),
 								cursor: searchPageURL ? 'pointer' : 'inherit',
 								shapeRendering: 'crispEdges',
--- a/client/charts/js/components/BarChart.jsx
+++ b/client/charts/js/components/BarChart.jsx
@@ -82,7 +82,6 @@ export default function BarChart({
 	setSvgEl = () => {},
 	interactive = true,
 	disableAnimation = false,
-	showCategoryButtons = true,
 }) {
 	if (!data.length) return null
 	const dataset = data.map((d, i) => ({ ...d, index: i }))
@@ -120,7 +119,7 @@ export default function BarChart({
 	)
 
 	const categoryButtonsLabels = calculateCategoriesLabelsLegend(datasetStackedByCategory, paddings, width)
-	const buttonsHeight = isMobileView || !allCategories || !showCategoryButtons
+	const buttonsHeight = !allCategories
 		? 0
 		: calculateButtonsHeight(categoryButtonsLabels)
 
@@ -219,7 +218,7 @@ export default function BarChart({
 					viewBox={[0, 0, width, height]}
 				>
 					{description ? (<desc>{description}</desc>) : null}
-					{allCategories && showCategoryButtons && (
+					{allCategories && (
 						<CategoryButtons
 							interactive={interactive}
 							categoryButtonsLabels={categoryButtonsLabels}
@@ -392,7 +391,7 @@ export default function BarChart({
 				{
 					// If there are multiple categories, ie a stacked bar chart,
 					// we show the category buttons
-					allCategories && showCategoryButtons && (
+					allCategories && (
 						<CategoryButtons
 							interactive={interactive}
 							datasetStackedByCategory={datasetStackedByCategory}
--- a/client/common/sass/components/_unconfirmed-incidents.sass
+++ b/client/common/sass/components/_unconfirmed-incidents.sass
@@ -1,8 +1,11 @@
 .unconfirmed-incidents
 	table-layout: fixed
 	border-collapse: collapse
-	width: 90%
-	margin: 10px auto
+	max-width: 670px
+	margin: 10px 8px
+
+	+desktop-up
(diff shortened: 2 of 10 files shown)


harrislapiroff:
@SaptakS and I chatted a bit about the gray color not looking sufficiently distinct to seem highlighted when using the slider. I realize this would require adding functionality, but I suggested using the orange color but with a diagonal stripes pattern. Just a thought, not a completion criteria :)

(I've used https://riccardoscalco.it/textures/ in the past to generate patterns that are easy to use in SVG and with D3)

SaptakS:
@harrislapiroff I think the issue is less with complexity of getting the pattern but integrating it with AnimatedDataset, Categories Button and also with chart pregeneration. I actually tried adding with Claude and the changes are quite significant and they flagged that it might work correctly with chart pregeneration

SaptakS:
I think this PR is mostly ready for review apart from the TBD by @ryanwhitney about the hover state.

ryanwhitney:
Looking at colors now, but also noticed at 320px we're losing all color in the bars. I think we can get just enough width to show color by creating mobile values for `paddingsInternal` -- maybe 0 left; 15–20ish right.

Screenshot 2026-06-24 at 6 38 05 AM

ryanwhitney:
@SaptakS for hover state, let's just go with a different color. I was initially thinking a darker gray but I think `#F4C280` works, already exists on the site, and appears different enough across quick sight tests.

(Longer term, I think the initial guidance of not whiting-out all other bars on mobile still stands, but for the sake of consistency, color is ok for now.)

client/common/sass/components/_unconfirmed-incidents.sass
```diff
@@ -15,41 +15,61 @@
...
 		color: $white
 		border: 3px solid
 		border-color: $black
-		font-size: 1.125rem
+		font-size: 1.075rem
 		font-weight: 700
 		line-height: 1
 		// width of one cell * incident count
-		width: calc(var(--incident-count) * 1.6rem)
+		width: calc(var(--incident-count) * var(--count-unit-width))
```
ryanwhitney: I still think we can make better use of this space by replacing the fixed unit width with something based on the dataset min max which can be passed in like --incident count:

```css
width: calc(var(--incident-count) / var(--max-count) * 100%);
min-width: max-content;
 ```

^^ this would scale each proportionally. It has `max-content` as a fallback to avoid the number ever being cut off, though the visual comparison becomes wrong if that ever happens. It also means a 1 or 2 will always fully fill the space, which seems a bit much to me visually, though not terrible. 

A more complicated version of it that solves the latter is setting min and maxes for how much width a unit of 1 is represented by and clamping between them:


```css
--min-unit: 1.6rem; // same as we already do
--max-unit: 25%; // 2 incidents will always max out at taking 50% of the space
width: calc(var(--incident-count) * clamp(var(--min-unit), calc(100% / var(--max-count)), var(--max-unit)));
min-width: max-content;
 ```

Comment (SaptakS) [reply client/common/sass/components/_unconfirmed-incidents.sass]: I like the clamp idea. Let me give that a try.

**Comment by SaptakS** [comment]
I have addressed all the checklists and the discussions I had with @ryanwhitney on some of the design decisions. Also opened a separate ticket for overall changes we might need to make to stacked charts which might need separate discussions.

This is how it looks currently:
Desktop:
image

Mobile:
image

File: client/common/sass/components/_unconfirmed-incidents.sass [hunk 2]
```diff
@@ -15,41 +16,63 @@
...
+		+mobile-down
+			padding: 0.6em
 
 	&--count
+		--min-unit: 1.6rem
+		--max-unit: 25%
 		box-sizing: border-box
-		display: inline-block
+		display: inline-flex
 		padding: 3px
```
ryanwhitney: should be able to vertically center without padding since its inline flex?

```suggestion
		align-items:center
		padding: 0 3px
```

client/common/sass/components/_unconfirmed-incidents.sass
```diff
@@ -15,41 +16,63 @@
...
 		// width of one cell * incident count
-		width: calc(var(--incident-count) * 1.6rem)
+		width: calc(var(--incident-count) * clamp(var(--min-unit), calc(100% / var(--max-count)), var(--max-unit)))
+		min-width: max-content
 		height: 1.6rem
 		cursor: pointer
 
+		+mobile-down
+			--min-unit: 1rem
+			padding: 3px 0
```
ryanwhitney:
```suggestion
			padding: 0
```

matching the above vertical padding change

client/common/sass/components/_unconfirmed-incidents.sass
```diff
@@ -15,41 +16,63 @@
...
 	th, td
 		padding: 0.6em
+		+mobile-down
+			padding: 0.4em 0.3em
 
 	tr > :nth-child(1)
-		width: 10%
+		width: 16%
+		+mobile-down
+			width: 20%
```
ryanwhitney:
```suggestion
			width: 24%
			padding-right: 0
```

client/common/sass/components/_unconfirmed-incidents.sass
```diff
@@ -15,41 +16,63 @@
...
 
 	tr > :nth-child(2)
-		width: 15%
+		width: 25%
+		+mobile-down
+			width: 33%
```
ryanwhitney:
```suggestion
			width: 36%
			padding-right: 0
```

client/common/sass/components/_unconfirmed-incidents.sass
```diff
@@ -15,41 +16,63 @@
 
 	th, td
 		padding: 0.6em
+		+mobile-down
+			padding: 0.4em 0.3em
```
ryanwhitney:
```suggestion
```

client/common/sass/components/_unconfirmed-incidents.sass
```diff
@@ -15,41 +16,63 @@
...
-		font-size: 1.125rem
+		font-size: 1.075rem
 		font-weight: 700
 		line-height: 1
 		// width of one cell * incident count
-		width: calc(var(--incident-count) * 1.6rem)
+		width: calc(var(--incident-count) * clamp(var(--min-unit), calc(100% / var(--max-count)), var(--max-unit)))
+		min-width: max-content
 		height: 1.6rem
 		cursor: pointer
```
ryanwhitney, line 79: can make the number a bit lighter since we're light text on dark bg

```suggestion
		cursor: pointer
		-webkit-font-smoothing: antialiased
		-moz-osx-font-smoothing: grayscale
```

client/charts/js/components/BarChart.jsx
```diff
@@ -395,11 +395,11 @@ export default function BarChart({
 						<CategoryButtons
 							interactive={interactive}
 							datasetStackedByCategory={datasetStackedByCategory}
+							categoryButtonsLabels={categoryButtonsLabels}
 							paddings={paddings}
 							width={width}
```
ryanwhitney:
```suggestion
							categoryButtonsLabels={categoryButtonsLabels}
```

Can get rid of these props since they're now calculated separately in categoryButtonsLabels and CategoryButtons doesn't use them?

ryanwhitney (approved):
Looks good to me!  Mostly small sizing changes requested. 

I think the only one that definitely needs doing is switching to `align-items` instead of vertical padding. The rest I leave to your judgement. 

The only other thing previously mentioned was the color next to the data in bar chart tooltips, but that's also a global thing for these so I left a comment in https://github.com/freedomofpress/pressfreedomtracker.us/issues/2276#issuecomment-4802561001

Thank you for handling so many things here. 🏆

ryanwhitney:
One more change request. At 320px the bars are all border with no space for the inner area/color. Could we: 

add a `thin: 3,` to `borders` on BarChart

Add a new const `const mobileThinBorderThreshold = 24` 

and then on Line 483, do: 
```jsx 
strokeWidth: dataset.length > mobileThinBorderThreshold ? borders.thin : borders.normal,
```

(Doesn't have to be done exactly like this if you see a better way, just want to make sure the inside is visible down to 320 at least.)

client/common/sass/components/_unconfirmed-incidents.sass
SaptakS: I agree logically this makes sense and I actually tried this when implementing but sadly, this logic doesn't work because the font has some spacing for the descender alphabets. This is how it looks in my firefox with the changes:

image

That's why I did some trial-and-error with padding till it looked mostly centered. I tried adjust line-height too but doesn't work with this font.

SaptakS:
@ryanwhitney I have addressed most of your comments apart from the vertical alignment suggestion. Left a [comment](https://github.com/freedomofpress/pressfreedomtracker.us/pull/2270#discussion_r3481184582) regarding the same

client/common/sass/components/_unconfirmed-incidents.sass
SaptakS: Same reason as above

SaptakS:
After discussion with @ryanwhitney we found that my alignment and his alignment was happening because our OS browsers were using different fonts. For now, I have implemented Ryan's solution, since that works correctly with Helvetica Neue which is our recommended font. But I have opened a [separate issue](https://github.com/freedomofpress/pressfreedomtracker.us/issues/2285) to maybe discuss more about having same font across all OS, or atleast similar fonts in CSS preference that behave similarly.

SaptakS:
Seems like this has conflicts again after the other PR got mergeed :(, so need to rebase conflicts again.

chigby:
I'll fix the conflicts

SaptakS:
Thanks @chigby ! I will go ahead and put it on merge queue then.

PR #2279, opened 2026-06-24
Support unconfirmed incidents with imprecise dates

### LINKED ISSUE #2271: [Unconfirmed Incidents] Handle imprecise dates
Sometimes incidents in the editorial spreadsheet will have only a month and year.

We should handle those cases—similar to how we already handle IncidentPages without exact dates.

It's OK to exclude such incidents from charts.

chigby:
## Description

Fixes #2271 

- Support for dates in the Google Sheet with the format `MM/YYYY`
- Rows with this date format are given a `date_precision` value of `MONTH`
- They _will_ be included in the counts of unconfirmed incidents (using similar logic to how this is is done for confirmed incidents).
- They _will not_ be included in the bar chart or the table, both of which break down the data by exact day.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

Verify that behavior described above is correct. I've written tests to exercise these situations (mostly through adding imprecisely-dated data and ensuring the counts were the same as before that concept was introduced), but check over it to make sure it seems correct to you.


SaptakS (approved):
Looks good to me logically.

PR #2299, opened 2026-06-29
Fix newsletter pregenerated chart rendering (#2297)

### LINKED ISSUE #2297: [Newsletter] Preegenerated chart rendering incorrectly
### Description

Noticing two bugs in the chart preview when sending a newsletter:

1. Image URLs are generated incorrectly, e.g., `https://pressfreedomtracker.ushttps//media.pressfreedomtracker.us/media/images/arrestcri-2026-05-01-nj-categories-type-ur-mont.original.png` (Note the extra `https://pressfreedomtracker.us` at the start and the missing colon from the following `https:`
2. Even when I correct the image URL, it appears to be blank https://media.pressfreedomtracker.us/media/images/arrestcri-2026-05-01-nj-categories-type-ur-mont.original.png

Image missing from newsletter:

Image

Chart _does_ appear fine on web:

Image

### Expected Behavior

Chart should render on both web and newsletter.

### Browser and OS

_No response_

### Urgency

Critical (ASAP)

### Notifications

Kirstin

harrislapiroff:
Fixes #2297. (Supersedes #2298, which closed automatically when this branch was renamed.)

Two bugs caused pregenerated charts to render incorrectly when sending a newsletter.

## 1. Malformed image URL

The newsletter template built the chart `src` as `{{ page.get_base_url }}{{ block.value.png_snapshot_url }}`. In production, media is served from a custom GCS endpoint, so `png_snapshot_url()` (`get_rendition("original").url`) already returns an **absolute** URL. Prepending the site root produced a doubled host:

```
https://pressfreedomtracker.ushttps//media.pressfreedomtracker.us/media/images/....png
```

(the inner `://` is then mangled by the MJML compiler choking on the double-protocol string).

**Fix:** `png_snapshot_url()` now returns the rendition's `full_url`, which is absolute on its own and is a no-op when the URL is already absolute (it only prepends `WAGTAILADMIN_BASE_URL` for relative/local URLs). This matches how the newsletter template already renders its other images (`{{ lead_img.full_url }}`, `{{ block_img.full_url }}`). The template no longer prepends `get_base_url`.

## 2. Blank chart

Even with a corrected URL the image was blank. The pregenerator filters incidents by state **client-side** (`filterDatasets` → `filterStates.has(state)`), but the bar and treemap charts fetched incidents with `fields=tags,date,categories` — **no `state` field**. With no `state` column every row failed the state match, the dataset filtered to empty, and the chart rendered blank. (The web view works because it sources state-aware data, so it never drops the rows.)

**Fix:** request the `state` field for the bar and treemap data fetches. (Bubble/hexbin maps use `homepage_csv`, which already includes state.)

Verified locally: a state-filtered bar chart PNG went from blank to correctly rendered.

## Tests
- `chart_pregenerator`: new `renders state-filtered Bar Chart` snapshot test; the mocked API now honors the `fields` param for the `state` column, so the test renders the empty fallback (and fails) without the fix.
- `common/tests/test_chart_value.py`: `test_png_snapshot_url` now asserts the returned URL is absolute (`full_url`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


SaptakS (approved):
Fixed the breaking tests. This actually looks good to me for merging. Don't see any breaking change, and the tests definitely show that absolute url is being rendered.

PR #2349, opened 2026-07-05
Allow `prepub_sync` command to skip rows

### LINKED ISSUE #2348: Make `prepub_sync` more forgiving of errors
# Summary
Currently, the `prepub_sync` command will abort and not update any data on the site if any of the incoming data is found to be invalid. This can be problematic if some of the rows in the Editorial Tracking spreadsheet are incomplete, which is a valid state for them to be in. But we don't want to hold up other rows from being presented on the Unconfirmed Incidents page if they have the required data fields present and valid.

This issue is for allowing the sync command to skip rows while continuing to sync other rows that are valid which will be shown on the dashboard where the validation errors are currently being shown. 

# Some notes

Currently, as of this writing, the sync command fails due to 3 problematic rows:

- A row with a blank date field
- A row with a location of `Washington , DC` -- this is a whitespace error
- A row with a location of `Brussels, Belgium`, which is not in our geolocation database (or, perhaps it needs not be said, even in the United States)

Ideally, the resolution of this issue would skip the rows with blank or entirely invalid data, and also strip whitespace before doing location lookups.

chigby:
## Description

Fixes #2348

1. Replaces the "abort" logic of the `prepub_sync` command with "skip rows" logic. This means we track different, more specific data about the sync each time it runs, and also report that data on the admin home page dashboard.
2. Make the handling of the location data sync more forgiving of whitespace errors.

There's slightly more detail about the specific changes made in the commit messages.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing
It is actually slightly hard to test the row-skipping code using the live data on [the Google Sheet](https://docs.google.com/spreadsheets/d/1PeMPpol5d0MrF0KH36ZviN7Z4PipK6ZeSDh9AlJ3-eA/edit?gid=2121586989#gid=2121586989&fvid=2071106054), as there has been a structural change to the document that causes the sync command to fail without syncing any rows (namely, the "Status" column has had its name changed--this situation is, I suspect, an accidental change with the sheet, but it's out of scope for this issue in my opinion). So, I recommend reading over the code and the changes to the tests to make sure you understand what it's supposed to be doing. To test out the changes to the dashboard, it might be necessary to manually update the models (in, say, `manage.py shell`) to create a few `PrepublicationSyncSkippedRow` objects and play with the UI.

### Post-deployment actions

Eventually, this command will have to be added to a cron job on staging/production, but that's also probably out of scope for this specific issue.

### General checks

Here's the unconfirmed incident sync section of the admin dashboard: 

image


SaptakS (approved):
I checked the code and makes sense to me.

PR #2415, opened 2026-07-15
fix: avoid exceptions on 403 errors

### LINKED ISSUE #2326: Some 403s getting logged (and alerted) as 500s
### Description

Some 403s appear to be getting logged to Kibana as 500s. We believe this is happening on the Django application side.

This was discovered because a user was triggering 403s by

1. Actively having a wagtail editing window open
2. Being logged out of Wagtail

When you have a Wagtail editor open, Wagtail sends a periodic ping to keep track of what pages are actively being edited. If you're not logged in when you send this ping, you get a `PermissionDenied` exception, which raises an `Http403` ultimately. That is all correct behavior.

Unfortunately this caused 500 storms in our logging! Two going theories as to why:

1. Our logging middleware is logging 500s before raising the correct response: https://stackoverflow.com/questions/74220458/django-permissiondenied-return-error-500-instead-of-403
2. Django is looking for a 403.html template to render, not finding one, and raising a 500 error for a missing template

(Personally I think the first theory is more likely or we'd see a traceback about a missing template instead.)

Investigation required by the web team.

### Expected Behavior

403s should not be logged as 500s

### Browser and OS

_No response_

### Urgency

### Notifications

_No response_

paulschreiber:
## Description
Avoid exceptions on 403s

Fixes https://github.com/freedomofpress/pressfreedomtracker.us/issues/2326

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

### DIFF
--- a/common/middleware/request_logger.py
+++ b/common/middleware/request_logger.py
@@ -1,6 +1,7 @@
 import time
 import uuid
 
+from django.core.exceptions import PermissionDenied
 from django.http import Http404
 import structlog
 
@@ -26,7 +27,7 @@ class RequestLogMiddleware(object):
         self.response_keys = ("charset", "reason_phrase", "status_code")
 
     def process_exception(self, request, exception):
-        if isinstance(exception, (Http404, PermissionDenied)):
+        if not isinstance(exception, (Http404, PermissionDenied)):
             logger.exception("request_failed")
 
     def __call__(self, request):
--- a/common/middleware/request_logger.py
+++ b/common/middleware/request_logger.py
@@ -26,7 +26,7 @@ class RequestLogMiddleware(object):
         self.response_keys = ("charset", "reason_phrase", "status_code")
 
     def process_exception(self, request, exception):
-        if not isinstance(exception, Http404):
+        if isinstance(exception, (Http404, PermissionDenied)):
             logger.exception("request_failed")
 
     def __call__(self, request):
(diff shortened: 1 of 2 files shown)


chigby (approved):

harrislapiroff:
@chigby I should have weighed in on this before merge, but, two questions:

1. Is there any way we can capture all 4XX errors instead of manually adding them to this list as they arise?
2. Should we implement this across all three projects?

chigby:
It is definitely the case that there are some other exceptions that Django seems to use to generate a 4xx (aka Bad Request) status response. Seems like perusing [this file](https://github.com/django/django/blob/stable/5.2.x/django/core/handlers/exception.py#L63) could help there. So ones that we are not listing in our own handler include:

* `BadRequest`
* `RequestDataTooBig`
* `SuspiciousOperation`
* `TooManyFieldsSent`
* `TooManyFilesSent`
* `MultiPartParserError`

So, we could add these to the list of things not to include in our own "unhandled exception" logging. But I might a little hesitant to do that, especially if these aren't occurring in the wild, as it feels a little bit like a code smell to me. And so I wonder if we're doing something else wrong and possibly our overall logging middleware could work more harmoniously with Django's internal response processing. I don't feel like I know enough about the whole situation to feel like I have decisive answer to what we ought to be doing here.

And two answer point two: yes, we should probably implement the `PermissionDenied` change to all 3 projects.

PR #2480, opened 2026-07-17
ci: don't pull an image just for curl

paulschreiber:
## Description
Same as https://github.com/freedomofpress/freedom.press/pull/3089, but for PFT

We were pulling an external image simply to run `curl` against the django service. We already have curl inside the django container, so let's just use it.

<details>
<summary>
example CI failure logs
</summary>

```
Run docker run --pull=missing --rm --network=freedompress_app "docker.io/curlimages/curl:latest" \
  docker run --pull=missing --rm --network=freedompress_app "docker.io/curlimages/curl:latest" \
  --ipv4 --retry 24 --retry-delay 5 --retry-all-errors http://django:8000/health/ok/
  shell: /usr/bin/bash -e {0}
  env:
    VERSION_DOCKER: latest
    COMPOSE_FILE_DEV: docker-compose.yaml
    COMPOSE_FILE_PROD: prod-docker-compose.yaml
Unable to find image 'curlimages/curl:latest' locally
latest: Pulling from curlimages/curl
8015b0f4edbc: Pulling fs layer
8015b0f4edbc: Download complete
8015b0f4edbc: Pull complete
Digest: sha256:7c12af72ceb38b7432ab85e1a265cff6ae58e06f95539d539b654f2cfa64bb13
Status: Downloaded newer image for curlimages/curl:latest
curl: (6) Could not resolve host: django
Warning: Problem : timeout. Retrying in 5 seconds. 24 retries left.
```
</details>

A few things in [that failed run](https://github.com/freedomofpress/freedom.press/actions/runs/29507956189/job/87658450175) stand out immediately:

1. We're pulling a container image from a remote registry in order to run `curl`, when our images already have it.
2. We're communicating cross-network rather than intra-network to poll the healthcheck endpoint.
3. We're manually running a bespoke command to poll for app readiness, rather than using a first-class healthcheck block via compose.

The solution in this PR addresses all of those points, and should prove more stable on future CI runs.

## Scope of changes

This PR only affects CI behavior, in order to reduce failure rate, so no concerns for live deployments.

## Additional fixes
* Build:Prod:Chartgen correctly uses NodeChartPregeneratorDockerfile
* Remove redundant apt-get clean commands
* update webpack-bundle-tracker to 3.2.4 to match other repos and update STATS_FILE path to match


chigby (approved):
Looks good to me!

tracker/settings/base.py
```diff
@@ -333,6 +333,7 @@
     "DEFAULT": {
         "CACHE": not DEBUG,
         "BUNDLE_DIR_NAME": "bundles/",  # must end with slash
+        "STATS_FILE": os.path.join(BASE_DIR, "build/static/bundles/webpack-stats.json"),
```
chigby, line 336: Actually, is there a need to put the stats file in the build directory?

package.json
```diff
@@ -47,7 +47,7 @@
     "stylelint-config-standard": "^33.0.0",
     "watchpack": "^2.5.2",
     "webpack": "^5.74.0",
-    "webpack-bundle-tracker": "^1.6.0",
+    "webpack-bundle-tracker": "^3.2.4",
```
conorsch, line 50: Moving from v1 to v3 is significant. I don't object to this, but it does feel out of scope for the current PR.

tracker/settings/base.py
conorsch: Presumably the v1 -> v3 bump for the webpack package has a different syntax and is therefore required. But you're right to flag it as surprising.

docker-compose.yaml
```diff
@@ -52,7 +52,7 @@ services:
     working_dir: /django
     user: ${UID:?err}
     healthcheck:
-      test: jq '.status' webpack-stats.json |grep -q -e compiling -e error -e done
+      test: jq '.status' build/static/bundles/webpack-stats.json |grep -q -e compiling -e error -e done
```
conorsch, line 55: The healthcheck on the node container isn't really doing anything: it's checking whether the webpack builder stats file reports "compiling" _or_ "error" _or_ "done" and is satisfied if any of those are true. That doesn't sound like a healthcheck to me. 

We might as well just `test -f webpack-stats.json` since that's functionally equivalent. Or, better, I'd say just remove the healthcheck on the node container entirely, and let the django container depend on the node container having started, not being "healthy."

conorsch (commented):
Overall looks good. I flagged some improvements on the health-check logic in-line.

package.json
paulschreiber: When I upgraded SDO to 3.2.4, I needed to update the path. https://github.com/freedomofpress/securedrop.org/pull/1404

I updated PFT to make the site config more similar.

docker-compose.yaml
paulschreiber: Here I copied what was already in FPF. This was from a couple years ago: https://github.com/freedomofpress/freedom.press/commit/c1c31f1fe0614e5afa55289ba13c6aa7edc8953a

@chigby wrote it and can better explain.

chigby: I agree it doesn't really feel like a health check! And I am remembering that I added that code to address a race condition in the dev-env where Django would start up and then crash because either (a) that file didn't exist, or (b) it wasn't valid JSON. So the "healthcheck" as it exists is more like a shim to prevent that from happening. No idea if that's still an issue in v3 of `django-webpack-loader`, or if there's a better way to encode what's going on here in the language of docker compose.

PR #2630, opened 2026-08-06
Align chart on /unconfirmed-incidents to chosen timespan

### LINKED ISSUE #2550: [Prepub] Update chart on /unconfirmed-incidents to match chosen date range
### Description

- [ ] if the threshold is set to only show unconfirmed incidents from the past week, the chart on /unconfirmed-incidents should do the same.

Per discussion with @KAMcCudden, it's OK to max out at 30 days, in the case they ever end up having a longer range set.

### Urgency

### Impact

Important

### Timing

_No response_

### Notifications

_No response_

ryanwhitney:
## Description

make the chart on /unconfirmed-incidents match the chosen prepub timespan used for other displays. Capped at 30 days. 

Fixes #2550.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

### Post-deployment actions

### General checks

- [ ] Verify categories work

### If it's a major change

### DIFF
--- a/incident/views.py
+++ b/incident/views.py
@@ -143,7 +143,11 @@ def prepub_list(request):
         raise Http404
 
     lower_bound = datetime.date.today() - prepub_settings.get_timespan()
-    bar_chart_lower_bound = datetime.date.today() - datetime.timedelta(days=29)
+    # Bar chart is always bucketed by day; capped at 30d to avoid bars colliding.
+    bar_chart_days = min((datetime.date.today() - lower_bound).days, 30)
+    bar_chart_lower_bound = datetime.date.today() - datetime.timedelta(
+        days=bar_chart_days - 1
+    )
 
     confirmed_by_date = Counter(
         IncidentPage.objects.live()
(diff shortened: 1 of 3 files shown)


incident/views.py
```diff
@@ -143,7 +143,11 @@ def prepub_list(request):
         raise Http404
 
     lower_bound = datetime.date.today() - prepub_settings.get_timespan()
-    bar_chart_lower_bound = datetime.date.today() - datetime.timedelta(days=29)
+    # Bar chart is always bucketed by day; capped at 30d to avoid bars colliding.
+    bar_chart_days = min((datetime.date.today() - lower_bound).days, 30)
```
SaptakS, line 147: @ryanwhitney  Isn't `datetime.date.today() - lower_bound` basically equal to `prepub_settings.get_timespan()` ? Maybe we can just use that? or is there a particular reason for using this logic?

SaptakS: Actually I got my answer. My mathematical logic doesn't work with months relativedelta, so I think this code makes sense.

SaptakS (approved):
Looks good to me. Tested it locally, and seems to be working.

incident/views.py
ryanwhitney: Yeah, would have required another conversion to get day count out of the relative months

PR #2635, opened 2026-08-10
Always default homepage charts to smallest available timespan

ryanwhitney:
One small tweak on top of #2206

For homepage charts, the prior PR had the default timespan cascade like so: 7d (if enabled + data) -> 4w (if data) -> 6mo

This updates it to also include 12w, so it always chooses the smallest available bucket. (7d -> 4w -> 12w -> 6mo)

This likely won't be needed for prod, but staging has no data within 4w and this is preferable and more expected behavior.


SaptakS (approved):
Makes sense to me.

PR #2647, opened 2026-08-24
Fixes duplicate IDs in HTML tags

### LINKED ISSUE #1327: Elements have non-unique ids in database incident cards
image

The above element with the id is present for all incident cards. All ids must be unique in HTML. And since this id is also referenced by an `aria-labelledby`, it breaks screen reader functionality as well.

SaptakS:
## Description

Fixes #1327.

Adds incident.pk to update the id attributes to not be duplicate.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

Check in the all-incidents page, in mobile mode, the ids for "Incident details" are not duplicate.

### General checks

### DIFF
--- a/incident/templates/incident/_category_table.html
+++ b/incident/templates/incident/_category_table.html
@@ -1,5 +1,5 @@
 {% load wagtailcore_tags %}
-<dl class="details-table__category-fields" aria-labelledby="incident-details-title">
+<dl class="details-table__category-fields" aria-labelledby="{{ category.slug }}-details-title">
 	<h3 class="heading-table category category-{{ category.page_symbol }}" id="{{ category.slug }}-details-title">
 		<a class="text-link text-link--card" href="{% pageurl category %}">
 			{{ category.title }}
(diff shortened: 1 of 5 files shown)


incident/templates/incident/_category_table.html
```diff
@@ -1,5 +1,5 @@
 {% load wagtailcore_tags %}
-<dl class="details-table__category-fields" aria-labelledby="incident-details-title">
+<dl class="details-table__category-fields" aria-labelledby="{{ category.slug }}-details-title">
```
paulschreiber, line 2: should this have a corresponding ID that gets updated?

SaptakS: The [ID was already updated](https://github.com/freedomofpress/pressfreedomtracker.us/pull/2647/changes/BASE..b0d69b484259ae5510faf4a1de8cc48c13ab558a#diff-8277395032515b281869aa6434a2a8ae46a8ab9c3c58c425f5a73d17bc9bb31bR3) before but seems like the aria-labelledby wasn't.

ryanwhitney (approved):

PR #2650, opened 2026-08-28
build: overhaul container build logic

conorsch:
## Description

This PR attempts to standardize the container build logic for this repo, matching the pattern already approved and merged in:

  * https://github.com/freedomofpress/freedom.press/pull/3173
  * https://github.com/freedomofpress/securedrop.org/pull/1566
  
Specifically, the changes are:

* use a single containerfile with a multi-stage build, copying assets between
* organizes deployment build tooling under top-level `ci/` directory
* rewrites the `Makefile` as a `justfile` :sparkles: 
* intentional support for SELinux volume mounts, for Linux users

The containerfile refactor leaves the pip-tools usage in-place. We'll circle back on the  Python package management story in the near future. 

## Related work

* ports https://github.com/freedomofpress/freedom.press/pull/3173
* ports  https://github.com/freedomofpress/securedrop.org/pull/1566
* towards https://github.com/freedomofpress/infrastructure/issues/6967

## Testing and review

Please make sure the local compose setup works for you. Make sure you can run the operations you normally do, and report anything that's broken or surprising.

As with the other PRs, the commits are discrete enough to drop any individual pieces if anything's objectionable.


ci/containers/Containerfile
```diff
@@ -0,0 +1,237 @@
...
+RUN npm install && npm run build && \
+    rm -rf node_modules .git
+
+# ---- python base (shared by dev + prod) ----
+FROM ${PYTHON_IMAGE} AS python-base
+
+# The previous Dev and Prod images installed the same set; Dev merely split it
+# across two stages. `gcc` is omitted as redundant -- build-essential depends on
+# it. The -dev headers are retained for any dependency that lacks a cp314 wheel
+# and falls back to an sdist.
```
willbarton: Minor thing, but we don't really need to reference what the previous images did, do we? I'd keep the gcc (though one could probably make the same argument that's in reference to the previous Dockerfiles) and -dev headers comment though.

```suggestion
# `gcc` is omitted as redundant -- build-essential depends on it. The -dev
# headers are retained for any dependency that lacks a cp314 wheel and falls
# back to an sdist.
```

willbarton (approved):
I think this looks good. I've tested both the dev and prod compose setups, the `just` recipes, and checked the chart generation service.

I have one super-duper-minor non-blocking comment about a comment.

PR #2664, opened 2026-09-07
Adds cloudflare cache purging on prepub sync toggle

### LINKED ISSUE #2553: [Prepub sync] Purge cloudflare cache when prepublication sync setting is toggled
### Description

We need to trigger a cloudflare cache purge of the homepage and the unpublished incidents page when the setting is toggled. We should add signals to perform this purge.

***

Original description

@KAMcCudden  reported being able to visit /unconfimed-incidents on mobile after disabling the feature. May need some additional cache purging. (esp if that's true in the other direction.)

Image

Image

### Expected Behavior

_No response_

### Browser and OS

_No response_

### Urgency

### Notifications

_No response_

SaptakS:
## Description

Fixes #2553.

On toggling is_enabled in settings for Prepub sync, a cache purge is trigger for the URL of prepub list page, and the HomePage.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [ ] Config changes

## Testing

### General checks

### DIFF
--- a/incident/signals.py
+++ b/incident/signals.py
@@ -1,13 +1,23 @@
-from django.db.models.signals import post_delete
+from urllib.parse import urljoin
+
+from django.db.models.signals import post_delete, post_save, pre_save
+from django.dispatch import receiver
+from django.urls import reverse
 
 from wagtail.contrib.frontend_cache.utils import purge_page_from_cache
+from wagtail.models import Site
 from wagtail.signals import page_published
 
 import structlog
 
-from cloudflare.utils import purge_tags_from_cache
+from cloudflare.utils import purge_tags_from_cache, purge_urls_from_cache
 from common.models import CategoryPage
-from incident.models import IncidentIndexPage, IncidentPage
+from home.models import HomePage
+from incident.models import (
+    IncidentIndexPage,
+    IncidentPage,
+    PrepublicationSettings,
+)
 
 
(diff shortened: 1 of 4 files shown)


SaptakS:
This should be ready for review. I think the main testing might still need to be done in staging to check if cloudflare cache is actually getting cleared.

incident/signals.py
```diff
@@ -39,6 +49,36 @@ def purge_incident_index_from_frontend_cache(**kwargs):
...
+def stash_prepub_enabled_state(instance=None, **kwargs):
+    if instance.pk is None:
+        previous = PrepublicationSettings._meta.get_field("is_enabled").get_default()
+    else:
+        previous = (
+            PrepublicationSettings.objects.filter(pk=instance.pk)
+            .values_list("is_enabled", flat=True)
+            .first()
+        )
+    instance._previously_enabled = previous
```
chigby, line 62: This is clever! It does kind of feel like a deficiency in Django and/or the signals system that there's not an easier way to do this, but that's the way it is.

chigby (approved):

incident/signals.py
SaptakS: Yeah. I implemented first on just saving the settings, but since the requirement was on toggle, hence needed to use a hack like this.

PR #2673, opened 2026-09-12
fix: update eslint and clean up JavaScript files

### LINKED ISSUE #2188: Update to eslint 10
## Description
Update eslint from 8.35.0 to 10.x.

paulschreiber:
## Description
- Update eslint to v9
- Use new flat config
- use new config plugins
- correct errors

Fixes https://github.com/freedomofpress/pressfreedomtracker.us/issues/2188

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Vulnerabilities update
- [X] Config changes


ryanwhitney:
I think we may want to start divesting from `propTypes` and disabling the rule from `eslint-plugin-react` rather than adding more. It's been [fully removed](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops) in react 19 after being deprecated for a long time and will no longer validate anything at runtime. 

I believe the current recommendation if we do want some enforcement around types (with js), is to write JSDoc comments and add a typescript dev dependency purely for `checkJs`.

build/static/js/picturefill.3.0.2.min.js
ryanwhitney: vendored, should not be linted

package.json
ryanwhitney: this is not actually linting most js/jsx files because of the glob thing.

client/charts/js/components/CategoryFilter.jsx
```diff
@@ -187,32 +198,45 @@ export default function CategoryFilter({
...
 									y={height - margins.bottom / 2}
-									transform={'translate(0, -7)'}
+									transform={"translate(0, -7)"}
 									opacity={d.count === 0 ? 0.3 : 1}
 								>
 									<div
-										className={classNames('category', `category-${d.symbol}`)}
+										className={classNames("category", `category-${d.symbol}`)}
+										role="button"
+										tabIndex={0}
```
ryanwhitney: are the `tabIndex` and the `onKeyDown` additions intended to be here?  

(it looks like we're adding tabIndex to items with no accessible name)

package.json
```diff
@@ -18,20 +18,24 @@
   ],
   "devDependencies": {
```
ryanwhitney: same q as SDO: should prettier be here?

client/charts/js/components/FilterSet.jsx
```diff
@@ -220,42 +177,48 @@ function RadioFilter({
...
 }
 
-export default function FilterSet({ filters, handleFilterChange, filterParameters, width, dataset, filterWithout}) {
+export default function FilterSet({
+	filters,
+	handleFilterChange,
+	filterParameters,
+	width,
+	dataset,
+	filterWithout: _filterWithout,
```
ryanwhitney: If `filterWithout` is unused, can we instead not include it when destructuring

ryanwhitney (changes requested):
I'm still getting errors here when I run eslint on all files. mostly `prop-types` issues. Can you fix the file glob in the package file and re-do this?

